PR #25, opened 2022-10-01
Fix #17 #20 #23 #24

### LINKED ISSUE #17: MFC website does not link to MFC code!
Title says it all. mflowcode.github.io does not seem to link back to the MFC org. page or MFC repo. I think that should be front and center.

### LINKED ISSUE #20: `cmake` build from binary issues
Two things

1. If you attempt to build MFC on a compute node, not every compute nodes seems to be able to fetch the cmake binary. Expanse seems to be an example of this. Maybe a `ping` or something could check if a connection is actually working.
2. Even if you do get `cmake` from binary, MFC still issues a warning 
```
exp-1-57: project-bryngelsonPI/MFC $ ./mfc.sh
./mfc.sh: line 82: cmake: command not found
```
even though it actually works in the end.

### LINKED ISSUE #23: `./mfc.sh` builds python packages for `venv` on first launch without letting the user know
`./mfc.sh` builds python packages for `venv` on the first launch without letting the user know what it's doing. I think this is a bit confusing, and a message before and after would be useful!

### LINKED ISSUE #24: Rename `pre_process_code`, ...?
Thoughts on removing `_code` from the `src/` subdirectories? e.g. `pre_process_code` -> `pre_process`. @henryleberre @anandrdbz

henryleberre:
This Pull Request
- Closes #17 
- Closes #20 
- Closes #23 
- Closes #24 

I also took this opportunity to make `mfc.sh` print what it does and detect when the CMake binary it sometimes has to download becomes incompatible (and downloads CMake again). It also extracts the required CMake version from `CMakeLists.txt` at runtime.


henryleberre:
@sbryngelson There currently is no support for running CI pipelines on GitLab for Pull Requests originating from forks on GitHub ([see discussion here](https://gitlab.com/gitlab-org/gitlab/-/issues/5667)). This is why the ORNL pipeline is not running.

The issue I fixed in #18 was actually related to the fact that I was using the commit ref's slug (instead of its name) when running `git clone -b <ref> <URL>` in the pipeline. I had first thought that the Pull Request was made from the (external) MFC-develop repository but I was wrong. It was a subtle issue because it only fails when the ref (commit name, tag, etc..) is contains certain characters (like capital letters and spaces). Here are the details:
> `CI_COMMIT_REF_SLUG`: `CI_COMMIT_REF_NAME` in lowercase, shortened to 63 bytes, and with everything except `0-9` and `a-z` replaced with `-`. No leading / trailing `-`. Use in URLs, host names and domain names.

([GitLab: Predefined variables reference](https://docs.gitlab.com/ee/ci/variables/predefined_variables.html))

PR #39, opened 2022-11-12
Close #13

### LINKED ISSUE #13: Equation of state is not DRY
The equation of state is used to compute the pressure from the variables. This EOS changes slightly for different models. The EOS implementation is repeated several places in the code, including in `m_variables_converison`, `m_data_output` (including serial data output and probe outputs), and likely more. Following the [DRY principle](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) (don't repeat yourself), we should fix this. It has also caused multiple false bugs in the past.

anshgupta1234:
In `m_variables_conversion` in pre_process and post_process, added a pointer `s_compute_pressure` that points to 3 new subroutines:
- `s_compute_pressure_from_energy` (when model_eqns != 4 and no bubbles)
- `s_compute_pressure_from_bubbles` (when model_eqns != 4 and bubbles)
- `s_compute_pressure_4eqns` (when model_eqns == 4)

In `m_data_output` in pre_process and `m_variables_conversion` in pre and post-process, the EOS has been replaced by a call to `s_compute_pressure`


sbryngelson:
Does this work on GPUs? Can you verify that the post-processing `.silo` files are the same before and after this change (they aren't part of the CI).

sbryngelson:
You also did not modify the `simulation` code, which is part of this issue

henryleberre:
@anshgupta1234 You might also want to consider placing these subroutines in a file in src/common

anshgupta1234:
@henryleberre Actually, the subroutines are slightly different in `pre_process` and `post_process` so I probably can't do that 😅

sbryngelson:
> @henryleberre Actually, the subroutines are slightly different in `pre_process` and `post_process` so I probably can't do that 😅

Consider changing and refactoring them so that they can all be the same.

anshgupta1234:
Sure! I think I can refactor them and try to bring them into src/common. Is there any specific file I should bring it into or create a new module?

I'll also work on `simulation`, but that might take a while since it has to be compatible with OpenACC.

Apart from that, I actually don't have any GPUs on my laptop so I'm unsure if I can do GPU testing for this code.

anshgupta1234:
After looking through the code in `simulation`, I saw that other procedure pointers (i.e `s_convert_abstract_to_mixture_variables`) had acc declarations in its initialization and subroutines. Therefore, I'm unsure if I can move `s_compute_pressure` code to `src/common` -- wouldn't I have to have 2 versions, one without acc and another one with in order for the code to work in simulation as well?

sbryngelson:
It should probably be a new module. Probably `m_variables_convert.f90`. I added you to a GPU cluster.

anshgupta1234:
Great! I'll test it out on a GPU whenever I have the chance. Till then, I did make sure to run `examples/2D_shockbubble` and `examples/2D_axisym_shockbubble` on this branch and master, and they both looked identical in ParaView. 

In this latest commit, I did name the module `m_compute_pressure` because naming it `m_variables_convert` might be more confusing when importing in other files, and could look like:

```
use m_variables_conversion    ! From local dir
use m_variables_convert       ! From src/common
```

If this is not an issue, I can definitely change the file name :)

sbryngelson:
I think this PR should also try to move as much of `m_variables_conversion.f90` to a `common` location as possible (perhaps all of it!). Great job so far.

src/pre_process/m_variables_conversion.f90
```diff
@@ -245,6 +247,15 @@ subroutine s_initialize_variables_conversion_module() ! -------------------
 
     end subroutine s_initialize_variables_conversion_module ! -----------------
 
+    !>  This procedure calculates the pressure based on energy when
```
sbryngelson: It looks like some doxygen comments from an old routine were left here? @anshgupta1234

sbryngelson:
@anshgupta1234 what is the status of this? I cannot tell.

sbryngelson:
It's fine how it is I was just curious.

src/common/m_variables_conversion.fpp
```diff
@@ -1,16 +1,17 @@
 !>
 !! @file m_variables_conversion.f90
 !! @brief Contains module m_variables_conversion
+!! @author A. Gupta
```
sbryngelson: remove all @author statements

src/common/m_variables_conversion.fpp
```diff
@@ -1,16 +1,17 @@
...
+!! @version 1.0
```
sbryngelson: remove @version and @date from all of your files please @anshgupta1234

sbryngelson (approved):

src/pre_process/m_variables_conversion.f90
anshgupta1234: Ah, but I believe it's alright since that file is already deleted in this version? That comment seems to be in the pre_process version of m_var_conversion which is gone now.

src/common/m_variables_conversion.fpp
```diff
@@ -128,39 +129,156 @@ end subroutine s_compute_abstract_average_state
...
+        rho_K => rho_sf(i, j, k)
+        gamma_K =>  gamma_sf(i, j, k)
+        pi_inf_K => pi_inf_sf(i, j, k)
+#endif
+
+#ifndef MFC_POST_PROCESS
+        rho_K  => rho
+        gamma_K => gamma
+        pi_inf_K => pi_inf
+#endif
```
henryleberre: I think having a single `#else` directive instead of both `#endif` and `#ifndef` would make for more readable code. The `n` is quite hard to spot! There are a few places in this file where this is the case.

henryleberre (approved):

PR #43, opened 2022-11-30
Refactoring  + Additional Optimizations

anandrdbz:
Code is fully refactored (RHS only 2800 lines) and passes all test cases on GPUs. Dead code removed from m_cbc, m_weno, m_monopole and m_bubbles as well (FYPP, removing _flat, _alt etc)

Run times are identical (slightly quicker in fact) for the fully resolved cases. 

Sub-grid bubbles, which was previously quite slow on GPUs, now runs 20X faster (had to modify the bubble source kernel).  Now, they exhibit identical speedups to the fully resolved case. 

QBMM cases also show slight (5%) improvement from previous state. However GPU speedup is limited (3-4X slower than fully resolved / bubbles)  due to the vast number of power operations (This is quite slow on OpenACC + F90). 

In summary, runtimes for all cases are either identical to previous state or significantly better, with a vast reduction in number of lines


sbryngelson:
@anandrdbz @wilfonba CI failed.

runTimeTests.sh
```diff
@@ -0,0 +1,4 @@
+#!bash
+
```
sbryngelson, line 2: remove this file

src/simulation/m_rhs.f90
```diff
@@ -920,58 +856,58 @@ subroutine s_compute_rhs(q_cons_vf, q_prim_vf, rhs_vf, t_step) ! -------
...
+                call s_reconstruct_cell_boundary_values( &
                     q_prim_qp%vf(iv%beg:iv%end), &
-                    qL_rsx_vf_flat, qL_rsy_vf_flat, qL_rsz_vf_flat, &
-                    qR_rsx_vf_flat, qR_rsy_vf_flat, qR_rsz_vf_flat, &
+                    qL_rsx_vf, qL_rsy_vf, qL_rsz_vf, &
+                    qR_rsx_vf, qR_rsy_vf, qR_rsz_vf, &
                     id)
 
                 iv%beg = mom_idx%beg; iv%end = mom_idx%end
                 if (weno_Re_flux) then
```
sbryngelson, line 894: it seems like this subroutine `s_reconstruct_cell_boundary_values_visc_deriv` should be moved to the viscous module?

sbryngelson (approved):

src/simulation/m_rhs.f90
wilfonba: I must have missed that one. I'll move it.

sbryngelson: thanks @wilfonba -- i already pulled your request so please create a new one for it.

wilfonba: Will do!

PR #46, opened 2022-12-16
Add support for cylindrical to monopoles

haochey:
Add the 3-dimensional spatial support for cylindrical coordinate system to monopoles sources.


sbryngelson:
Please update the documentation for how to use the different monopole "supports". For example, `support(nm) == 6` is now for the cylindrical coordinates, and the other numbers correspond to other things. I think this should go near the same tables we use for describing what bound condition number is which.

haochey:
> Please update the documentation for how to use the different monopole "supports". For example, `support(nm) == 6` is now for the cylindrical coordinates, and the other numbers correspond to other things. I think this should go near the same tables we use for describing what bound condition number is which.

Could you please clarify how to edit the documentation for everyone?

sbryngelson:
https://github.com/MFlowCode/MFlowCode.github.io/blob/master/documentation/md_running.html

https://github.com/MFlowCode/MFlowCode.github.io/blob/4b134e8eb3d0bbfe3baef60e48eca20b4c15557f/documentation/md_running.html#L427-L459

sbryngelson:
Sorry - please change it here https://github.com/MFlowCode/MFC/tree/master/doc/documentation

haochey:
> Sorry - please change it here https://github.com/MFlowCode/MFC/tree/master/doc/documentation

Thanks. The documentation is updated.

sbryngelson (approved):

Issue #65, opened 2023-01-15
postprocessed silo files don't have correct primitive vars


sbryngelson:
The post-processed silo files don't have correct primitive variables. Likely an issue with `common/m_variables_conversion.fpp` subroutine `s_convert_conservative_to_primitive`.

sbryngelson:
fixed with #66

PR #68, opened 2023-01-20
Fixed #48

### LINKED ISSUE #48: Refactor for new variable conversion and modularize sound speed
From here: 
https://github.com/MFlowCode/MFC/issues/13#issuecomment-1340966703

We see that we compute the primitive variables "by hand" several times throughout the code.
This happens in `m_data_output` and some other places as well.
Now that @anshgupta1234 has made the routines nice and simple for this, we should call those instead. 

Also, the speed of sound `c` is computed by hand in several places, though this could be moved out to `common/`. `c` is also sometimes called things like `c_avg` or `c_L` or `c_R`.

anshgupta1234:
**Changes:**
- `prim_vars_wrt` now works for 2D and 3D cases and prints out the primitive variables to D/
- Added `common/inline_conversion` with the macro `compute_speed_of_sound` that has been implemented in `simulation`. 
- Made the fix for #65 less ugly


sbryngelson:
@anshgupta1234 you now have a merge conflict

PR #73, opened 2023-01-27
Fixed error with calling conservative to primitive vars in m_data_output

anshgupta1234:
Co-authored-by: Henry Le Berre <hberre3@gatech.edu>

This PR passes in `q_prim_vf` into `s_write_data_files` in order to use the array for converting to conservative and primitive variables. It also declares alpha_L and alpha_R as private variables in `m_reimann_solvers`


henryleberre (approved):
lgtm

PR #74, opened 2023-01-28
Fix post-process bug & Small fixes

henryleberre:
- Fix bugs in the common variable conversion routines for post-process where the wrong indices were being used:
```f90
#ifdef MFC_POST_PROCESS
        rho_sf(i, j, k) = rho
        gamma_sf(i, j, k) = gamma
        pi_inf_sf(i, j, k) = pi_inf
#endif
```
instead of
```f90
#ifdef MFC_POST_PROCESS
        rho_sf   (j, k, l) = rho
        gamma_sf (j, k, l) = gamma
        pi_inf_sf(j, k, l) = pi_inf
#endif
```

Here `i` was just a generic loop iterator.

- Remove the requirement to specify `case_dir`. All cases set it to `.` because `mfc.sh` just runs mfc from within the case directory. This user input was probably useful when mfc was still using the `master_scripts` setup for running jobs. It may confuse users more than it benefits them to have to included. The default is now `.`.

- Removed unnecessary whitespace and `\` (at the end of lines) in example cases.

- Fixed a few other small issues like
```f90
        real(kind(0d0)), parameter :: alpha_mp = 2d0
        real(kind(0d0)), parameter :: beta_mp  = 4d0/3d0
```
never being declared.


sbryngelson:
Can we create tests to ensure bugs like this don't go unnoticed in the future?

henryleberre:
We could perhaps just post-process all test cases (without checking results). This bug resulted in the process aborting so it would have been caught in this way. What do you think? If you agree, should this be the default behavior of `./mfc.sh test` ?

sbryngelson:
> We could perhaps just post-process all test cases (without checking results). This bug resulted in the process aborting so it would have been caught in this way. What do you think? If you agree, should this be the default behavior of `./mfc.sh test` ?

Yes I think that would be good (to try out at least). An additional test is an `h5dump` on the HDF files and doing a `grep` for `NaN`s (and ensuring that the files even exist). Perhaps you can find more clever ideas as well.

sbryngelson (approved):

PR #75, opened 2023-01-29
Move easily portable and repeated MPI calls to `common/` 

sbryngelson:
This PR moves portable MPI calls to the common directory. All of these calls were shared across all 3 sub-codes and only a couple `ifdef`s were needed. Clean change that reduces SLOC. I think the other MPI calls can also be moved, but will require more care.

### DIFF
--- a/src/common/m_mpi_common.fpp
+++ b/src/common/m_mpi_common.fpp
@@ -1,0 +1,354 @@
+
+#:include 'macros.fpp'
+
+!> @brief The module serves as a proxy to the parameters and subroutines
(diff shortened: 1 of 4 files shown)


sbryngelson:
@henryleberre @anshgupta1234 someone please run/test this on GPUs somewhere before merging.

anshgupta1234 (approved):
Tested successfully on Bridges2 w/ GPU.
 
image

henryleberre (approved):
I confirm that multi-GPU runs complete on Summit (e.g. `ppn=4` on `3D_shockdroplet`) as well as `ppn=2` test cases.

PR #85, opened 2023-02-02
Remove -f from rm in s_delete_directory

### LINKED ISSUE #50: Unnecessary flags in directory deletion
**Current Behavior**
Removing a directory on *nix systems through a call to the `s_delete_directory` subroutine invokes `rm -rf` to remove a directory along with any files contained within. Calling the wrong path with elevated permissions can be dangerous. Only directories created by the program should be deleted, which should not necessitate the use of the force flag with correct filesystem permissions. If there are incorrect filesystem permissions, it is not the responsibility of the subroutine to rectify them.

**Proposed Change**
Change the system call to `rm -r`

belericant:
Resolves #50

### DIFF
--- a/src/common/m_compile_specific.f90
+++ b/src/common/m_compile_specific.f90
@@ -43,7 +43,7 @@ contains
 #ifdef _WIN32
         call system('rmdir "'//dir_name//'" /s /q')
 #else
-        call system('rm -rf "'//dir_name//'"')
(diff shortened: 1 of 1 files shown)


belericant:
@henryleberre I tested and this change doesn't break functionality on Bridges2. I don't have access to other systems to test on.

sbryngelson (approved):

PR #86, opened 2023-02-02
Conversions Update

anshgupta1234:
This PR accomplishes two separate things:

- It consolidates the hypoelastic pressure calculation within the `s_compute_pressure` subroutine.
- It does all speed of sound calculations with the `compute_speed_of_sound` and eliminates a few parameters for the inline subroutine.

Tested on Bridges2 w/ GPU and visualized examples/2D_shockbubble correctly.


sbryngelson:
Can you do benchmark tests to see what degree this slowed things down?

sbryngelson:
Thanks! I found a few things, for example:

https://github.com/anshgupta1234/MFC/blob/ccf1f443679249277e013f5027bc3db1ca8dfe2b/src/post_process/m_derived_variables.f90#L250-L304

and 

https://github.com/anshgupta1234/MFC/blob/ccf1f443679249277e013f5027bc3db1ca8dfe2b/src/simulation/m_data_output.fpp#L1038

https://github.com/anshgupta1234/MFC/blob/ccf1f443679249277e013f5027bc3db1ca8dfe2b/src/simulation/m_start_up.fpp#L1127

sbryngelson:
I'm a little concerned about the readability of these inline functions. The main problem is that you do not get to define a scope. So for example, in your case one needs `blkmod1`, `blkmod2`, etc. etc. to all be defined whenever one calls your inlined routine, which lives in a different file. This won't be obvious to the developer. This gets more complicated if one isn't sure what type a variable should be (in which case I suppose they would hunt around for another invocation of this routine and copy/paste that). 

I'm open to ideas that make these "inlined" fypp macro functions more readable without sacrificing speed. One option is to make this a private subroutine that gets stamped at the bottom of all relevant f90 files on compilation. With this, you get to define the scope yourself, but you don't sacrifice speed because it will be in the appropriate module. There are a few problems with even this, though, like having to include the fypp file at the top of all the relevant modules (maybe this isn't such a high price?). 

I think there is probably a better way than this suggestion though. @henryleberre ? @anshgupta1234 ? (any ideas..)

sbryngelson:
@anshgupta1234 can you recheck the runtimes on GPU? sorry

sbryngelson:
this does seem like quite the clean fix to the problem, though.

anshgupta1234:
@sbryngelson I would certainly test it if Bridges2 would just give me a GPU!

sbryngelson:
You can also try Phoenix/Wingtip3/etc.

anshgupta1234:
Master: 
- Riemann solvers: 6e-4
- Total time: 20.6s

Current:
- Riemann solvers: 7e-4
- Total time: 28s

@sbryngelson

sbryngelson:
This is quite the change (a 50% increase in runtime!). I'm not sure we should proceed this way (or why it is so slow).

sbryngelson:
you should be able to see if the call was indeed inlined using `Minfo=accel`.  I'm guessing it wasn't.

anshgupta1234:
@sbryngelson It looks like it's being inlined...
image

sbryngelson:
It does appear that way. And can you confirm that there is still this massive slowdown in m_riemann_solvers? if so, can you confirm it is associated with this subroutine by using nvtx statements (which will tell you how much time is spent in certain code regions when you perform a profile with Nsight Systems).

anshgupta1234:
This is very odd...I just ran `master` and `conversions_update` on Phoenix SLURM, and the times were nearly identical! (28s vs 29s) I'll run it again tomorrow to make sure it's not a big slowdown.

sbryngelson:
Please also check on an older commit as well.

anshgupta1234:
@sbryngelson Tested on a [week-old commit](https://github.com/MFlowCode/MFC/commit/90f0e406d2fc32c402268d619f4b653299f32347). Still 28s.

sbryngelson:
Great. Can you do an nsight systems profile to check the proportions of which kernel takes however much time? Reimann solvers and WENO should be similar. This is useful information regardless, since I'm not sure the last time we did such a profile. In the future it will be in the CI workflow.

anshgupta1234:
As expected, this branch and master are basically identical in benchmarking. 

**Benchmark**

Test 1:
- My Branch: 2.93e-2 s/timestep
- Master: 3.04-2 s/timestep

Test 2:
- My Branch: 1.57e-2 s/timestep
- Master: 1.58e-2 s/timestep

These are 3D cases run on 2 V100s on Phoenix. I think this PR is good to go, I'll just compress it.

sbryngelson:
@henryleberre can you please have a quick review of this

sbryngelson (approved):

PR #90, opened 2023-02-09
Add NoSlipWall as "-14"

haochey:
Add the NoSlip Wall boundary condition to the options as "-14" which works with "WENO order = 1"


sbryngelson:
Cannot be merged until validated and working for all WENO orders.

PR #91, opened 2023-02-09
Helper Functions

wilfonba:
This pull request completes all of the refactoring suggested in issue 86. Run time for three 3D representative problems (viscous, bubbles, etc) is the same as before making any changes on GPUs.

I'm expecting CI to fail the first time, but posting in hopes of someone being able to provide some insight as to why (@sbryngelson @anandrdbz @henryleberre). Currently, the test suite passes on my M2 Mac with GNU 12.2 on CPUs, Henry's Linux machine (not sure what compiler), and my Linux desktop with NVHPC 23.1 and an RTX 2060. It fails CI with GNU 11.3 though.


wilfonba:
Small update. It is now failing consistently on different devices
.

sbryngelson:
Enable GitHub actions and run CI on your fork at various different places in your long-ish commit history. Based on the above, it's unclear what commit actually 'broke' the CI (which would be useful info.)

sbryngelson:
Also, I see you have some new fypp inline subroutines. Can you elaborate on what new such routines you have and why you put them there? 

If there is only one such subroutine and it only needs to go one place (e.g., `comp_n_from_cons`), and it is short, I have no problem with just putting it in the code somewhere. I think having it in one place as an obscure/inappropriate part of the normal fortran code is better than hashing it away in an fypp routine (which many developers will not be used to).

wilfonba:
Ah, I forgot to move that subroutine back to m_helpers. It turns out that with the right command line options FYPP inlining isn't necessary for any of the routines in m_helper.

henryleberre:
@wilfonba You pushed your local version of `case.fpp` in simulation which is causing the loading of the namelist to fail because some variables are parameters (i.e. constants). You could maybe update the `.gitignore`, replacing the old `src/*/case.fpp` with `src/*/include/case.fpp`.

henryleberre:
Your `src/*/autogen/` folders seem to have been pushed somehow..

sbryngelson:
@wilfonba get rid of the `autogen/` files please!

sbryngelson:
also, we _really_ need to do a proper nsight sys profile of both this PR and @anshgupta1234. I will merge both once I see that the runtimes (basically) match before and after the PR and that the relative compute time in each nvtx range is similar. 

I think we need not wait on @belericant 's new feature (auto-benchmarking) to do this. Instead, you can just perform this manually before and after the PR on an 8M pts/GPU shock-droplet case and post the results here. As I said, once @belericant is done, this will be checked automatically.

sbryngelson:
@wilfonba can you remove your autogen files please and then confirm if this is mergeable

wilfonba:
@sbryngelson Done and is mergeable as far as I can tell

sbryngelson:
I see several files have duplicated the line `#:include 'macros.fpp'` like this:

```
#:include 'macros.fpp'

#:include 'macros.fpp'
```

For example, `m_viscous.fpp` (and several others)

sbryngelson:
The calls to subroutine `s_comp_n_from_prim` seem to have vanished from the code, where did they go?

sbryngelson:
It seems like _small_ functions or subroutines that are only called from one place should be put back into their "homes." 

Also, I see no calls to some of these other routines as well. like `s_quad`. I suspect these were also "hard-coded".

src/simulation/m_viscous.fpp
```diff
@@ -1,6 +1,7 @@
 !>
 !! @file m_viscous.f90
 !! @brief Contains module m_viscous
+#:include 'macros.fpp'
```
sbryngelson, line 4: this is included twice?

src/simulation/m_weno.fpp
```diff
@@ -1,6 +1,7 @@
 !>
 !! @file m_weno.f90
 !! @brief Contains module m_weno
+#:include 'macros.fpp'
```
sbryngelson, line 4: included twice?

src/common/m_helper.f90
```diff
@@ -0,0 +1,191 @@
...
+                fd_coeff_s(-1, i) = -8d0*fd_coeff_s(-2, i)
+                fd_coeff_s(0, i) = 0d0
+                fd_coeff_s(1, i) = -fd_coeff_s(-1, i)
+                fd_coeff_s(2, i) = -fd_coeff_s(-2, i)
+            end do
+
+        end if
+
+    end subroutine s_compute_finite_difference_coefficients ! --------------
+#else
```
sbryngelson: merge these two subroutines by placing an ifdef on the loop bounds, since that seems to be the only difference? or am I missing something?

sbryngelson (changes requested):

src/common/m_helper.f90
wilfonba: Fixed with an optional argument instead

wilfonba: Figured that was slightly cleaner

sbryngelson:
@wilfonba failed test

wilfonba:
@sbryngelson It failed at the first hypoelasticity post-process test

wilfonba:
Not sure how those passed on Ansh's pull request. I haven't changed anything with regards to hypoelasticity

anshgupta1234:
@wilfonba Hmm I'm seeing that the first failed test was on 3D -> bc=-1? The error was "Fortran runtime error: Attempting to allocate already allocated variable 
'fd_coeff_s'"

wilfonba:
@anshgupta1234 oops. I didn't read that closely. check slack for a different issue

sbryngelson:
it passed the tests before your last change...

wilfonba:
@sbryngelson Yeah. I didn't read the output very carefully. I know what the issue is now, but I assumed it had to do with the issue I'm having on summit

wilfonba:
Execution times for three test cases (3d, 27 mil cells, 6 GPUs) are as follows:

1. Inviscid sub-grid bubbles
- MFC: 1.632e-2 sec/timestep
- This: 1.637e-2 sec/timestep

2. Inviscid shock droplet
- MFC: 1.365e-2 sec/timestep
- This: 1.362e-2 sec/timestep

3. Inviscid monopole
- MFC: 1.275e-2 sec/timestep
- This: 1.275e-2 sec/timestep

I'm not sure why it's still showing 1 change requested. I fixed all the three you requested and marked them all as resolved. @sbryngelson

sbryngelson (approved):

PR #96, opened 2023-02-14
mfc.sh load updated for Phoenix SLURM

anshgupta1234:


henryleberre:
@anshgupta1234 is there a reason for loading both Intel and GNU compilers? Either one should do the trick!

anshgupta1234:
You're right :) I removed intel, since `mvapich2` has a `gcc` dependency.

henryleberre (approved):
@anshgupta1234 lgtm provided you successfully built MFC with it!

Issue #37, opened 2022-11-11
Unclear which variables are which


sbryngelson:
This has been a persisting issue for years. It is unclear what `q_prim_vf(i)` corresponds to what variable for what cases for each `i`. Likewise, what `q_cons_vf(i)` gets paired with it. Need a resolution for this and add it to the docs. Probably want a simple table with two columns that list the variables in the order they would appear, should they exist. For example, I believe sub-grid bubble variables come before hypoelastic ones, but I'm not 100% sure. What about the 6-equation model?

PR #100, opened 2023-02-17
Variables and Patch Documentation

### LINKED ISSUE #88: Document undocumented patches
Currently, patch types 14 to 19 are undocumented and they perhaps should be.

While on the topic, wouldn't it be clearer if each patch type had its own derived type within `patch_icpp(i)`? For example, an STL patch (#83) could be defined as:
```
patch_icpp(2)%geometry:      20,
patch_icpp(2)%stl%filepath:  'path/to/stl/file`,
patch_icpp(2)%stl%scale(2):  1.0,
patch_icpp(2)%stl%offset(3): 0.25,
```

This, in a sense, self-documents what properties each patch type can have. Every attribute that is common to all patches would be in the base `patch_icpp(i)%` scope.

wilfonba:
This fixes #88 #37


sbryngelson:
@wilfonba please link the issues this fixes

wilfonba:
Done @sbryngelson

sbryngelson (approved):

PR #107, opened 2023-03-01
Post Process Tests

anshgupta1234:
This PR adds a new feature to `./mfc.sh test`: `-a` or `--test-all`. This argument re-runs the tests, but this time with `parallel_io` on to make sure `post_process` isn't producing NaNs or Infinities. A few notable things:
- `-a` is turned off by default.
- `ci.yml` turns on `-a`.
- **Currently, this feature does not test the hypoelasticity cases because post process fails on them. Once that issue has been fixed, we can make sure post process for hypoelasticity is tested as well.**

Tested on CPUs and GPUs. Ensured that it caught NaNs and Infinities by forcing post process to produce them.


sbryngelson:
Please enable this feature for _all_ cases (even the failing ones), then fix the failing ones. Hypoelasticity post-processing used to work fine, so it should be clear what the code looks like.

anshgupta1234:
@sbryngelson found and fixed the hypoelasticity error. It was really minor.

sbryngelson (approved):

Issue #103, opened 2023-02-27
patch creation missing check


sbryngelson:
I can write an input file `case.py` with two patches, but `num_patches = 2`, and it will still run.

wilfonba:
I'm assuming you mean num_patches != 2?

sbryngelson:
closed with #110

PR #110, opened 2023-03-03
Check if num_patches mismatches the specified patches

anshgupta1234:
Fixes issue #103. The loop was just broken and it wasn't going through all of `num_patches_max`.

### DIFF
--- a/src/pre_process/m_check_patches.fpp
+++ b/src/pre_process/m_check_patches.fpp
@@ -32,7 +32,7 @@ contains
 
         integer :: i 
 
-        do i = 1, num_patches
+        do i = 1, num_patches_max
             if (i <= num_patches) then
                 ! call s_check_patch_geometry(i)
 


sbryngelson:
https://github.com/MFlowCode/MFC/blob/31420aaccaeecfefd01363be800a972e98fd2506/src/pre_process/m_start_up.fpp#L800-L858

I do not believe this fixes the problem (did you check?)

anshgupta1234:
Yep, I checked on a couple of different cases and this worked. Changing the loop to `num_patches_max` starts triggering the `else` statement on the `if (i <= num_patches)` block. The `else` statement effectively checks: has the patch been assigned to dflt_int? If not, then it's been specified, therefore the number of actual patches is higher than it should be and it throws an error.

Moreover, the current code does not make sense:

```
do i = 1, num_patches
    if (i <= num_patches)
       ...
```

It doesn't make sense for the if statement to be there unless the loop is through num_patches_max.

sbryngelson:
You are correct, and this check was broken a long time ago (~1 yr ago). Thanks.

Issue #42, opened 2022-11-27
Switching to modern and modular precision declaration


sbryngelson:
MFC is filled with lines like this:

https://github.com/MFlowCode/MFC/blob/a39962ca7378fab292380a0e8694cc953f38b7e9/src/simulation/m_hypoelastic.f90#L30-L32

where the precision is declared via `kind(0d0)`. 
We also have a ton of this 

https://github.com/MFlowCode/MFC/blob/a39962ca7378fab292380a0e8694cc953f38b7e9/src/simulation/m_riemann_solvers.fpp#L547-L556

where inline constants have precision declared in a "hard-coded" way. 

What is better is declaring a separate constant that we can change as needed, like this 
[example](https://github.com/crscillitoe/FintechTM/blob/3336a7170e5012f921e62e49bf4e0232ee47d349/FRONTEND/flibs-0.9/flibs/src/computing/select_precision.f90), though there are many others.

A fix for this issue would remove all cases of `0d0` and `kind(0d0)` and replace them with a constant declared in the `common/` directory. I think this is a suitable task for @anshgupta1234 .

I realize one can force precision via compiler variables, but I believe we should avoid this because there is an established language standard.

sbryngelson:
note this from the miniweather app:

https://github.com/mrnorman/miniWeather/blob/9869dcffb19c4b07bc88f2b8adf98ca9e95bb997/fortran/miniWeather_mpi_openacc.F90#L13-L20

an important thing to notice is that if one switches to single precision, you have to switch the MPI kind as well (notice all the MPI_DOUBLE_PRECISION business in `m_mpi_proxy.f90`).

sbryngelson:
Bump to top as the PR #119 was closed

PR #119, opened 2023-03-10
Add Modular fp

belericant:
Addresses issue #42


sbryngelson:
@anshgupta1234 would you mind giving this PR a review as well, just to check for typos that might be around? I looked, but another set of eyes would be useful. Thanks!

anshgupta1234 (approved):
Didn't find anything else. Looking good!

henryleberre:
@sbryngelson What do you think about moving the contents of `m_precision_select` into `m_constants` ? 

I wonder how NVFortran will handle precision selection because we specify `-Mr8`:
> -Mr8: (Fortran only) The compiler promotes REAL variables and constants to DOUBLE PRECISION variables and constants, respectively. DOUBLE PRECISION elements are 8 bytes in length.
>
> -Mnor8: (Fortran only) The compiler does not promote REAL variables and constants to DOUBLE PRECISION. REAL variables will be single precision (4 bytes in length).

From https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-ref-guide/#cmdln-m-opts-optimize. It's unclear to me when and how this "promotion" happens.

In any case, should this feature be made available to the build system? I feel like another way we could have implemented this would have been to use `real` everywhere and let CMake set the appropriate compiler flags. We already use the flags that turn `real` into `double` everywhere. Setting single precision could (possibly) be as simple as not specifying these flags.

The benefit of the approach @belericant took is that you could elect not to use working precision for kernels that need double precision. It's modularity at the cost of verbosity. 

One last remark, we may lose precision in a few places by being aggressive in the way we enforce working precision. Consider this example
```fortran
x_cc(i) = x_domain%beg + (5._wp * (10._wp ** -(1)))*dx*real(2*i + 1, wp)
```

`(5._wp * (10._wp ** -(1)))` is a constant evaluated at compile-time. All operations that lead to its evaluation are done in working precision. It might be preferable to compute this in double precision and then cast it to single working precision for better fidelity.

sbryngelson:
```
@sbryngelson What do you think about moving the contents of m_precision_select into m_constants ?
```
@henryleberre I don't have particular preference about this, seems fine.

```

```
I wonder how NVFortran will handle precision selection because we specify -Mr8:

-Mr8: (Fortran only) The compiler promotes REAL variables and constants to DOUBLE PRECISION variables and constants, respectively. DOUBLE PRECISION elements are 8 bytes in length.

-Mnor8: (Fortran only) The compiler does not promote REAL variables and constants to DOUBLE PRECISION. REAL variables will be single precision (4 bytes in length).

From [docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-ref-guide/#cmdln-m-opts-optimize](https://docs.nvidia.com/hpc-sdk/compilers/hpc-compilers-ref-guide/#cmdln-m-opts-optimize). It's unclear to me when and how this "promotion" happens.

In any case, should this feature be made available to the build system? I feel like another way we could have implemented this would have been to use real everywhere and let CMake set the appropriate compiler flags. We already use the flags that turn real into double everywhere. Setting single precision could (possibly) be as simple as not specifying these flags.

The benefit of the approach @belericant took is that you could elect not to use working precision for kernels that need double precision. It's modularity at the cost of verbosity. Note that we could still achieve this with the other approach by adding/removing compiler flags depending on what file is being compiled.
```

This current setup might be messing with @belericant 's tests, then, no?

But to your main point: I think we will probably have to abandon any forced precision flags. The reasoning is that we now have fine-grained control, which is what I would like. I'm not sure even compiling each module with different flags would do the trick here because some "setup" routines like `s_initialize_module_XXXXX` should presumably be done in higher precision (since it's only done once). Then we only use a lower precision in compute kernels.


```
One last remark, we may lose precision in a few places by being aggressive in the way we enforce working precision. Consider this example

x_cc(i) = x_domain%beg + (5._wp * (10._wp ** -(1)))*dx*real(2*i + 1, wp)
(5._wp * (10._wp ** -(1))) 

is a constant evaluated at compile-time. All operations that lead to its evaluation are done in working precision. It might be preferable to compute this in double precision and then cast it to single working precision for better fidelity.
```

@henryleberre If I understand correctly, there shouldn't be a meaningful difference as long as the computation is backward stable and not victim to catastrophic cancelation.
But I'm not sure I am fully understanding.

sbryngelson:
is this ready for merge? @belericant

sbryngelson:
we should also check performance on cpu and gpu, perhaps @wilfonba can pass along his script for that. i would also be curious to see the performance numbers on one of those cases using single precision on a V100 or A100

belericant:
> is this ready for merge? @belericant

@wilfonba found issues exporting 3d runs in single precision but that isn't due to the modifications necessary to have modular working precision. I don't know if it makes more sense for that to be fixed in this PR or in a new PR.

sbryngelson:
What happens if you use wp=single and precision (output) = 1 or 2, and if you use wp=double and precision = 1 or 2? Do any of these cases break? are they documented? @belericant @wilfonba

wilfonba:
@sbryngelson That's the logic I still need to add

sbryngelson:
Ok I will merge once that is complete.

wilfonba:
Numerical Schlieren doesn't seem to work with single working precision. When I run with double working precision and save single precision silo files it's nearly uniformly 1 after the first 10 time steps, but if I run with single precision and save in single precision it's uniformly zero after the first 10 time steps.

sbryngelson:
> Numerical Schlieren doesn't seem to work with single working precision. When I run with double working precision and save single precision silo files it's nearly uniformly 1 after the first 10 time steps, but if I run with single precision and save in single precision it's uniformly zero after the first 10 time steps.

https://github.com/belericant/MFC/blob/4134a660e77c20e0639f5565e768768728caa39f/src/post_process/m_derived_variables.fpp#L573-L718 here is the relevant code. It could be sensitive to precision issues, as well as the value of `schlieren_alpha(i)`. What do you use for these?

wilfonba:
@sbryngelson 
`'schlieren_alpha(1)'   : 40
  'schlieren_alpha(2)'  : 400`

sbryngelson:
I don't know if these are reasonable or not, I haven't used the feature in a while.

wilfonba:
It's one of the example cases, so I'd assume they're reasonable, but I can run it longer to see

wilfonba:
@sbryngelson Silo output works for all combinations now. For a 3D shock droplet with 2 million cells on 2 V100's with case optimization enabled, I get the following times per time-step (over a 1000 time-step run)

Double Precision: 3.288e-2 s/step (~109,500 steps/hr)
Single Precision: 2.499e-2 s/step (~144,000 steps/hr)

for a 31.5% speedup in steps/hr.

I did have to comment out the calls to `cufftexecd2z` and `cufftexecz2d` in `m_fftw.fpp` for it to compile on GPUs with NVHPC in single precision, works fine on CPUs with GNU though. It looked like on GPUs fftw3 was compiled after I got these errors, not sure what's up with that. Also, some of the test cases don't pass on double precision, likely a bug introduced somewhere in all the changes this pr has made or me merging it with upstream. None of the test cases pass with `wp = single_precision` because they all request double-precision silo files and those options are incompatible. Perhaps there's a way to modify the toolchain such that you pass the working precision as an option and `m_precision_select.f90` is automatically modified and changes all the test cases to use the correct precision silo files? @henryleberre or @anshgupta1234 could probably add this far more quickly than I can given they have experience with it.

sbryngelson:
> It looked like on GPUs fftw3 was compiled after I got these errors, not sure what's up with that.  

I've seen this recently as well. It doesn't build fftw when using GPUs (because I guess it finds cuFFT?) but it will build fftw for the post-processing code for CPU purposes.

> None of the test cases pass with wp = single_precision because they all request double-precision silo files and those options are incompatible.

I don't understand what you mean here. What options? We don't have silo golden files (though we do use h5dump to check for obviously incorrect hdf5/silo files).

> Perhaps there's a way to modify the toolchain such that you pass the working precision as an option and m_precision_select.f90 is automatically modified and changes all the test cases to use the correct precision silo files?

Presumably, we would like the precision to be selected in the `case.py` file. In this situation, we would just double the number of test cases, a set for single and a set for double precision.

wilfonba:
@sbryngelson 

> I don't understand what you mean here. What options? We don't have silo golden files (though we do use h5dump to check for obviously incorrect hdf5/silo files).< 
I have a check in m_checker that says single working precision and double precision silo files are incompatible because running a simulation in single precision and then requesting a double precision silo file doesn't make much sense. I could make it so that if single working precision is used and double silo precision is requested the single precision data is cast to double precision, but I think it makes more sense to just disallow that combination for practical purposes. I suppose allowing it would fix the testing problem.

>Presumably, we would like the precision to be selected in the case.py file. In this situation, we would just double the number of test cases, a set for single and a set for double precision.<
I see how this would be desirable, but I don't think it's as straightforward as other case file options. All of the working precision differences appear to take effect at compile time, not run time, while all current case file options are run time. I'm guessing trying to change precision at run time would cause a bunch of run time errors, but I can check.

I'll work on getting the test cases to pass given all the changes, and talk with Henry/Ansh about toolchain changes for testing.

wilfonba:
Came across a cool tool that takes math expressions and the expected ranges of each variable and outputs the best way to write the expression with the least loss of precision. May be useful if there are any calculations in the code that we might suspect to have problems with single precision.

https://herbie.uwplse.org/

sbryngelson:
Seems like it could be useful!

wilfonba:
Not sure why CI failed, I ran it on NVHPC using `./mfc.sh test -a --no-debug` and with GNU using both `./mfc.sh test -a --debug` and `./mfc.sh test -a --no-debug` and it passed all three times (using double precision).

PR #126, opened 2023-03-18
remove unused vars

sbryngelson:


sbryngelson:
this PR is just a partial fix, but its purpose is  
1. to remove situations some situations where a global scope variable overlaps with a local scope one (vscode calls this `Variable "X" masks variable in parent scope`)
2. to remove dead code

PR #137, opened 2023-04-02
Add support for counting lines of code via `./mfc.sh count`, remove `./mfc.sh cloc`

sbryngelson:
Add support for counting lines of code via `./mfc.sh count`, remove `./mfc.sh cloc` since it didn't support `.fpp` files. This PR still isn't great because it counts empty lines, but better than what we had before.


henryleberre (approved):
lgtm with my commit.

PR #138, opened 2023-04-03
Single precision fix

wilfonba:
Fixed single precision silo output, removed unnecessary coarsen silo feature, and cleaned up unused documentation. All test cases pass on CPUs and GPUs even with the modified compiler flags.


henryleberre (approved):
The parts I have jurisdiction over look good to me, provided @wilfonba is confident post-processing works as intended with both `precision=1` and `precision=2`.

sbryngelson (approved):

PR #141, opened 2023-04-06
Add Nsight Docs

anshgupta1234:
This PR adds  `--nsys` and `--ncu` to the documentation. I wasn't sure how much detail to add, please let me know if there's anything important that I missed :)

### DIFF
--- a/docs/documentation/running.md
+++ b/docs/documentation/running.md
@@ -89,6 +89,12 @@ modified by users.
 **Disclaimer**: IBM's JSRUN on LSF-managed computers does not use the traditional node-based approach to
 allocate resources. Therefore, the MFC constructs equivalent resource-sets in task and GPU count.
 
+### Profiling with NVIDIA Nsight
+
+MFC provides two different argument to facilitate profiling with NVIDIA Nsight. **Please ensure that the used argument is placed at the end so that their respective flags can be appended.**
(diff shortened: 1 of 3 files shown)


sbryngelson:
why did you change the CI file

anshgupta1234:
> why did you change the CI file

My apologies, I reverted it.

sbryngelson:
Please add links to what nsight compute and nsight sys are and their official documentation, and how to use additional flags for them. You also should describe what _they are used for_.

anshgupta1234:
@sbryngelson Sounds good, I'll add that.

henryleberre:
@sbryngelson @anshgupta1234 I pushed to this PR the feature that lets you specify additional arguments to `nsys` and `ncu`. For brevity, I won't explain why but it requires that `--nsys` (or `--ncu`) be specified as the last command line argument. You can just append to the invocation the arguments you want to have forwarded.

For example you could:

```console
$ ./mfc.sh run case.py [(optional) case file arguments] -N 5 -n 2 --gpu --no-mpi -j $(nproc) -t simulation -e batch .... --nsys [nsys arguments here]
```

This isn't necessarily realistic but it shows how you can use all the features together.

sbryngelson:
@henryleberre thanks! should be useful. can you add this to the documentation as well, @anshgupta1234 ?

sbryngelson (approved):

PR #142, opened 2023-04-10
Add intel compilers

### LINKED ISSUE #109: Intel oneapi compilers - add to CI matrix
Add Intel compilers to github actions OS matrix (ubuntu is fine) using something like [this](https://neelravi.com/post/oneapi-github-workflow/).

More examples that actually use this: https://github.com/AMReX-Codes/amrex/blob/b96f6e5bd13d156998edead7d92b21235bbea99d/.github/workflows/intel.yml 

Also see https://github.com/rscohn2/oneapi-ci

anshgupta1234:
Closes #109.

There was a weird SIGSEGV error, had to edit a loop in RHS to fix it.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,8 +23,18 @@ jobs:
         include:
           - os:  ubuntu-latest
             debug: --no-debug
+            intel: false
           - os:  ubuntu-latest
             debug: --debug
+            intel: false
+          - os:  ubuntu-latest
+            debug: --no-debug
+            intel: true
+            intel-command: source /opt/intel/oneapi/setvars.sh
+          - os:  ubuntu-latest
+            debug: --debug
+            intel: true
+            intel-command: source /opt/intel/oneapi/setvars.sh
           - os:  macos-latest
             debug: --no-debug
             gcc: 11
@@ -36,7 +46,7 @@ jobs:
       - name: Clone
         uses: actions/checkout@v3
 
-      - name: Setup
+      - name: Setup MacOS
         if:   matrix.os == 'macos-latest'
         run: |
           echo "OMPI_FC=gfortran-${{ matrix.gcc }}" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


sbryngelson:
This doesn't have any flags for intel compilers. what flags is it using? if `--debug` is taking longer than `--no-debug`, why? I would expect them to be the same other than a few print statements that are enabled with `--debug` via the `#MFC_DEBUG` ifdef.

sbryngelson:
Looking at `Cmakelists.txt`, I see that there are _no_ `-O3` or `-Ofast` or `-fast` flags in the release flags for _any_ of the compilers. @henryleberre can you comment on this?

anshgupta1234:
I just added -Ofast for GNU and -fast for Intel on another branch to check. It looks like the runtimes are nearly identical.

Example: GNU

- Without flags (49min): https://github.com/anshgupta1234/MFC/actions/runs/4670870704/jobs/8271134778
- With flags (48min): https://github.com/anshgupta1234/MFC/actions/runs/4679173580/jobs/8288843420

Code:
https://github.com/anshgupta1234/MFC/blob/intel-flags/CMakeLists.txt#LL72C1-L133C8

sbryngelson (approved):

henryleberre:
@sbryngelson I forgot to comment on this earlier. It is most likely unnecessary to supply these flags. In `--no-debug` mode we run CMake in `--config Release` which enables optimizations. You can take a look at `build/<target>/compile_commands.json` for more details. For examples, on my machine with GNU:

> /usr/bin/gfortran -DMFC_MPI -DMFC_PRE_PROCESS -I/usr/lib/x86_64-linux-gnu/fortran/gfortran-mod-15/openmpi -I/usr/lib/x86_64-linux-gnu/openmpi/lib -O3 -DNDEBUG -O3 -flto -fno-fat-lto-objects -ffree-line-length-none -fallow-invalid-boz -fallow-argument-mismatch -march=native -cpp -c /home/henryleberre/CPG/MFC/src/pre_process/autogen/m_variables_conversion.fpp.f90 -o CMakeFiles/pre_process.dir/src/pre_process/autogen/m_variables_conversion.fpp.f90.o

I would recommend only specifying special flags that CMake doesn't already provide. We already do this, but rarely for extra optimizations. One GPU option I think we should try out with NVIDIA is to set the GPU compute capability to be that of the cards we are currently targeting. We can't automatically determine this, however.

sbryngelson:
Thanks @henryleberre

PR #145, opened 2023-04-14
Update case.py

anandrdbz:
3D sphbub case fixed


sbryngelson:
Ok thanks, waiting on @wilfonba to determine where this issue first started arising.

wilfonba:
@sbryngelson I ran this case file in MFC 4.0.0 and it had the same behavior. NaNs after one time-step.

sbryngelson (approved):

PR #147, opened 2023-04-18
New CI using phoenix and slurm

sbryngelson:
New self-hosted CI, replacing wingtip that uses Phoenix and SLURM. Jobs are grabbed from one of multiple CI runners within a runner group on Phoenix. The advantages of the new workflow are that you are guaranteed a node to yourself (also useful for benchmarking) and multiple runners are available at once.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -98,86 +98,139 @@ jobs:
           ${{ matrix.intel-command }}
           /bin/bash mfc.sh test -j $(nproc) -a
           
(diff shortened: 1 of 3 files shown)


anshgupta1234 (approved):
The CPU case has a separate section for "Print" while the GPU doesn't--I don't think it matters but just wanted to point it out :)

Other than that, looking good!

sbryngelson:
@anshgupta1234 i don't think that's true... but there was a different mistake that I fixed.

PR #148, opened 2023-04-21
HLL + CBC fix

anandrdbz:
A fix for the hypoelastic case with HLL + CBC


sbryngelson:
@anandrdbz please explain what was wrong and what the solution was

anandrdbz:
The HLL Riemann solver + CBC needs flux indices from advxb to advxe but in the initialize and finalize routines in CBC,  the starting indices were 1, advxe instead of advxb, advxe leading to an illegal memory access.

sbryngelson (approved):

PR #150, opened 2023-04-27
Test Cases: 2D Taylor Green Vortex

haochey:


sbryngelson:
Please make this it's own patch like the isentropic vortex: https://github.com/MFlowCode/MFC/blob/222bb905301cc1d577c533b23c43ff9d5c0b9f0f/src/pre_process/m_patches.f90#L623-L673

You will have to change `assign_patch_variables` as well, which you can see forks here in the assign _mixture_ variables: https://github.com/MFlowCode/MFC/blob/222bb905301cc1d577c533b23c43ff9d5c0b9f0f/src/pre_process/m_assign_variables.f90#L156-L205 

However, this seems to be a bug because there is no equivalent statement in `s_assign_patch_species_primitive_variables_bubbles` or `s_assign_patch_species_primitive_variables`. Similar forks should be added for different "analytically" pre-defined cases we want on hand.

src/pre_process/m_patches.f90
```diff
@@ -940,6 +942,13 @@ subroutine s_2D_analytical(patch_id, patch_id_fp, q_prim_vf) ! -----------------
                     !END IF
                     ! ================================================================================
 
+                    ! Taylor Green Vortex===============================================
```
sbryngelson, line 1018: This shouldn't go here... it masks any other possible analytic patches someone might want to use and forces one to recompile to comment out the code to change it. Instead do something similar to the isentropic vortex case.

sbryngelson (changes requested):

src/pre_process/m_patches.f90
haochey: Ok, I will work on this.

sbryngelson:
Please get rid of these lines. Notice what they do: assign the variable `Pi` and `Gamma` to the primitive variables. This is only valid for the so-called Pi-Gamma model, where those variables are transported on the grid (`model_eqns == 1`). Instead, you should just eliminate these lines and let the user define the volume/void fractions themselves.

https://github.com/haochey/MFChaocheyfork/blob/6b7f3194a1df0b183dbf6aafc7db4078d0dc8dc0/src/pre_process/m_assign_variables.f90#L605-L609 

and here

https://github.com/haochey/MFChaocheyfork/blob/6b7f3194a1df0b183dbf6aafc7db4078d0dc8dc0/src/pre_process/m_assign_variables.f90#L814-L819

sbryngelson (approved):

Issue #149, opened 2023-04-27
`3d_turb_mixing` not working


sbryngelson:
Example case 3d turbulent mixing https://github.com/MFlowCode/MFC/blob/master/examples/3D_turb_mixing/case.py is not working on GPU with error:
```
[C]wingtip-gpu3: sbryngelson3/MFC $ ./mfc.sh run --gpu -n 2 ./examples/3D_turb_mixing/case.py -t pre_process
mfc: Found CMake: /nethome/sbryngelson3/MFC/build/cmake/bin/cmake.
mfc: OK > (venv) Entered the Python virtual environment.
      ___            ___          ___
     /__/\          /  /\        /  /\       sbryngelson3@wingtip-gpu3 [Linux]
    |  |::\        /  /:/_      /  /:/       ---------------------------------
    |  |:|:\      /  /:/ /\    /  /:/        --jobs 1
  __|__|:|\:\    /  /:/ /:/   /  /:/  ___    --no-mpi
 /__/::::| \:\  /__/:/ /:/   /__/:/  /  /\   --gpu
 \  \:\~~\__\/  \  \:\/:/    \  \:\ /  /:/   --no-debug
  \  \:\         \  \::/      \  \:\  /:/    --targets pre_process
   \  \:\         \  \:\       \  \:\/:/     ---------------------------------------------------
    \  \:\         \  \:\       \  \::/      $ ./mfc.sh [build, run, test, clean, count] --help
     \__\/          \__\/        \__\/

Run
  Acquiring ./examples/3D_turb_mixing/case.py...
  Configuration:
    Input               ./examples/3D_turb_mixing/case.py
    Job Name      (-#)  MFC
    Engine        (-e)  interactive
    Nodes         (-N)  1
    Tasks (/node) (-n)  2
    MPI Binary    (-b)  mpirun
  Generating input files...
  Building pre_process:

    $ cmake --build /nethome/sbryngelson3/MFC/build/pre_process --target pre_process -j 1 --config Release

[100%] Built target pre_process

    $ cmake --install /nethome/sbryngelson3/MFC/build/pre_process

-- Install configuration: "Release"
-- Installing: /nethome/sbryngelson3/MFC/build/install/bin/pre_process

  Running pre_process:

    $ /nethome/sbryngelson3/MFC/build/install/bin/pre_process

 Pre-processing a 255x255x255 case on 1 rank(s)
 x_cb.dat in  contains non-positive cell-spacings. Exiting ...
[...]

PR #151, opened 2023-05-01
Fixed a bug on pi_inf and enabled subgrid bubble model for viscous flow

lee-hyeoksu:
1. Fixed a bug that the capital 'pi_inf' was used where the little 'pi_inf' should be used. Also, computing speed of sound is also corrected.

2. Enabled subgrid bubble model for viscous flow
    - Added codes that computes the Reynolds number properly when 'bubbles=T'

3. Modified codes to make 'vel_profile' and 'instability_wave' options compatible with 'bubbles=T' and 'num_fluids=1'

4. Changed grid numbers in example/3D_turb_mixing to prevent memory issue (#149)


anandrdbz:
These changes seem fine, you might wanna change the definition of B_tait in m_qbmm as well. Other than that everything looks fine. The tests won't pass as of now due to this change as the golden files need to be regenerated for this definition of pi-inf

lee-hyeoksu:
Thanks @anandrdbz and @sbryngelson. The definition of `B_tait` in m_qbmm is now changed, and the golden files for the cases with sub-grid bubbles are updated.

sbryngelson:
@lee-hyeoksu your PR is failing GPU tests on the bubble cases I think.

lee-hyeoksu:
@sbryngelson The failure on GPU was because I didn't add `Re_L` and `Re_R` to the private clause for OpenACC when I added codes for computing `Re_L` and `Re_R` in `m_riemann_solvers`. Now it is fixed.

sbryngelson:
Great, thanks @lee-hyeoksu ! Can you also add _new_ tests for the case of viscous effects + subgrid bubbles? So we can track if code associated with those effects changes or breaks. Again, there should be some documentation on this (adding new tests).

lee-hyeoksu:
@sbryngelson I added new tests, and they appear in the list by `./mfc.sh test -l`. But is there anywhere I should add documentation?

sbryngelson (approved):

PR #155, opened 2023-05-22
Fix the computation of vcfl

haochey:
Correct the computation of vcfl with dimensional input cases.


sbryngelson:
@haochey this PR failed tests. Please run `./mfc.sh test` on your computers before submitting a PR.

anandrdbz:
@haochey replace alpha_rho with rho in vcfl computation

haochey:
> @haochey replace alpha_rho with rho in vcfl computation, that should do it

@anandrdbz Thanks for the help

sbryngelson (approved):

PR #158, opened 2023-05-24
Documentation for qbmm

anandrdbz:
Documentation added for the relevant qbmm variables


anandrdbz:
@sbryngelson , test fails due to the intel compiler issues as I only changed documentation

PR #172, opened 2023-06-16
Add weno_avg back

haochey:
weno_avg is added along with a bunch of new golden files with weno_Re_flux turned on. "weno_avg" was on the case_dict so no new variable need to be added in case_dicts.py


sbryngelson:
Hi @haochey -- will review this PR when the tests pass.

haochey:
Hi @sbryngelson, it looks like all the tests pass with some some fix. Thanks!

sbryngelson:
Thanks, @haochey ! Can you "prove" that the `weno_avg` feature is working as expected? Perhaps by showing results with it turned on and turned off? I suggest adding your 2D Taylor-Green vortex and Blausius boundary layer problems to the `examples/` folder and using those as examples to "test" the feature.

haochey:
@sbryngelson Yes, I can do that. For "examples", the results of 2D_Viscous and 2D_TaylorGreenVortex with weno_avg turned on and off are attached. Both of them are in the folder of examples. I will upload the updated case.py for both of them. Should the Blasius boundary layer problems need the No-slip boundary condition which I am still merging it with the main branch.


https://github.com/MFlowCode/MFC/assets/98496194/10f4fbbe-be57-4df9-a506-47f1d6368b17

https://github.com/MFlowCode/MFC/assets/98496194/983b7ca2-e152-496b-bc3f-f076286b334e


https://github.com/MFlowCode/MFC/assets/98496194/7d5c82b0-17b9-4a65-bee6-52d916e3c79d


https://github.com/MFlowCode/MFC/assets/98496194/8aee7712-53d6-4510-86c2-8b5eaefac743

haochey:
Hi @sbryngelson, the attached animations are two 3D two-component cases (one in cartesian and the other in cylindrical coordinates). They appear to match the expectation as the mixing layer shows up at the interface of velocity discontinuity.  

https://github.com/MFlowCode/MFC/assets/98496194/ba86deee-150e-4307-86f2-b7dbd3ebb798


https://github.com/MFlowCode/MFC/assets/98496194/4ac7b534-2c78-4492-b6c9-99a85431d125

sbryngelson:
@haochey thanks, looks good. Just need this part now:

> For this PR, I ask that you go through the other case.py files in examples/ and add the additional dictionary line weno_avg = 'F' (or weno_avg = 'T' if appropriate) so that folks know that this feature/option is available. We should also add the other viscous-related input variables (like weno_re_flux) to these example files, so we are enumerating all options for each case file.

henryleberre:
@haochey Could you consider pushing less often? It triggers CI every time and can become quite costly. GitHub runner hours are free but Phoenix GPU hours might not (@sbryngelson ?). In any case, it creates a backlog of commits to run CI on. I took the liberty to cancel a few CI runs (@sbryngelson).

haochey:
> @haochey Could you consider pushing less often? It triggers CI every time and can become quite costly. GitHub runner hours are free but Phoenix GPU hours might not (@sbryngelson ?). In any case, it creates a backlog of commits to run CI on. I took the liberty to cancel a few CI runs (@sbryngelson). As a fun fact, I also get a tap on the wrist whenever your commit (an Outlook notification).

Hi @henryleberre , I am sorry. I will keep this in mind in the future.

henryleberre:
No worries, @haochey! You can also commit & push to another branch and merge your changes in later once you're done. I've encountered this problem myself recently.

sbryngelson:
Hi @haochey yes this is important - please submit your commits to a branch rather than the PR (or some other solution). Otherwise, everytime you commit you are launching another CI run that uses computer resources.

examples/1D_poly_bubscreen/case.py
```diff
@@ -106,6 +106,8 @@
     'time_stepper'                 : 1,
     'weno_order'                   : 1,
     'weno_eps'                     : 1.E-16,
+    'weno_Re_flux'                 : 'T',
```
sbryngelson: this example does not have viscosity, make these both 'F'

haochey: Hi @sbryngelson, thanks for pointing out it. It's resolved.

haochey:
Hi @sbryngelson , thanks for your advise. I will learn the features and make some fixes.

sbryngelson:
@haochey I'm not sure why but the test golden files you have created have the variables all in a random order. See this example https://github.com/MFlowCode/MFC/blob/master/tests/9DAC4DDC/golden.txt for how things should look. It is also showing a diff on a goldenfile that already existed (there should be no difference though).

henryleberre:
@haochey Once you are done, you could maybe consider squashing these commits together (or into a few different ones) so as not to add 100+ commits.

haochey:
Hi @sbryngelson, I corrected the golden files. The check for the PACE phoenix seems to be cancelled. Hopefully this will work. I can try to squash the commits together once we agree it's done as suggested by @henryleberre.

sbryngelson:
It looks like PACE is under maintenance.

haochey:
Hi @sbryngelson, the gpu test failed. I am not quite sure if this is affected by the outage this morning. Do you have any suggestions on how to fix it?

sbryngelson:
it failed on the first test you added. so I don't think this is associated with the phoenix outage. reasons for failure could include differences between compilers when generating test cases. if you run MFC in GPU mode for that test on for example Phoenix, do you get the same results as what is showing up in the CI test suite?

haochey:
> it failed on the first test you added. so I don't think this is associated with the phoenix outage. reasons for failure could include differences between compilers when generating test cases. if you run MFC in GPU mode for that test on for example Phoenix, do you get the same results as what is showing up in the CI test suite?

Yes! I got the same result for this case. I also tested the rest of the cases with weno_avg added and I got the same issue.

sbryngelson:
> > it failed on the first test you added. so I don't think this is associated with the phoenix outage. reasons for failure could include differences between compilers when generating test cases. if you run MFC in GPU mode for that test on for example Phoenix, do you get the same results as what is showing up in the CI test suite?
> 
> Yes! I got the same result for this case. I also tested the rest of the cases with weno_avg added and I got the same issue.

What do you mean _same result_. The same result as the CI gets? Can you create the goldenfiles using nvhpc in GPU mode on Phoenix and, for example, GNU in CPU mode on your Mac? How different are they?

haochey:
> > > it failed on the first test you added. so I don't think this is associated with the phoenix outage. reasons for failure could include differences between compilers when generating test cases. if you run MFC in GPU mode for that test on for example Phoenix, do you get the same results as what is showing up in the CI test suite?
> > 
> > 
> > Yes! I got the same result for this case. I also tested the rest of the cases with weno_avg added and I got the same issue.
> 
> What do you mean _same result_. The same result as the CI gets? Can you create the goldenfiles using nvhpc in GPU mode on Phoenix and, for example, GNU in CPU mode on your Mac? How different are they?

Yes, I ran the tests myself on Phoenix and I got the same results (i.e. the candidate values) as what the CI gets. The values in golden files created by GPU and CPU look the same in terms of the printed decimal precision. This makes sense because the output of the tests shows the difference between the candidate value and the target value is O(1e-12).

Golden Files

sbryngelson:
@haochey something doesn't seem quite right here, the results (primitive and conservative variables) after 50 time steps are exactly the same...

haochey:
> @haochey something doesn't seem quite right here, the results (primitive and conservative variables) after 50 time steps are exactly the same as the initial condition...

@Spencer: The other case **C79E1D3C** (with both weno_Re_flux and weno_avg turned off) shows the same result. https://github.com/MFlowCode/MFC/blob/master/tests/C79E1D3C/golden.txt

I think this is caused by the initial conditions where velocities are set to be zero in these test cases.

sbryngelson:
> > @haochey something doesn't seem quite right here, the results (primitive and conservative variables) after 50 time steps are exactly the same as the initial condition...
> 
> @spencer: The other case **C79E1D3C** (with both weno_Re_flux and weno_avg turned off) shows the same result. [`master`/tests/C79E1D3C/golden.txt](https://github.com/MFlowCode/MFC/blob/master/tests/C79E1D3C/golden.txt?rgh-link-date=2023-06-29T01%3A26%3A10Z)
> 
> I think this is caused by the initial conditions where velocities are set to be zero in these test cases. Except for this 1D case, values are different between the initial condition and 50 times steps in the 2D and 3D cases (e.g. 7DEA8C54).

Yikes! That does not sound like a very good test...

henryleberre:
@sbryngelson To address one of your comments, the ordering of .dat files in golden/pack files used not to be deterministic. I actually *fixed* this recently:
https://github.com/MFlowCode/MFC/blob/f2bee499d15bed6bc037c806403e12c59b36fbe0/toolchain/mfc/test/pack.py#L92-L93

It wasn't a _bug_ per se because it wouldn't interfere with testing since the logic doesn't assume ordering.

haochey:
> @sbryngelson To address one of your comments, the ordering of .dat files in golden/pack files used not to be deterministic. I actually _fixed_ this recently:
> 
> https://github.com/MFlowCode/MFC/blob/f2bee499d15bed6bc037c806403e12c59b36fbe0/toolchain/mfc/test/pack.py#L92-L93
> 
> It wasn't a _bug_ per se because it wouldn't interfere with testing since the logic doesn't assume ordering.

@henryleberre Yes! I merged your recent commits and it fixes the order of the printed variables. I also noticed that this PR could "squash and merge" all the commits into a single commit to avoid 100+ commits. I think this would work for this time, but I will be careful with being concise in the future! Thanks for your suggestion.

@sbryngelson It looks like the difference between compilers do affect the test cases on GPU.

sbryngelson:
@henryleberre this is useful info thanks. Good to know. @haochey any test that doesn’t product a difference between initial condition and output should be changed so that they are meaningfully different. Then we can get into compiler specific issues.

haochey:
> @henryleberre this is useful info thanks. Good to know. @haochey any test that doesn’t product a difference between initial condition and output should be changed so that they are meaningfully different. Then we can get into compiler specific issues.

@sbryngelson For any 1D cases, I am not sure how to show viscosity properly (how to allow flow to deviate from its initial condition) because the flow properties are assumed to only vary in one direction --the effects of viscosity would theoretically be zero because there are no velocity gradients in the perpendicular direction for the viscosity to work upon.

sbryngelson:
> > @henryleberre this is useful info thanks. Good to know. @haochey any test that doesn’t product a difference between initial condition and output should be changed so that they are meaningfully different. Then we can get into compiler specific issues.
> 
> @sbryngelson For any 1D cases, I am not sure how to show viscosity properly (how to allow flow to deviate from its initial condition) because the flow properties are assumed to only vary in one direction --the effects of viscosity would theoretically be zero because there are no velocity gradients in the perpendicular direction for the viscosity to work upon.

This isn't true -- the Laplacian of the velocity isn't zero when you go to 1D, it's just $$d^2 u / dx^2$$ which is non-zero so long as $u$ doesn't have zero curvature.

examples/3D_turb_mixing/case.py
sbryngelson, line 1: please remove changes to this file other than `weno_avg` and then tell @lee-hyeoksu what you found, thanks!

sbryngelson (changes requested):

docs/documentation/case.md
```diff
@@ -331,6 +333,10 @@ Note that `time_stepper` $=$ 3 specifies the total variation diminishing (TVD),
 - `wave_speeds` specifies the choice of the method to compute the left, right, and middle wave speeds in the Riemann solver by an integer of 1 and 2.
 `wave_speeds` $=$ 1 and 2 correspond to the direct method ([Batten et al., 1997](references.md#Batten97)), and indirect method that approximates the pressures and velocity ([Toro, 2013](references.md#Toro13)), respectively.
 
+- `weno_Re_flux` activates the scaler divergence theorem in computing the velocity gradients using WENO-reconstructed cell boundary values.
+
+- `weno_avg` activates the arithmetic average of the left and right, WENO-reconstructed, cell-boundary values.
```
sbryngelson: add text here that says that this option requires `weno_Re_flux` to be true (and why).

docs/documentation/case.md
```diff
@@ -331,6 +333,10 @@ Note that `time_stepper` $=$ 3 specifies the total variation diminishing (TVD),
 - `wave_speeds` specifies the choice of the method to compute the left, right, and middle wave speeds in the Riemann solver by an integer of 1 and 2.
 `wave_speeds` $=$ 1 and 2 correspond to the direct method ([Batten et al., 1997](references.md#Batten97)), and indirect method that approximates the pressures and velocity ([Toro, 2013](references.md#Toro13)), respectively.
 
+- `weno_Re_flux` activates the scaler divergence theorem in computing the velocity gradients using WENO-reconstructed cell boundary values.
```
sbryngelson: add text here that says that if this is _false_ then the code will use finite differences of order X (please fill in what X is).

sbryngelson: X depends on? .... order of weno? fd_coef? ?

sbryngelson (changes requested):

sbryngelson (changes requested):
...

haochey:
Hi @sbryngelson, I made the requested changes. Please check if it's ready for merge. Thanks!

sbryngelson (approved):

PR #183, opened 2023-07-13
Minor bug fix with polytropic qbmm and cleaning up p_main

anandrdbz:
A bug with qbmm + polytropic + We /= dflt_real that pertains to non-dimensionalization (this was done in pre_process but forgot to do it in simulation


anandrdbz:
Phoenix is down at the moment but the rest of the tests passed with the exception of one which got cancelled for some reason, I think it should be okay to merge at this point.

anandrdbz:
@sbryngelson I have added the cleaning up of p_main into this PR for pre_process, simulation and post_process. All of the files now have <100 lines and the tests pass locally and on GPUs on Bridges2

sbryngelson:
Why is this called `m_clean`? Does "clean" mean something physically or computationally? Please use more proper naming.

anandrdbz:
I couldn't think of an appropriate name for it as it performs seemingly unrelated tasks, I went with clean because it cleaned up p_main physically (computationally there's no difference). If you have some suggestions do let me know, I can think of m_setup perhaps, but not sure if that's totally appropriate either

sbryngelson (approved):

Issue #187, opened 2023-07-17
Put all boundary conditions in a module called `m_boundary_conditions.fpp` 


sbryngelson:
Right now, we have some boundary conditions in `m_rhs`, which is not really the appropriate place for them. We also have the CBCs in a different module, which makes is surprising to find some BCs in `m_rhs` as well. For example, in `m_rhs`:

https://github.com/MFlowCode/MFC/blob/1aefc097fea9acff82819634a161114b9411d5fe/src/simulation/m_rhs.fpp#L2146-L3110

Please refactor this into a different module like `m_boundary_conditions` that `m_rhs` uses and calls. We can leave the CBCs in `m_cbc`, which would then be called by `m_boundary_conditions`.

We also have to be careful about the logic here, for example, some non-CBC boundary conditions might accidentally end up being called by the CBC module via the logic in the code like this:
https://github.com/MFlowCode/MFC/blob/1aefc097fea9acff82819634a161114b9411d5fe/src/simulation/m_rhs.fpp#L877-L885

We have to re-organize the boundary conditions so that this logic is not a problem or is otherwise resolved somehow.

sbryngelson:
Also relevant! https://github.com/anandrdbz/MFC/blob/9140e50e9a99bba74d68956fe18799b9c3fdce4d/src/simulation/m_start_up.fpp#L561-L579

Issue #185, opened 2023-07-17
Adding boundary conditions that already exist in forks


sbryngelson:
Several boundary conditions have been added to MFC in various forks. 
I would like to merge these via isolated PRs as soon as possible. 

For example:
- [x] @wilfonba non-characteristic no-slip wall
- [x] @wilfonba non-characteristic slip wall (also please fix documentation on current slip wall)
- [x] characteristic slip wall
- [ ] @haochey characteristic no-slip wall
- [x] @anshgupta1234 inlet boundary condition (or was this @haochey)

Be careful that if you give your BC a number that is `<= -5` then the code will interpret it as a CBC, which you might not want:
https://github.com/MFlowCode/MFC/blob/1aefc097fea9acff82819634a161114b9411d5fe/src/simulation/m_rhs.fpp#L877-L885

We have to re-organize the boundary conditions so that this logic is not a problem or is otherwise resolved somehow.

Related to issue #187 and #186

sbryngelson:
Also relevant! 
https://github.com/anandrdbz/MFC/blob/9140e50e9a99bba74d68956fe18799b9c3fdce4d/src/simulation/m_start_up.fpp#L561-L579

wilfonba:
Non-characteristic no-slip and slip walls have been added and merged to the master branch.

sbryngelson:
@haochey can you post the status of characteristic no-slip walls? Or did they have that specific corner problem making them challenging to use?

sbryngelson:
@wilfonba do we have a characteristic slip wall, or no?

wilfonba:
Yes, characteristics slip walls have been in the code for a while now.

haochey:
> @haochey can you post the status of characteristic no-slip walls? Or did they have that specific corner problem making them challenging to use?

Characteristic no-slip wall works in 2D by treating the corners as slip wall nevertheless it's not perfect because the velocity is not perfectly zero in the wall. It's shown in the animation below. This method does not resolve the corner/edge issue in 3D (i.e. cases still crashing in 3D). 

https://github.com/MFlowCode/MFC/assets/98496194/89c58ad7-0a67-4d95-bf72-94d37107ce54

sbryngelson:
Thanks @haochey I am closing this issue.

PR #189, opened 2023-07-18
Boundary conditions refactor and addition of non-characteristic slip and no-slip walls

wilfonba:
This takes care of #187 and part of #185 . It moves all of the boundary conditions stuff that isn't characteristic to a new module `m_boundary_conditions` that is called from `m_rhs`. I don't think moving CBCs to this module makes much sense since those are utilized in `m_rhs` in such a different way. All old test cases pass and 6 new test cases have been added for the new boundary conditions. I haven't yet verified the new BCs, but wanted this code to be available to @anshgupta1234 and @haochey so they had an idea what things will look like soon. Also, if someone could take a careful look at my changes the the `m_checker` modules to make sure I didn't mess up any of the checks, particularly when cylindrical coordinates are being used, that'd be greatly appreciated.


sbryngelson:
Lines of code in simulation/ doubled here, adding 900 lines for an 1800-line long m_boundary_conditions. How did this happen/is it needed? It seems like much could be refactored there.

wilfonba:
~650 of the additional lines are a result of the new boundary conditions I added, ~60 lines are from adding missing QBMM assignments to the symmetry case for `bc_x%beg` and `bc_x%end`, and the rest is just added white space and a change in the logic. The logic in the `s_populate_conservative_variables_buffers` is more or less repeated in all of the subroutines for each individual BC because this makes calls to those subroutines cleaner.

sbryngelson:
@wilfonba looks like this is mostly good to go - rectify the merge conflicts and then do the minor refactor I believe you mentioned earlier today with `qbmm`. I don't want to actually merge this one until the GPU check is fixed.

sbryngelson:
> I am wondering if we can use two-digit numbers (>-100) to name normal boundaries and three-digit numbers (<-100) to name characteristic boundaries. This approach helps avoid confusion between newly added boundary conditions and existing ones, and also distinguishes between normal and characteristic boundaries for coding logics in cases like "if (bc_x < -100) then ...".

@haochey good thinking. This seems reasonable to my eye. I'll let @wilfonba tell us if there's anything inherently wrong/problematic/annoying about doing this.

wilfonba:
@sbryngelson This would make the code easier to read. The only down side is that it breaks every input file ever made. I'm not sure how beat to communicate with all relevant parties the change has been made to avoid frustrations.

sbryngelson:
> @sbryngelson This would make the code easier to read. The only down side is that it breaks every input file ever made. I'm not sure how beat to communicate with all relevant parties the change has been made to avoid frustrations.

This is true. For now, let's stick to a solution that doesn't break old input files. This is an easy change later if appropriate.

wilfonba:
@sbryngelson @haochey Might it be best to rename -5 to wall/solid-wall, and than have an if condition in the code that selectes no-slip wall if the problem is viscous and selects slip wall if the problem isn't? This would maintain all existing boundary conditions, including the seeimingly undocumented -13.

sbryngelson:
> @sbryngelson @haochey Might it be best to rename -5 to wall/solid-wall, and than have an if condition in the code that selectes no-slip wall if the problem is viscous and selects slip wall if the problem isn't? This would maintain all existing boundary conditions, including the seeimingly undocumented -13.

Better would be to abort the case via `m_checker` if viscous + slip or inviscid + no-slip are chosen.

wilfonba:
As long as this passes the tests this time, @haochey is going to make a PR to my fork that I'll then merge since this branch already has the logic for the additional boundary condition. Once he or I resolve merge conflicts and I add @haochey or @anshgupta1234 simple inflow condition (it's working well enough for them now), this should be ready to go.

src/simulation/m_cbc.fpp
```diff
@@ -418,19 +421,115 @@ contains
...
                 ! ==================================================================
 
                 ! Computing CBC2 Coefficients ======================================
             elseif (weno_order == 3) then
 
-                fd_coef_${XYZ}$(:, cbc_loc_in) = 0d0
-                fd_coef_${XYZ}$(0, cbc_loc_in) = -6d0/(3d0*ds(0) + 2d0*ds(1) - ds(2))
-                fd_coef_${XYZ}$(1, cbc_loc_in) = -4d0*fd_coef_${XYZ}$(0, cbc_loc_in)/3d0
-                fd_coef_${XYZ}$(2, cbc_loc_in) = fd_coef_${XYZ}$(0, cbc_loc_in)/3d0
+                fd_coef_${XYZ}$(:, 1, cbc_loc_in) = 0d0
```
sbryngelson, line 481: please add comments to a paper/link for where these coefficients come from

src/simulation/m_cbc.fpp
```diff
@@ -418,19 +421,115 @@ contains
...
+                    if (m*n*p > 0) then
+                    ! Trv2
+                        fd_coef_${XYZ}$(:, 3, cbc_loc_in) = 0d0
+                        fd_coef_${XYZ}$(0, 3, cbc_loc_in) = -1d0/(dz(0) + dz(1))
+                        fd_coef_${XYZ}$(1, 3, cbc_loc_in) = 0d0
+                        fd_coef_${XYZ}$(2, 3, cbc_loc_in) = -fd_coef_${XYZ}$(0, 3, cbc_loc_in)
+                    end if
+                
+                elseif (cbc_dir_in == 3) then
+                    ! Trv1
```
sbryngelson, line 519: trv1?

src/simulation/m_cbc.fpp
```diff
@@ -418,19 +421,115 @@ contains
...
+                    fd_coef_${XYZ}$(:, 2, cbc_loc_in) = 0d0
+                    fd_coef_${XYZ}$(0, 2, cbc_loc_in) = -1d0/(dy(0) + dy(1))
+                    fd_coef_${XYZ}$(1, 2, cbc_loc_in) = 0d0
+                    fd_coef_${XYZ}$(2, 2, cbc_loc_in) = -fd_coef_${XYZ}$(0, 2, cbc_loc_in)
+
+                    if (m*n*p > 0) then
+                    ! Trv2
```
sbryngelson, line 526: ?

src/simulation/m_cbc.fpp
```diff
@@ -418,19 +421,115 @@ contains
...
+                    end if
+                
+                elseif (cbc_dir_in == 3) then
+                    ! Trv1
+                    fd_coef_${XYZ}$(:, 2, cbc_loc_in) = 0d0
+                    fd_coef_${XYZ}$(0, 2, cbc_loc_in) = -1d0/(dy(0) + dy(1))
+                    fd_coef_${XYZ}$(1, 2, cbc_loc_in) = 0d0
+                    fd_coef_${XYZ}$(2, 2, cbc_loc_in) = -fd_coef_${XYZ}$(0, 2, cbc_loc_in)
+
+                    if (m*n*p > 0) then
```
sbryngelson, line 525: what does this condition mean? please add more comments

Issue #188, opened 2023-07-17
Create `m_analytic_patches` in `pre_process/` for initial condition edge cases


sbryngelson:
In situations where the analytic features (thanks to #176) built into the `input.py` files are not sufficient to represent a case, due to weird geometries (or whatever), we can have a separate module that accommodates these. It could be called `m_analytic_patches`, for example. This is an alternative to having many hard-coded cases living in `s_analytic_patch_2D` even when #176 is available.

sbryngelson:
closed by #201 and #188

PR #197, opened 2023-07-28
Address #188

wilfonba:
Address #188. Not necessarily an `m_analytic_patches` module, but I think this approach better separates the definition of hard coded patches from the code used to assign the initial condition.


sbryngelson:
You have an inconsistency with
```f90
    if (patch_icpp%hcid == 201) then
        ! Primitive variables assignment
    end if
```
in the documentation but 

```
    if (patch_icpp(patch_id)%hcid == 200) then

    end if
```
in the actual code.

PR #201, opened 2023-07-31
Hardcoded ICs Update

wilfonba:
Added example case, added error checking, and switched to `select case` from `if... elseif` in `1[2,3]DHardcodedIC.fpp`.


wilfonba:
Also fully closes #188

sbryngelson (approved):

Issue #177, opened 2023-06-27
Clean up isentropic vortex, turn into `example/`


sbryngelson:
A super easy issue: Create a big red diff by leveraging the merged PR #176 to get rid of the hard-coded isentropic vortex case of [this paper](https://authors.library.caltech.edu/49575/7/nihms606496.pdf) (fig 4 sec 4.4) that lives in at least three different places as `patch_icpp(patch_id)%geometry == 6`. In the process can also kill the documentation we have on this patch type (if it exists). Here are some code snippets:

https://github.com/MFlowCode/MFC/blob/aad787bf90a371e87aa441d6e8ac3ff019cf2625/src/pre_process/m_assign_variables.f90#L156-L205

https://github.com/MFlowCode/MFC/blob/aad787bf90a371e87aa441d6e8ac3ff019cf2625/src/pre_process/m_assign_variables.f90#L570-L609

https://github.com/MFlowCode/MFC/blob/aad787bf90a371e87aa441d6e8ac3ff019cf2625/src/pre_process/m_assign_variables.f90#L773-L812

RasmitDevkota:
Once the isentropic vortex code becomes an example case, would it make sense to get rid of vortex geometry 6 in its entirety as it isn't necessary for simulating such scenarios?

sbryngelson:
Yes, that's the goal.

sbryngelson:
Closed by #205

PR #205, opened 2023-08-12
Converted isentropic vortex to example case

RasmitDevkota:
Converted isentropic vortex to example case, resolving issue #177, using [Coralic and Colonius](https://authors.library.caltech.edu/49575/7/nihms606496.pdf) and [Yee, Sandham, and Djomehri](https://ntrs.nasa.gov/api/citations/19980201047/downloads/19980201047.pdf). Most notably, patch type 6 now results in an abort with an appropriate message, and density and pressure have been reformulated due to inconsistencies with references. Since this is a stationary problem, the initial condition and output at any given time are the same—visualizations instead display key properties of the vortex.


src/common/m_derived_types.fpp
```diff
@@ -122,9 +122,7 @@ module m_derived_types
 
         type(ic_model_parameters) :: model !< Model parameters
 
-        real(kind(0d0)) :: epsilon, beta !<
-        !! The isentropic vortex parameters administrating, respectively, both
-        !! the amplitude of the disturbance as well as its domain of influence.
+        real(kind(0d0)) :: epsilon, beta
```
sbryngelson: what are these variables still used for? i think the might still be used elsewhere. can you document it like in the doxygen hook before?

src/pre_process/m_initial_condition.fpp
```diff
@@ -190,9 +196,13 @@ contains
                 elseif (patch_icpp(i)%geometry == 5) then
                     call s_ellipse(i, patch_id_fp, q_prim_vf)
 
-                    ! Isentropic vortex patch
+                    ! Unimplemented patch
                 elseif (patch_icpp(i)%geometry == 6) then
-                    call s_isentropic_vortex(i, patch_id_fp, q_prim_vf)
+                    call s_int_to_str(i, iStr)
+
+                    call s_mpi_abort('Unimplemented choice of geometry 6'// &
```
sbryngelson: you have two calls to `mpi_abort` for the same thing, one in `m_initial_condition` and one in `m_check_patches`.... only one of these needs to exist (whichever the code calls first). the second can be turned into a code comment.

sbryngelson (commented):
Also, can you show an image where you post_process the vorticity? it has a hook in post_process called `omega_wrt(1/2/3` I believe...

src/common/m_derived_types.fpp
RasmitDevkota: They appear to be spherical harmonics parameters - I have documented them as such.

src/pre_process/m_initial_condition.fpp
RasmitDevkota: Removed the one in `m_initial_condition` since it occurs second.

src/common/m_derived_types.fpp
sbryngelson: you need the `!<` part

examples/2D_isentropicvortex/omega_wrt3.png
sbryngelson, line 1: are these the correct vorticities? compared to the paper?

examples/2D_isentropicvortex/vel1.png
sbryngelson, line 1: can remove this

examples/2D_isentropicvortex/vel2.png
sbryngelson, line 1: can remove this

examples/2D_isentropicvortex/alpha_rho1.png
sbryngelson, line 1: this doesn't seem to match fig 4 of coralic and colonius 2014

RasmitDevkota: Figure 4 uses the coarsest possible grid, 25x25, for visualization, whereas the numerical computations are done using a much more precise grid, 400x400, which is what the case file uses.

examples/2D_isentropicvortex/omega_wrt3.png
RasmitDevkota: The original paper does not contain any vorticity plot. No other papers on this particular problem seem to mention it either.

examples/2D_isentropicvortex/alpha_rho1.png
sbryngelson: does the initial condition match the one in the paper at that resolution?

examples/2D_isentropicvortex/case.py
```diff
@@ -0,0 +1,107 @@
...
+    'patch_icpp(1)%vel(1)'         : vel1,
+    'patch_icpp(1)%vel(2)'         : vel2,
+    'patch_icpp(1)%pres'           : pres,
+    'patch_icpp(1)%rho'            : 1,
+    'patch_icpp(1)%alpha_rho(1)'   : alpha_rho1,
+    'patch_icpp(1)%alpha(1)'       : 1.,
+    # ==========================================================================
+
+    # Fluids Physical Parameters ===============================================
+    'fluid_pp(1)%gamma'            : 1.4,
```
sbryngelson: this is not the gamma you want, it is actually capital Gamma. see Coralic and Colonius 14 (and the code documentation).

sbryngelson (changes requested):

examples/2D_isentropicvortex/alpha_rho1.png
RasmitDevkota: Given all the same parameters for the 25x25 case, the fluxes, velocities, and densities diverge at this low resolution (returning NaNs at timesteps). I've traced this issue back to `q(L|R)_prim_rs`, though it is not clear yet how it sometimes gains a NaN value.

Another relevant general issue is that MFC only has up to the third-order TVD RK time-steppers while the paper uses a fifth-order time-stepper (which it notes as a limiting factor on the accuracy).

examples/2D_isentropicvortex/case.py
RasmitDevkota: I reverted this back

examples/2D_isentropicvortex/alpha_rho1.png
sbryngelson: this might be because they used a nonuniform grid size, or you have the time step size wrong, or the time stepper itself. however, the reason for 5th order time stepper has to do with those accuracy plots on the right of the figure (matching 5th order accurate weno with 5th order accurate runge kutta). the simulation should be similarly stable for 3rd order TVK runge kutta, though perhaps you need a slightly smaller time step.

if a slightly smaller time step doesn't fix the problem then i think there is likely a different problem. also, does the IC for the problem you describe match the one in the paper? (the \rho field, that is?)

PR #213, opened 2023-09-12
Boundary conditions refactor

### LINKED ISSUE #187: Put all boundary conditions in a module called `m_boundary_conditions.fpp`

wilfonba:
This closes #187 and partially completes #185.


sbryngelson:
@wilfonba can you run a few example cases comparing run time on GPU like we've done in the past?

src/simulation/m_boundary_conditions.fpp
```diff
@@ -0,0 +1,1440 @@
...
+                end do
+
+                if(qbmm .and. .not. polytropic) then
+                    !$acc parallel loop collapse(5) gang vector default(present)
+                    do i = 1, nb
+                        do j = 1, buff_size
+                            do l = -buff_size, n + buff_size
+                                do k = -buff_size, m + buff_size
+                                    do q = 1, nnode
+                                        pb(k, l, p+j, q, i) = &
```
sbryngelson, line 352: some of these loop orderings don't make sense to me, should they not go from outer to inner index, so in this case the loops would go from outer to inner as i, q, j, l, k?

wilfonba: @anandrdbz I believe what I have here is the same as what was in your initial PR for this feature.

sbryngelson: perhaps, but still requires explaining (some of the other loops also have this issue)

anandrdbz: I think these kernels weren't optimized to the fullest because they are orders of magnitude faster than the critical points as they only loop through two of the three dimensions at most. But yes, this might show a slight benefit by either using the q loop outside or adding a '!$acc loop seq' in front of the q while preserving the same order. (collapse(3) should be changed to collapse(2) in this case)

anandrdbz: However, it should make no meaningful difference to the overall run time for a single time step

wilfonba: These have been reordered outer to inner with a 0.13% performance improvement for the test case I ran.

sbryngelson: understood, what @anandrdbz says makes sense.

sbryngelson:
> @wilfonba can you run a few example cases comparing run time on GPU like we've done in the past?

Only thing before merging is this part

wilfonba:
- Run-times between master and this branch on GPUs for a 100x100x100 test case ran on 2 V100s. Only one characteristic BC is included because all characteristic boundary conditions utilize ghost-cell extrapolation at the boundary. The following table shows seconds/time-step and % difference between this branch and the master branch.

| BC# | Type | MFC-Wilfong | Master | % Difference
|--------|--------|--------|--------|--------|
|  -1 | Normal | 0.0641 | 0.0636 | 0.75% |
| -2 | Normal | 0.0309 | 0.0309 | -0.007% |
| -3 | Normal | 0.0309 | 0.0308 | 0.08% |
| -4 | Normal | 0.0316 | 0.0315 | 0.18% |
| -5 | Normal | 0.0355 | 0.0355 | 0.12% |

- Run time for different loop ordering in the QBMM boundaries for a 100x100x100 case ran on 2 V100s. The following table shows seconds/time-step and the % difference for different loop orderings.

| i,j,l,k,q | i,q,j,l,k | % Difference |
|--------|--------|--------|
| 0.1789 | 0.1786 | 0.13% |

sbryngelson (approved):

PR #214, opened 2023-09-24
Condensed code in src/simulation/m_riemann_solvers.fpp

KrishnaSr71:
Condensed 2 if-else blocks ('model_eqns == 2' and 'bubbles' booleans) into 1 if-else block

### DIFF
--- a/src/simulation/m_riemann_solvers.fpp
+++ b/src/simulation/m_riemann_solvers.fpp
@@ -1417,13 +1417,8 @@ contains
                 
                 elseif(model_eqns == 2) then
 
-                    if(bubbles) then
-                        !$acc parallel loop collapse(3) gang vector default(present) private(R0_L, R0_R, V0_L, V0_R, P0_L, P0_R, pbw_L, pbw_R, vel_L, vel_R, & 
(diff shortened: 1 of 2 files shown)


src/simulation/m_riemann_solvers.fpp
```diff
@@ -1418,750 +1418,728 @@ contains
...
-                                    end do
-
-                                    !$acc loop seq
-                                    do i = 1, num_fluids
-             qL_prim_rs${XYZ}$_vf(j, k, l, E_idx + i) = qL_prim_rs${XYZ}$_vf(j, k, l, E_idx + i)/max(alpha_L_sum, sgm_eps)
-                                    end do
-
+                    elseif(model_eqns == 2) then
+                        !! DECLARE PRAGMAS
+                        if(bubbles) then
```
sbryngelson: this won't work, the `!$acc parallel loop` pragma has to go right before a loop.

sbryngelson:
The code is a bit long (not your fault), so it's a little hard to tell, but it seems like there is still a rather significant amount of twice-repeated code in the model_eqns==2 if block that could be condensed.

PR #216, opened 2023-10-01
updated example case

wilfonba:
The 2D_shockdroplet example case didn't work as given (it produced NaNs because initial densities were set to zero rather than some small epsilon), so I corrected it to match the example in the seven-equation paper and provided a README with a reference, initial, and final condition picture.


wilfonba:
Seems to be an intel runner issue

sbryngelson:
issue is explained here https://community.intel.com/t5/Intel-oneAPI-Math-Kernel-Library/APT-Repository-not-working-signatures-invalid/m-p/1170457

sbryngelson:
people were complaining about this already, yesterday:

https://community.intel.com/t5/Intel-oneAPI-Base-Toolkit/The-GPG-PUB-KEY-INTEL-SW-PRODUCTS-PUB-expired/m-p/1529173

sbryngelson (approved):

Issue #194, opened 2023-07-27
What is the proper `goldenfile` generation process? Let's standardize!


sbryngelson:
I like this part of the documentation: https://github.com/MFlowCode/MFC/blob/de90fba18f991adf7ce1e57241993e2d1846ac2d/docs/documentation/testing.md

But I think something important could be added/improved. _How do we generate goldenfiles_?! The obvious answer is 
1. Create them using `mfc.sh` commands with some compiler
2. Test with other compilers in the CI/`mfc.sh test` as new PRs come in

In a closer-to-perfect world, this is fine. In the present world, this creates some confusion. If the goldenfile and the test don't match to a "good-enough" tolerance, one has to figure out what the culprit is. Some possibilities: 
1. `gcc` and `nvhpc` are doing different things on CPUs
2. `nvhpc` is doing different things on CPUs and GPUs
3. `gcc` on CPUs is doing something different than `nvhpc` on GPUs. 

This becomes more complicated once Intel compilers are added to the CI.

Right now, I suspect that the present goldenfiles were actually created from a mixture of compilers (yikes!). This should be standardized, with all goldenfiles created using the same compiler. Right now, we are creating goldenfiles using whatever compiler the user happens to be using at the time. 

I see a resolution to this issue by doing two things:
1. Creating a standard and putting it in the documentation, e.g., we create goldenfiles using `gcc` in release mode. Period.
2. Putting a check in the `goldenfile` creation routine in `mfc.sh`/toolchain that does a check to make sure the user is _using_ the "correct" compiler when creating a new goldenfile. For example, if  `gcc` is the standard, then at the very least, it should check that the user is actually using `gcc` when creating those files. I believe the present compiler information is hashed away in a CMakeCache (or some such) file somewhere in `build/`. One could also check that it used `release` mode and not `debug` mode. 

This seems like a pretty easy-to-fix issue but is surprisingly important, given the rabbit holes we have gone down trying to pin down tolerance issues with @haochey @js-spratt etc.

PR #234, opened 2023-11-18
goldenfile metadata generation

sbryngelson:
Fixes https://github.com/MFlowCode/MFC/issues/194


sbryngelson:
@henryleberre requesting your review on this

henryleberre:
@sbryngelson I made a few edits to use CMake to obtain compiler information since using environment variables can unreliable. Let me know what you think. You can try regenerating the golden files.

sbryngelson:
@henryleberre I was thinking we would include the `pack-metadata.txt` in the repo. What do you think?

henryleberre:
@sbryngelson I agree, we should include them. I just didn't re-generate them because I thought you would like to control how the new golden files were generated.

PR #238, opened 2023-11-22
Update CMakeLists.txt to fix #237 Add fatal error to cmakelists if debug + gpu is enabled with NVHPC <= 22.11

### LINKED ISSUE #237: Add fatal error to cmakelists if debug + gpu is enabled with NVHPC <= 22.11
Debug is broken on old NVHPC versions.
Add fatal error to cmakelists if `--debug` + `--gpu` is enabled and NVHPC <= 22.11 (or so, I'm not sure what version it was fixed exactly).

Update: Turns out the CPU build with `--debug` doesn't like old versions either.

sbryngelson:
Add fatal error to cmakelists if debug + gpu is enabled with NVHPC <= 23.0


sbryngelson:
I found that I cannot build `--debug` with NVHPC + CPU on new nor old NVHPC versions. I have updated the CMakeLists to reflect the current state.

sbryngelson:
For review + squash-merge by @henryleberre

sbryngelson:
Update: I found that even after removing all of the `CMakeLists.txt` compiler flag additions, the debug CPU build with NVHPC v23.5 does not work. The `CMakeCache` shows that debug NVHPC builds automatically add flags `-g -O0 -Mbounds`. 

I assume it's `-Mbounds` that's causing the problem.

CMakeLists.txt
```diff
@@ -65,7 +65,16 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     endif()
 elseif ((CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC") OR (CMAKE_Fortran_COMPILER_ID STREQUAL "PGI"))
     if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 21.7)
-        message(FATAL_ERROR "${__err_msg}ERROR: NVHPC v21.7 or newer is required to build MFC.")
+        message(FATAL_ERROR "${__err_msg}ERROR: When using NVHPC, v21.7 or newer is required to build MFC.")
+    endif()
+    if (MAKE_BUILD_TYPE STREQUAL "Debug")
+        if (MFC_OpenACC)
+            if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 23.0)
```
henryleberre, line 72: Some older versions of NVHPC might work, I am not quite sure. I recall the debug builds starting to fail after an NVHPC update, although it might also have been related to a code change. Did you test older versions?

CMakeLists.txt
```diff
@@ -145,7 +154,7 @@ elseif ((CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC") OR (CMAKE_Fortran_COMPILER_
     endif()
     
     if (CMAKE_BUILD_TYPE STREQUAL "Debug")
-        add_compile_options(-C -g -traceback -Mchkptr -Mchkstk -Minform=inform -Mbounds)
+        add_compile_options(-C -g -O0 -traceback -Mchkptr -Mchkstk -Minform=inform -Mbounds)
```
henryleberre, line 158: Does NVHPC support `-Og` or is `-O0` required?

CMakeLists.txt
```diff
@@ -65,7 +65,16 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
     endif()
 elseif ((CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC") OR (CMAKE_Fortran_COMPILER_ID STREQUAL "PGI"))
     if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 21.7)
-        message(FATAL_ERROR "${__err_msg}ERROR: NVHPC v21.7 or newer is required to build MFC.")
+        message(FATAL_ERROR "${__err_msg}ERROR: When using NVHPC, v21.7 or newer is required to build MFC.")
+    endif()
+    if (MAKE_BUILD_TYPE STREQUAL "Debug")
+        if (MFC_OpenACC)
+            if (CMAKE_Fortran_COMPILER_VERSION VERSION_LESS 23.0)
+                message(FATAL_ERROR "${__err_msg}ERROR: When using NVHPC, v23.0 or newer is required to build MFC with Debug and GPU options.")
```
henryleberre, line 73: I am not sure if this is relevant (or important) but once a fix comes out, if you try to build an older version of NVHPC is Debug mode, you will get this fatal error and would have to remove it manually from the CMakLists.txt. Do you have any thoughts on this?

CMakeLists.txt
sbryngelson: I found most of 22.X to fail, but didn't test every minor release. I don't know about 21.X, but most computers don't have this version anyway.

CMakeLists.txt
sbryngelson: I don't believe it supports `-Og` (I did try it).

CMakeLists.txt
sbryngelson: It seems that the logic is there such that you only get this error if NVHPC is out of date and won't work anyway, no? 

Either you have some version 23.5 or so = works and no message; or, an old version 22.5 or so = doesn't work and get this message.

henryleberre: I see, I was thinking they might back-port some of these fixes but it appears not.

sbryngelson: I have no idea if they will. If they do, I guess we could go back and change the CMake file. But, even if they patch old versions, those versions are unlikely to be actually installed on clusters anyway. Either they will have the original old versions, or a newer proper major/minor version (or both), but not a patched old version. At least in my experience.

henryleberre: Taking another look at nice, I realize I meant this for the case of NVHPC in CPU + Debug mode. I don't think this is too much of a problem since we'll just remove it once it is fixed but you can add a version check if you feel so inclined.

sbryngelson: added

PR #248, opened 2023-12-11
Case Optimization Refresh

henryleberre:
- `--case-optimization` cases get their own build directories named using a hash of the case-optimization options they use. This removes the need to constantly re-build MFC when switching to/from case-optimization.
- The above, along with slight modifications to the test suite, lets us add test cases for case-optimization.


sbryngelson:
Can you provide a brief description of the change?

henryleberre:
Added. We should probably lint but I don't think a linter could have caught this error..

henryleberre:
@sbryngelson This PR should be ready for review!

sbryngelson (approved):

PR #274, opened 2023-12-24
Update Bridges2 modules 

sbryngelson:
Bridges2 also seems to require `export FC=nvfortran`, or at least loading the nvhpc modules doesn't set this env variable (it is set to gfortran by default and doesn't change). 

I wonder if there's a slick way to build that in somehow @henryleberre ?

### DIFF
--- a/mfc.sh
+++ b/mfc.sh
@@ -142,7 +142,14 @@ if [ "$1" == 'load' ]; then
         fi
     done
 
-    ok "All modules have been loaded for $M$COMPUTER$CR on $M$CG$CR"'s.'
+    if [ "$cg" == 'gpu' ]; then
+        isnv=$($FC --version | grep NVIDIA | wc -l)
+        if [ $isnv -eq 0 ]; then
+            export CC=nvc CXX=nvc++ FC=nvfortran
+        fi
+    fi
+    ok 'All modules and environment variables have been loaded.'
+
 
     return
 elif [ "$1" == "format" ]; then
--- a/toolchain/modules
+++ b/toolchain/modules
@@ -15,7 +15,7 @@ s-gpu nvhpc/22.11 cuda/nvhpc
 b     PSC Bridges2
 b-all python/3.8.6
 b-cpu allocations/1.0 gcc/10.2.0 openmpi/4.0.5-gcc10.2.0
-b-gpu openmpi/4.0.5-nvhpc22.9 nvhpc/22.9
+b-gpu openmpi/4.0.5-nvhpc22.9 nvhpc/22.9 cuda
 
 a     OLCF Ascent
 a-all python cmake/3.22.2


henryleberre:
@sbryngelson, what about:
```patch
From 14517af1db47da30adc802e6fc31ff454bd09190 Mon Sep 17 00:00:00 2001
From: Henry LE BERRE <hleberre64@gmail.com>
Date: Mon, 25 Dec 2023 19:49:04 -0500
Subject: [PATCH] mfc.sh load: Add env vars

---
 mfc.sh | 18 ++++++++++++------
 1 file changed, 12 insertions(+), 6 deletions(-)

diff --git a/mfc.sh b/mfc.sh
index 75c4a0f9..c8f95dbd 100755
--- a/mfc.sh
+++ b/mfc.sh
@@ -81,14 +81,14 @@ if [ "$1" == 'load' ]; then
     }

     COMPUTER="$(__extract "$u_c")"
-
+
     if [[ -z "$COMPUTER" ]]; then
         error "Computer $M$u_cg$CR not recognized."
         return
     fi
-
+
     MODULES=($(__extract "$u_c-$cg") $(__combine $(__extract "$u_c-all")))
-
+
     log "Loading modules for $M$COMPUTER$CR on $M$CG$CR"'s:'

     # Reset modules to default system configuration
@@ -124,7 +124,7 @@ if [ "$1" == 'load' ]; then
         if [ "$delta" -ge "0" ]; then
             printf "%0.s " $(seq 0 $delta)
         fi
-
+
         # Load the module
         module load "$module_name" > /dev/null 2>&1

@@ -142,7 +142,13 @@ if [ "$1" == 'load' ]; then
         fi
     done

-    ok "All modules have been loaded for $M$COMPUTER$CR on $M$CG$CR"'s.'
+    if [ "$cg" == 'gpu' ]; then
+        if command -v nvc > /dev/null 2>&1; then
+            export CC=nvc CXX=nvc++ FC=nvfortran
+        fi
+    fi
+
+    ok 'All modules and environment variables have been loaded.'

     return
 elif [ "$1" == "format" ]; then
@@ -183,7 +189,7 @@ elif [ "$1" == "docker" ]; then
         dockerintopts='--interactive --tty'
     fi

-    __docker_run() {
+    __docker_run() {
         docker run $dockerintopts --rm --workdir /home/me/MFC            \
                    --mount type=bind,source="$(pwd)",target=/home/me/MFC \
                    sbryngelson/mfc:latest $@
--
2.39.3

```


[0001-mfc.sh-load-Add-env-vars.patch](https://github.com/MFlowCode/MFC/files/13767652/0001-mfc.sh-load-Add-env-vars.patch)

henryleberre:
@sbryngelson, a more proper fix would be to add that info to the [toolchain/modules](toolchain/modules) file.

sbryngelson:
> @sbryngelson, a more proper fix would be to add that info to the [toolchain/modules](toolchain/modules) file. The fix I proposed isn't perfect. You can certainly use `sed` to do wonders.

@henryleberre, this fix doesn't work, unfortunately, because `nvc` etc., are still in the path, though the environment variables aren't congruent with it. Even the call to `mpif90` and `mpicc` link into NVHPC, so it seems like one needs to `grep` the environment variable or something.

```console
[I]br014: bryngel/MFC $ module list

Currently Loaded Modules:
  1) openmpi/4.0.5-nvhpc22.9   2) nvhpc/22.9   3) python/3.8.6   4) cuda/11.7.1



[I]br014: bryngel/MFC $ nvc
nvc-Warning-CUDA_HOME has been deprecated. Please, use NVHPC_CUDA_HOME instead.
nvc-Warning-No files to process

[I]br014: bryngel/MFC $ echo $CC
/opt/packages/spack/opt/spack/linux-centos8-zen/gcc-8.3.1/gcc-10.2.0-tfzxq7udz2a53dmujvasy4uz33t27iwv/bin/gcc
```

sbryngelson:
@henryleberre what about
```patch
> +    if [ "$cg" == 'gpu' ]; then
> +        isnv=$(echo $FC | grep 'nvfortran')
> +        if [[ -z "$isnv" ]]; then
> +            export CC=nvc CXX=nvc++ FC=nvfortran
> +        fi
> +    fi
```

henryleberre:
@sbryngelson Hmm, that's a good point. I would rewrite like this:
```patch
> +    if [ "$cg" == 'gpu' ]; then
> +        isnv=$($FC --version | grep NVIDIA | wc -l)
> +        if [ $isnv -eq 0 ]; then
> +            export CC=nvc CXX=nvc++ FC=nvfortran
> +        fi
> +    fi
```

since `$FC` could still point down the road to `nvfortran` without containing `nvfortran`.

henryleberre:
I had also meant that part of the `modules` file could include what environment modules to override and to what value. You would `sed` through the `modules` file as we currently do for modules to get these env variables.

sbryngelson:
> @sbryngelson Hmm, that's a good point. I would rewrite like this:
> 
> ```diff
> > +    if [ "$cg" == 'gpu' ]; then
> > +        isnv=$($FC --version | grep NVIDIA | wc -l)
> > +        if [ $isnv -eq 0 ]; then
> > +            export CC=nvc CXX=nvc++ FC=nvfortran
> > +        fi
> > +    fi
> ```
> 
> since `$FC` could still point down the road to `nvfortran` without containing `nvfortran`.

you are correct

sbryngelson:
> I had also meant that part of the `modules` file could include what environment modules to override and to what value. You would `sed` through the `modules` file as we currently do for modules to get these env variables.

Ah, I see. Yes, that's another option. One has to be careful here, too, though, because some computers have background modules either 'idle' or loaded that support some other feature that was in the default module file. So, it's possible one could find both nvidia and gnu in a `module list` stdout. Since different computers call their nvhpc module slightly different things, we would have to be careful about that (in general, for Bridges2, we already know what they do). 

I think I will use the above fix and punt this idea down the road if we need a "better" or more general fix.

Issue #174, opened 2023-06-20
New example cases: Lid-driven cavity (compressible), double Mach, and more


sbryngelson:
Simulate and add several new examples/test cases to the `examples/` directory:

I will accept PRs that add these individually (they do not have to be added all at once).

* Shu-Osher problem (1D):
	* https://arxiv.org/pdf/2110.05461.pdf

* Titarev-Toro problem (1D):
	* https://arxiv.org/pdf/2110.05461.pdf

* Riemann test problem (2D):
	* https://epubs.siam.org/doi/10.1137/0914082
	* https://arxiv.org/pdf/2110.05461.pdf

* Rayleigh-Taylor instability (2D):
	* https://arxiv.org/pdf/2110.05461.pdf (example 4.7)
	* https://www.sciencedirect.com/science/article/abs/pii/S0021999101968925?via%3Dihub

* Lid-driven compressible cavity (2D), finite Reynolds number:
	* https://doi.org/10.1016/j.cpc.2022.108527

* Double-Mach reflection (2D):
	* https://doi.org/10.1016/j.camwa.2022.12.003
	* https://arxiv.org/pdf/2110.05461.pdf (example 4.8)

* Viscous shock tube (2D):
	* https://arxiv.org/pdf/2110.05461.pdf (example 4.9)
	* https://doi.org/10.1016/j.compfluid.2008.06.008
	* https://arxiv.org/pdf/2204.00393.pdf

* Double periodic shear layer (2D):
	* https://arxiv.org/pdf/2206.14891.pdf (sec 4.2)

* Compressible decaying isotropic turbulence (2D and 3D!):
	* https://doi.org/10.1016/j.cpc.2022.108527

wilfonba:
@sbryngelson How do you want to handle the cases that will require hardcoding analytic initial conditions?

sbryngelson:
@wilfonba for cases with analytic functions, I think we should add some functionality to make this possible without commenting/uncommenting source code. I'm open to suggestions. Some example possibilities are 
* Allowing analytic functions in the Python dictionary
* Adding new functions as new patch types
* etc.

wilfonba:
What do you think of having the user supply an additional input file with fypp functions that can be included in `m_patches.f90`?

sbryngelson:
I'm not sure I understand, I would have to see a small example I think.

wilfonba:
I'll put something together

henryleberre:
[...]

PR #279, opened 2023-12-26
Velocity BCs in all coordinate directions, new example cases, nonpolytropic bc refactor

### LINKED ISSUE #186: Add velocity boundary condition for all coordinate directions
Like the inlet velocity boundary condition of @haochey and/or @anshgupta1234 , we should make it possible to set the boundary condition to have a velocity that isn't just into the domain, but also transverse to it. So setting the velocity to be a constant upwards, or `u_y = constant` for a boundary condition on the left or right boundary. This is related to issue #185 .

wilfonba:
This close #186 , addresses the lid-driven cavity problem in #174 , and reduces repeated code in the buffer population of nonpolytropic QBMM variables. All test cases pass with existing golden files except for no-slip walls. I'm not sure why these were different, but the difference was small. It also adds 2D flow between plates and 3D pipe flow example cases to validate inflow BCs in 2 and 3D.


sbryngelson:
@wilfonba Nice PR! I will look more closely soon, but it seems mostly great. It is failing on GPU, but the Phoenix GPU runner was giving many bizarre messages. I re-ran it, we will see what happens. 

If it still fails, can you do the favor of running compute-sanitizer (https://developer.nvidia.com/compute-sanitizer) and see if something comes up that's different between the master branch and this PR? I want to be sure we don't break something (and no-slip walls are an important BC, of course).

wilfonba:
It failed on GPUs again. I'll take a look.

wilfonba:
@sbryngelson Maybe try restarting it again? I ran the test suite both interactively and in batch mode using `run-phoenix-release-gpu.sh`, and all the tests passed both times.

src/post_process/m_global_parameters.fpp
```diff
@@ -276,6 +276,13 @@ contains
         bc_y%beg = dflt_int; bc_y%end = dflt_int
         bc_z%beg = dflt_int; bc_z%end = dflt_int
 
+        #:for DIM in ['x', 'y', 'z']
+            #:for DIR in [1, 2, 3]
+                bc_${DIM}$%vb${DIR}$ = 0d0
```
sbryngelson, line 284: should this not be `dflt_real`?

src/pre_process/m_global_parameters.fpp
```diff
@@ -250,6 +250,13 @@ contains
         bc_y%beg = dflt_int; bc_y%end = dflt_int
         bc_z%beg = dflt_int; bc_z%end = dflt_int
 
+        #:for DIM in ['x', 'y', 'z']
+            #:for DIR in [1, 2, 3]
+                bc_${DIM}$%vb${DIR}$ = 0d0
```
sbryngelson, line 261: dflt_real?

examples/3D_pipe_flow/case.py
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python3
+import math
```
sbryngelson: are you sure this case is working as well as it was before you introduced the BC changes? can you run the same case with the master branch and compare the result? you did have to regenerate the goldenfiles it looks like...

examples/2D_flow_between_plates/case.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python3
+
+import json
```
sbryngelson: are you sure this case is working as well as it was before you introduced the BC changes? can you run the same case with the master branch and compare the result? you did have to regenerate the goldenfiles it looks like...

docs/documentation/case.md
```diff
@@ -323,6 +326,10 @@ Details of the simulation algorithms and implementation of the WENO scheme can b
 - `bc_[x,y,z]%[beg,end]` specifies the boundary conditions at the beginning and the end of domain boundaries in each coordinate direction by a negative integer from -1 through -12.
 See table [Boundary Conditions](#boundary-conditions) for details.
 
+- `bc_[x,y,z]\%vb[1,2,3]` specifies the velocity in the (x,1), (y,2), (z,3) direction applied to `bc_[x,y,z]%beg` when using `bc_[x,y,z]%beg = -16`. Tangential velocities require viscosity, `weno_avg = T`, and `bc_[x,y,z]%beg = -16` to work properly. Normal velocities require `bc_[x,y,z]\%end = -15` or `\bc_[x,y,z]\%end = -16` to work properly.
```
sbryngelson: new sentences go on new lines. make sure no whitespace after each final line character.

sbryngelson (changes requested):

src/post_process/m_global_parameters.fpp
wilfonba: The way these BCs are implemented requires the default to be zero. See here: https://github.com/wilfonba/MFC-Wilfong/blob/4cdb21bece7cb10f3be3c785302bf668781147f6/src/simulation/m_boundary_conditions.fpp#L1051-L1072

Edit 1: I suppose it could be reworked so that these are set to dflt_real, though.
Edit 2: This change would actually double the number of if clauses in the application of boundary conditions because it would have to check the direction and then check if there's an applied velocity or not.

examples/3D_pipe_flow/case.py
wilfonba: I can approximate these cases on the master branch. Without the new inflow BCs, the comparison won't be exactly the same. I think the reason I had to regenerate the golden files was that the old implementation didn't take into account the effect assigning momentum had on the energy equation. I'll try running this with a characteristic inflow and a small non-zero velocity region near the inlet and see how it compares.

examples/2D_flow_between_plates/case.py
wilfonba: See above.

src/pre_process/m_global_parameters.fpp
wilfonba: See above

docs/documentation/case.md
wilfonba: Fixed

wilfonba:
I tried to revert to old golden files after finding a bug in one of the BCs. I'm not sure it al went correctly as I still had to regenerate one of the golden files. I'll rerun the jet-impinging on a wall test case on master and this fork to make sure things are the same, and then rerun the example cases I added with this fix.

examples/2D_lid_driven_cavity/case.py
```diff
@@ -0,0 +1,83 @@
...
+    'patch_icpp(1)%alpha(2)'       : 0.5,
+    # ==========================================================
+
+    # Fluids Physical Parameters ===============================
+    'fluid_pp(1)%gamma'            : 1./(1.4-1.),
+    'fluid_pp(1)%pi_inf'           : 0.,
+    'fluid_pp(1)%Re(1)'            : 1e4,
+    'fluid_pp(2)%gamma'            : 1./(1.4-1.),
+    'fluid_pp(2)%pi_inf'           : 0.,
+    'fluid_pp(2)%Re(1)'            : 1ex4,
```
sbryngelson: is this a bug? `1ex4`

examples/3D_pipe_flow/velocity_profile.png
sbryngelson, line 1: do we have any idea if this is right?

examples/2D_flow_between_plates/velocity_profile.png
sbryngelson, line 1: seems reasonable but do we have any way to ensure this is right? perhaps via the max velocity or boundary layer thickness?

sbryngelson (changes requested):

examples/2D_lid_driven_cavity/case.py
wilfonba: whoops... fixed

examples/3D_pipe_flow/velocity_profile.png
wilfonba: I'll calculate a 95% boundary layer thickness and add the theoretical value, simulation value, and error to the case README file.

examples/2D_flow_between_plates/velocity_profile.png
wilfonba: I'll calculate a 95% boundary layer thickness and add the theoretical value, simulation value, and error to the case README file.

sbryngelson: since this has finite Mach you'll have to be careful (perhaps use a Mach number that matches a reference case from some paper, I'm not sure if a theoretical value is known)

examples/3D_pipe_flow/velocity_profile.png
sbryngelson: since this has finite Mach you'll have to be careful (perhaps use a Mach number that matches a reference case from some paper, I'm not sure if a theoretical value is known)

wilfonba:
@sbryngelson  This should be ready to merge as long as all the tests pass.

sbryngelson:
@wilfonba run `./mfc.sh format` your failing one of the CIs.

wilfonba:
`./mfc.sh format` fails to run on my machine, but CI calls it just as I would locally. If you look at the runner, line 133 indicates that formatting was successful, but then line 1349 indicates that it exited with error code 1. I'm not sure what is happening here, but it doesn't seem to have to do with my changes to the code.

henryleberre:
@wilfonba CI runs `./mfc.sh format` and checks whether there is a diff. What failed is not the formatting, but rather the check for an empty diff. This indicates that you need to format your code with `./mfc.sh format`. What error are you getting locally?

wilfonba:
@henryleberre `UnicodeDecodeError: 'utf-8' codec can't decode byte 0x86 in position 23: invalid start byte`

wilfonba:
Here's the entire traceback.

[Uploading format.txt…]()

henryleberre:
I can't access the traceback you provided but it sounds like an issue in @sbryngelson's code. Reading each file with utf-8 encoding might fix it. Not sure why it would fail for you and not CI.

sbryngelson:
@henryleberre it appears to fail in both fprettify and the indenter. @wilfonba you can run `file file.fpp` on the flagrant file to see if it isn't UTF-8. Regardless -- can you run format from, say, Phoenix or some other machine?

sbryngelson:
@wilfonba I was able to clone your fork/branch and run `format` without issue from my laptop. So, I just did that. If this comes up again in a seemingly independent context I will look closer into it.

wilfonba:
@sbryngelson The file it fails at is `src/post_process/m_checker.f90', and running file on it says is ASCII text. Formatting worked fine on Phoenix.

sbryngelson (approved):

PR #281, opened 2023-12-29
Improve count with intent to add to CI

sbryngelson:
Improve `./mfc.sh count` with intent to add to CI. Removes comment lines from count but includes openacc directives.


sbryngelson:
Really want this to print the line count different between master and current

toolchain/mfc/count.py
```diff
@@ -12,7 +12,13 @@ def handle_dir(dirpath: str) -> typing.Tuple[typing.List[typing.Tuple[str, int]]
 
     for filepath in glob.glob(os.path.join(dirpath, '*.*f*')):
         with open(filepath) as f:
-            count = sum(1 if not l.isspace() else 0 for l in f.read().split('\n'))
+            count = 0
+            for l in f.read().split('\n'):
```
henryleberre, line 15: It might be easier to read if this was rewritten as:
```python
for l in f.read().split('\n'):
    # Skip withespace
    if l.isspace() or len(l) == 0:
        continue
    # Skip comments but not !$acc ones!
    if l.lstrip().startswith("!") and not l.lstrip().startswith("!$acc"):
        continue
    count += 1
```

henryleberre:
> Really want this to print the line count different between master and current.
> 
> Example https://github.com/tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886

Do you have a use-case in mind?

sbryngelson:
> > Really want this to print the line count different between master and current.
> > Example [tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886](https://github.com/tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886)
> 
> Do you have a use-case in mind?

I would like to know how many lines are being added or subtracted via a PR. I see some PRs where it's not immediately clear that a refactor actually reduces line count. The GitHub PR linecount diff includes all sorts of things that we don't care about.

henryleberre:
> > > Really want this to print the line count different between master and current.
> > > Example [tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886](https://github.com/tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886)
> > 
> > 
> > Do you have a use-case in mind?
> 
> I would like to know how many lines are being added or subtracted via a PR. I see some PRs where it's not immediately clear that a refactor actually reduces line count. The GitHub PR linecount diff includes all sorts of things that we don't care about.

I see. For the specific case where you make a code-only refactor, you can subtract the + and - lines as shown on the PR's Github page since when you only modify a line, it counts as both a deletion and an addition. I just realized that things like adding (not changing) golden files do skew results, although they don't have many lines.

sbryngelson:
> > > > Really want this to print the line count different between master and current.
> > > > Example [tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886](https://github.com/tinygrad/tinygrad/actions/runs/7306212421/workflow?pr=2886)
> > > 
> > > 
> > > Do you have a use-case in mind?
> > 
> > 
> > I would like to know how many lines are being added or subtracted via a PR. I see some PRs where it's not immediately clear that a refactor actually reduces line count. The GitHub PR linecount diff includes all sorts of things that we don't care about.
> 
> I see. For the specific case where you make a code-only refactor, you can subtract the + and - lines as shown on the PR's Github page since when you only modify a line, it counts as both a deletion and an addition. I just realized that things like adding (not changing) golden files do skew results, although they don't have many lines.

Yes this _mostly_ does the job. Though I figure the CI mechanisms used for this would also be useful for other presumably more important things.

toolchain/mfc/count.py
sbryngelson: Agree

sbryngelson: resolved via [f5032c7](https://github.com/MFlowCode/MFC/pull/281/commits/f5032c75cde50fa2aa99850c4e163618aa798ae1)

.github/workflows/count-2.yml
```diff
@@ -0,0 +1,77 @@
...
+        with:
+          python-version: '3.10'
+
+      - name: Count Line Diff
+        run: |
+          EOF=$(dd if=/dev/urandom bs=15 count=1 status=none | base64)
+          echo "test_result<<$EOF" >> "$GITHUB_ENV"
+          echo "this is only a test" >> "GITHUB_ENV"
+          echo "$EOF" >> "$GITHUB_ENV"
+      - uses: marocchino/sticky-pull-request-comment@v2
```
henryleberre: @sbryngelson I am not sure if this will work for PRs made from forks because of the (lack of) privileges granted in the GITHUB_TOKEN secret. This was an issue I encountered while working on the benchmarking PR.

sbryngelson: i'll have a look thanks

sbryngelson:
@henryleberre I think this is the best this PR is going to get without more serious effort. The commit-bot is a no-go as you noticed. I was thinking that one could add an argument to `./mfc.sh count` that allows a diff between two versions of `MFC/src/*` but this seems like more work/tricks than I can muster at the moment (I don't quite understand how it's going through and making those nice line count tables to begin with).

PR #282, opened 2023-12-29
modernize the f77 eigenvalue module, prettify

sbryngelson:
I confirmed that this still produces the instability wave expected. Prepping to "prettify" the whole code then use CI to make sure it stays pretty.

### DIFF
--- a/mfc.sh
+++ b/mfc.sh
@@ -159,12 +159,10 @@ elif [ "$1" == "format" ]; then
         pip3 install --upgrade fprettify
     fi
 
+    opts="--recursive --silent --indent 4 --c-relations --enable-replacements --enable-decl --whitespace-comma 0 --whitespace-multdiv 0 --whitespace-plusminus 1 --case 1 1 1 1 --strict-indent --line-length 1000"
     if [ "$1" == "diff" ]; then 
         shift
-        out=$(fprettify ${@:-src} --exclude "src/*/autogen" --recursive --silent \
-            --indent 4 --c-relations --enable-replacements --enable-decl \
-            --whitespace-comma 0 --whitespace-multdiv 0 --whitespace-plusminus 1 \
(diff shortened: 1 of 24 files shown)


mfc.sh
```diff
@@ -159,10 +159,26 @@ elif [ "$1" == "format" ]; then
         pip3 install --upgrade fprettify
     fi
 
+    if [ "$1" == "diff" ]; then 
+        shift
+        out=$(fprettify ${@:-src} --exclude "src/*/autogen" --recursive --silent \
```
henryleberre: @sbryngelson Can you put the options to `fprettify` in a variable so we don't have to maintain two copies? Perhaps something like this:
```bash
opts="--opt1 value1 --opt2 value2 ...other args..."
fprettify $opts -s -d ...other args...
```

Notice that in `fprettify $opts`, `$opts` will be expanded since it isn't quoted.

mfc.sh
```diff
@@ -159,10 +159,26 @@ elif [ "$1" == "format" ]; then
...
+            --indent 4 --c-relations --enable-replacements --enable-decl \
+            --whitespace-comma 0 --whitespace-multdiv 0 --whitespace-plusminus 1 \
+            --case 1 1 1 1 --strict-indent --line-length 1000 -s -d)
+        if [ -z "${out}" ]; then
```
henryleberre, line 166: I usually prefer counting the lines but this just as correct as far as I can tell.

sbryngelson: Yeah, I checked it.

mfc.sh
sbryngelson: Indeed. Will do this.

sbryngelson: resolved by [a0c48e4](https://github.com/MFlowCode/MFC/pull/282/commits/a0c48e4854ae7648e964fcd66ccb32affefec9a5)

PR #283, opened 2023-12-31
Update run-phoenix-release-gpu.sh

sbryngelson:


henryleberre:
@sbryngelson I expect this change to at least double the time taken by Phoenix CI. If I am mistaken, something else is in need of attention

sbryngelson:
@henryleberre good to know -- I needed to make the change just to get the runner stable. It was trying to issue 24 queues (`$(nproc)`) to 2 GPUs and causing slurmderrors. We can issue 2 queues to 2 GPUs, though some of the tests require 2 ranks, but it should still be OK.

PR #285, opened 2024-01-01
Operator splitting, adaptive time stepping, and other fixes for mixing layer and bubbles

lee-hyeoksu:
## Description

This PR introduces several new features including
1. Strang operator splitting with adaptive time stepping strategy (`adap_dt = T`),
2. Solving the number density equation directly and computing void fraction from the number density (for `adv_n = T`),
3. Fixing the instability wave code to apply slip-wall BC correctly,
4. Support for artificial Mach number (or artificial pi_inf) - use artificial pi_inf for background flow and true pi_inf for bubbles. The ratio between artificial and true pi_inf's, `pi_fac`, is added as an input parameter.

### Type of change

- [x] Bug fix (non-breaking change which fixes an issue) - option 3
- [x] New feature (non-breaking change which adds functionality) - option 1, 2, 4

## How Has This Been Tested?

- [x] Passed test suite with CPUs on Richardson and Delta
- [x] Passed test suite with GPUs on Delta
- [x] Passed all CI tests on github
- [x] Tested on a single collapsing bubble case - added to examples (0d_bubblecollapse_adap)
- [x] Tested on a 2D mixing layer case - added to examples (2d_mixing_artificial_Ma)

**Test Configuration**:

- Test suite with CPUs on Richardson: /opt/gcc/9.3.0/bin/, /act/openmpi-2.0/gcc-9/bin/, python-3.10.13 (installed on my local dir)
- Test suite with CPUs on Delta: gcc/11.4.0, openmpi, python/3.11.6 (modules loaded)
- Test suite with GPUs on Delta: nvhpc/22.11, openmpi+cude/4.1.5+cuda, cmake, python/3.11.6 (modules loaded)

## Checklist:

- [x] I have added comments for new code
- [x] I added Doxygen docstrings to new code
- [x] I have added example cases in `examples/` that demonstrate my new feature performing as expected


sbryngelson:
@lee-hyeoksu it looks like you regenerated all the goldenfiles. Why?

lee-hyeoksu:
@sbryngelson With the current golden files, the CI tests failed like right now. I think it is due to my changes, although I am not sure which parts are responsible for that. That's why I regenerated the golden files in the previous commit, which was reverted now. Do you think the CI test should pass with the current golden files? Or if you think it's fine, I will update the golden files again.

henryleberre:
@lee-hyeoksu Do you expect a meaningful difference in output between current master and your branch for these test cases that fail? If so, you should regenerate the golden files for these cases specifically and validate the new results.

lee-hyeoksu:
@henryleberre I don't think there is an important update on output. Most changes are about input parameter and simulation process, not output file/data. I will check the code again and look into the test case to see what happens there.

sbryngelson:
@lee-hyeoksu it is failing all of the sub-grid bubbles cases. I think we may need to have a discussion about having two problem length scales (a bubble one and another one). I am confident this is what is causing the issue, and even if you get them to match, it seems confusing. 

I recommend bringing it up with Tim. Again, my preference is one length scale for the entire problem setup. Otherwise, this just feels like it's asking for confusion/problems.

lee-hyeoksu:
I tried to remove the second length and velocity scales (bubble scales). On this PR, I was able to remove the second velocity scale, but I needed at least one length scale parameter, which is `len_ratio` representing the ratio of two lenth scales. This is because the current MFC defines `R0` that have values centered at 1 (for monodisperse, `R0 = 1`. But, here, `R0 = 1` means `R0` is normalized with the initial equilibrium radius, which introduces the use of length scale other than the background flow length scale. This is why I had to introduce `len_ratio`. 

The use of `len_ratio` kind of parameter can be avoided by using the bubble scales (not background flow scales) for the entire simulation. In this case, however, I need to  somehow rescale the output data to the background flow scales to analyze them. 

I will discuss about this with Tim. If you have any suggestions/thoughts, it would be greatly appreciated!

sbryngelson:
> @sbryngelson I tried to remove the second length and velocity scales (bubble scales). On this PR, I was able to remove the second velocity scale, but I needed at least one length scale parameter, which is `len_ratio` representing the ratio of two lenth scales. This is because the current MFC defines `R0` that have values centered at 1 (for monodisperse, `R0 = 1`. But, here, `R0 = 1` means `R0` is normalized with the initial equilibrium radius, which introduces the use of length scale other than the background flow length scale. This is why I had to introduce `len_ratio`.
> 
> The use of `len_ratio` kind of parameter can be avoided by using the bubble scales (not background flow scales) for the entire simulation. In this case, however, I need to somehow rescale the output data to the background flow scales to analyze them.
> 
> I will discuss about this with Tim. If you have any suggestions/thoughts, it would be greatly appreciated!

I'm not sure I understand. The length scale of the broader flow physics can be much larger than the bubble length scale without causing issues in the code or simulations. This just means that the relevant broader flow length scale is larger than unity (in code units), which is fine.

sbryngelson:
You can resolve the merge conflicts with the formatting discussion here: https://github.com/MFlowCode/MFC/pull/296

lee-hyeoksu:
> You can resolve the merge conflicts with the formatting discussion here: #296

I didn't recognize about this so I resolved conflicts in a much more complicated way😅... I don't know it works or not, but if it doesn't work, I may need to revert those commits and try the way introduced in #296.

lee-hyeoksu:
I removed scaling parameters that I introduced previously, so the MFC has single normalization throughout the code.

For now, tests on Phoenix have not been done, but all the others passed. So I think this draft PR is ready for review!

sbryngelson:
@lee-hyeoksu sorry for the delay. Please add regular MFC full documentation, code comments, and doxygen docstrings for everything you added. Thanks!

lee-hyeoksu:
@sbryngelson Is there any procedure to test PR on GPUs with nvhpc/nvfortran other than just creating/updating PR? I haven't done simulations on GPUs or Phoenix, so I have no clue why CI tests on GPU/Pheonix failed, especially if it is due to compile time. Do you have any idea on what's causing the failure?

sbryngelson:
@lee-hyeoksu you should have access to computers with GPU devices on them, like NCSA Delta or PSC Bridges2. From there we have standard procedures for building MFC. If your code does not run efficiently on GPU devices then we cannot merge it. The CI automatically checks if your code builds and runs on multiple platforms, including different GPUs.

lee-hyeoksu:
@sbryngelson I see. I can check it on Bridges2. Thanks!

lee-hyeoksu:
@sbryngelson I have updated the documentation, code comments, and doxygen docstrings. Please let me know if you want more details or if anything is written in wrong format.

src/simulation/m_time_steppers.fpp
```diff
@@ -723,6 +841,26 @@ contains
             call s_pressure_relaxation_procedure(q_cons_ts(1)%vf)
         end if
 
+        if (adv_n .and. alter_alpha) then
+            !$acc parallel loop collapse(3) gang vector default(present)
```
wilfonba: This (and the other copies of it) should probably be abstracted into a subroutine to keep the time-stepping subroutine cleaner

sbryngelson: agreed

sbryngelson:
@lee-hyeoksu I updated your first PR comment at the top, can you go through the relevant steps?

lee-hyeoksu:
> @lee-hyeoksu I updated your first PR comment at the top, can you go through the relevant steps?

Sure! Sorry for the delay.

sbryngelson:
> @sbryngelson I have a question about a checklist item regarding `--nsys` since I have almost zero knowledge on Nsight Systems profile. I got informed a little bit from the MFC documentation though.
> 
> To my understanding, I can run any simulation with `--nsys` flag, and it will provide a summary of overall performance. But, the PR checklist above says "... have attached the output file (.nsys-rep) and plain text results to this PR". For this checklist, how many cases or what cases should I run with `--nsys` and include the results in this PR?

Hey @lee-hyeoksu -- good question, I should add more details to this template. You should only run cases that show the new feature. So, for example, a 2D case using the operator/timestep splitting would be one example and a case using the artificial Mach number. I think that covers all of your new features that could impact the simulation time/efficiency. 

Yes if you run with `--nsys` you will get an nsight systems profile output (assuming you are running in GPU mode in a GPU node). You can append options that get passed to the nsight systems command line call if you like. They're documented here:  https://docs.nvidia.com/nsight-systems/UserGuide/index.html

lee-hyeoksu:
Hi @sbryngelson I am working on generating and testing new test cases. So I generated golden files for new test cases on Richardson and tried to test on Delta with GPUs, but one of them fails due to tolerance check. Also, the CI test on Phoenix GPU fails at the same test case.

```
Error: Test tests/8F5D5EAD: 1D -> Bubbles -> Monopole -> adap_dt=T: Variable n°306 (1-indexed) in D/prim.3.00.000050.dat is not within tolerance:
  - Candidate:   1.00010853032233
  - Golden:      1.00010853042249
  - Error:       abs: 1.00E-10, rel: 1.00E-10
  - Tolerance:   abs: 1.00E-10, rel: 1.00E-10
```

Meanwhile, I found that when I re-generate golden file on Delta GPU, the test passes. Also it passes on Delta CPU test. But it now fails on Richardson (and CI test on CPUs). Do you have any idea of what causes the inconsistency?

lee-hyeoksu:
@sbryngelson I managed to generate `.nsys-rep` files, and checked them on Nsight Systems GUI. At this point, I have a few questions.

1. Regarding "Ran a Nsight Systems profile using ./mfc.sh run XXXX --gpu -t simulation --nsys, and have attached the output file (.nsys-rep) and plain text results to this PR",
- I have two files generated: `report1.nsys-rep` and `report2.nsys-rep`. Is it expected to have two files? If so, can I just upload all of them to the `example/0d_bubblecollapse_adap`, for example?
- What does the "plain text results" refer to? Also, should I upload this one to the same location with `.nsys-rep` files?

2. Regarding "Ran my code using various numbers of different GPUs (1, 2, and 8, for example) in parallel and made sure that the results scale similarly to what happens if you run without the new code/feature",
- Here, do you expect me to compare the normalized wall time, like shown in the [MFC documentation](https://mflowcode.github.io/documentation/md_expectedPerformance.html)?

lee-hyeoksu:
As I mentioned in Slack channel, attached are the NSYS reports and plain text output for three example cases. For all cases, the reports were generated with the options `-N 1 -n 4` on a gpuA100x4 node on Delta. For each case, there are three report files generated, but the third one is larger than GitHub file size limit, so I just attached first two reports. 

[0D_bubblecollapse_adap.zip](https://github.com/MFlowCode/MFC/files/14552386/0D_bubblecollapse_adap.zip)
[2D_mixing_arfitifical_Ma.zip](https://github.com/MFlowCode/MFC/files/14552392/2D_mixing_arfitifical_Ma.zip)
[3D_weak_scaling.zip](https://github.com/MFlowCode/MFC/files/14552395/3D_weak_scaling.zip)

lee-hyeoksu:
Also, the table shows the `3D_weak_scaling` results for different numbers of nodes with all of the GPUs on each nodes. The final time and wall time do increase with the ranks, but there is no significant difference between current master branch (without change) and my branch

Screenshot 2024-03-10 at 5 35 32 PM

lee-hyeoksu:
> > As I mentioned in Slack channel, attached are the NSYS reports and plain text output for three example cases. For all cases, the reports were generated with the options `-N 1 -n 4` on a gpuA100x4 node on Delta. For each case, there are three report files generated, but the third one is larger than GitHub file size limit, so I just attached first two reports. Please let me know if you'd like to see the reports for other cases or different ranks.
> > [0D_bubblecollapse_adap.zip](https://github.com/MFlowCode/MFC/files/14552386/0D_bubblecollapse_adap.zip) [2D_mixing_arfitifical_Ma.zip](https://github.com/MFlowCode/MFC/files/14552392/2D_mixing_arfitifical_Ma.zip) [3D_weak_scaling.zip](https://github.com/MFlowCode/MFC/files/14552395/3D_weak_scaling.zip)
> 
> Thanks but the third report is actually the only important one (the other two are for the `syscheck` and `preprocess`, which are irrelevant here).

I see. But the report file sizes are too large to be uploaded on Github (25MB limit). Could you access to the box link below?

https://caltech.box.com/s/zb7b6h4ph1k8pfue0631r1fi7sb9s4ls

sbryngelson:
You can send it to me via Slack or Email or some other mechanism. I cannot access that Box link.

sbryngelson:
@lee-hyeoksu, thanks for keeping this up to date. I will merge after @anandrdbz finishes the Frontier PR #368, a blocking change.

lee-hyeoksu:
@sbryngelson I merged with the PR #368 and checked it works correctly on Richardson CPUs and Delta GPUs.

sbryngelson:
Note: I found this https://github.com/MFlowCode/MFC/issues/390

.github/workflows/phoenix/test.sh
```diff
@@ -16,5 +16,5 @@ if [ "$job_device" == "gpu" ]; then
     n_test_threads=`expr $gpu_count \* 2`
 fi
 
-./mfc.sh test -a -j $n_test_threads $device_opts -- -c phoenix
+./mfc.sh test -a -r -j $n_test_threads $device_opts -- -c phoenix
```
sbryngelson: remove this line please

docs/documentation/case.md
```diff
@@ -396,13 +399,17 @@ $$ \alpha_N=1-\sum^{N-1}_{i=1} \alpha_i $$
...
+- `adv_n` activates the direct computation of number density by the Riemann solver instead of computing number density from the void fraction in the method of classes.
+
 - `mpp_lim` activates correction of solutions to avoid a negative void fraction of each component in each grid cell, such that $\alpha_i>\varepsilon$ is satisfied at each time step.
 
 - `mixture_err` activates correction of solutions to avoid imaginary speed of sound at each grid cell.
 
 - `time_stepper` specifies the order of the Runge-Kutta (RK) time integration scheme that is used for temporal integration in simulation, from the 1st to 5th order by corresponding integer. 
 Note that `time_stepper = 3` specifies the total variation diminishing (TVD), third order RK scheme ([Gottlieb and Shu, 1998](references.md#Gottlieb98)).
 
+- `adap_dt` activates the Strang operator splitting scheme which splits flux and source terms in time marching, and an adaptive time stepping strategy is implemented for the source term. It requires `bubbles = TRUE`, `polytropic = TRUE`, `adv_n = TRUE` and `time_stepper = 3`.
```
sbryngelson: replace TRUE -> 'T'

docs/documentation/case.md
```diff
@@ -636,8 +643,7 @@ The parameters are optionally used to define initial velocity profiles and pertu
 
 - `vel_profile` activates setting the mean streamwise velocity to hyperbolic tangent profile. This option works only for 2D and 3D cases.
 
-- `instability_wave` activates the perturbation of initial velocity by instability waves obtained from linear stability analysis for a mixing layer with hyperbolic tangent mean streamwise velocity profile.
-This option only works for 2D and 3D cases, together with `vel_profile = 'T'`.
+- `instability_wave` activates the perturbation of initial velocity by instability waves obtained from linear stability analysis for a mixing layer with hyperbolic tangent mean streamwise velocity profile. This option only works for `n > 0`, `bc_y%[beg,end] = -5`, and `vel_profile = TRUE`.
```
sbryngelson: TRUE -> 'T'

examples/3D_weak_scaling/case.py
```diff
@@ -186,8 +186,8 @@
     'cyl_coord'                    : 'F',
     'dt'                           : dt,
     't_step_start'                 : 0,
-    't_step_stop'                  : int(5000*16.0/ARGS["gbpp"]),
-    't_step_save'                  : int(1000*16.0/ARGS["gbpp"]),
+    't_step_stop'                  : int(50*16.0/ARGS["gbpp"]),
```
sbryngelson: please put these back to how they were before

src/simulation/m_data_output.fpp
```diff
@@ -1161,15 +1161,17 @@ contains
...
-                        nbub = DSQRT((4.d0*pi/3.d0)*nR3/alf)
+                        if (adv_n) then
+                            nbub = q_cons_vf(n_idx)%sf(j - 2, k, l)
+                        else
+                            nR3 = 0d0
+                            do s = 1, nb
+                                nR3 = nR3 + weight(s)*(nR(s)**3d0)
+                            end do
 
+                            nbub = DSQRT((4.d0*pi/3.d0)*nR3/alf)
```
sbryngelson: DSQRT -> dsqrt

src/simulation/m_bubbles.fpp
```diff
@@ -298,35 +322,136 @@ contains
...
+
+                            ! Evaluate f(x0+h0,y0+h0*f(x0,y0))
+                            myR_tmp(2) = myR_tmp(1) + h0*myV_tmp(1)
+                            myV_tmp(2) = myV_tmp(1) + h0*myA_tmp(1)
+                            myA_tmp(2) = f_rddot(myRho, myP, myR_tmp(2), myV_tmp(2), R0(q), &
+                                                 pb, pbdot, alf, n_tait, B_tait, &
+                                                 bub_adv_src(j, k, l), divu%sf(j, k, l))
+
+                            ! Compute d2 = ||f(x0+h0,y0+h0*f(x0,y0))-f(x0,y0)||/h0
+                            d2 = DSQRT(((myV_tmp(2) - myV_tmp(1))**2d0 + (myA_tmp(2) - myA_tmp(1))**2d0)/2d0)/h0
```
sbryngelson: DSQRT -> dsqrt

toolchain/templates/include/helpers.mako
```diff
@@ -86,7 +86,7 @@ unset CRAY_ACC_MODULE
 cd "${MFC_ROOTDIR}"
 
 cat >>"${output_summary}" <<EOL
-${target.name}: $(expr $t_${target.name}_stop - $t_${target.name}_start)
+${target.name}: $(expr (($t_${target.name}_stop - $t_${target.name}_start)*1.0/1000000000) )
```
sbryngelson: this does not work

anandrdbz: I think I have fixed it in the latest commit

sbryngelson:
@anandrdbz something is wrong with your fix using `bc` because now all times are 0. From the gpu log artifact for the PR for any of the cases...

```yaml
- -n
- '2'
lock:
  debug: false
  gpu: true
  mpi: true
syscheck: 0
pre_process: 0
simulation: 0
post_process: 0
```

anandrdbz:
@sbryngelson I've reverted the changes, the reason benchmark was failing was due to the fact that pre_process takes just about 1 second so it gets rounded down to zero sometimes. A temporary fix I can add (so that it never fails) is to set this value to 1 if we ever encounter 0

sbryngelson:
@anandrdbz we can just do the proper fix using actual floats instead of integers. But that will be a separate PR.

sbryngelson:
This PR is about ready to merge. It is failing Frontier tests because all of Frontier is being used by a reservation the last few days. 

@anandrdbz can you run this PR with your non-polytropic QBMM and regular QBMM cases to make sure it "works" before we merge? there are some pretty substantial changes in the bubble code and I want to be sure it didn't break your features.

anandrdbz:
@sbryngelson I ran a few sample tests and everything seems to work (also the test suite has those features anyways

anandrdbz:
And the QBMM module wasn't altered in this PR

sbryngelson (approved):

Issue #286, opened 2024-01-01
`./mfc.sh` not enforcing minimum Python version - what is the minimum version anyhow?


sbryngelson:
Reprising some old issues - This is the current line in `./mfc.sh`:

https://github.com/MFlowCode/MFC/blob/1899a6c7a8edb69b21403f0f0c8d312e9de47dbf/mfc.sh#L6C1-L7

On Bridges2, we run Python 3.8.6, which works -- both in terms of _working_ (build, run, test all work fine seemingly) and the bash script not stopping the job. 

So, I'm not sure what part of Python 3.9 is required, and regardless, the script isn't doing its job somehow.

PR #289, opened 2024-01-04
#269: Fix Python version check

### LINKED ISSUE #269: Python min version still a problem
`./mfc.sh` tests the command `python3` for the Python version. This checks the "usual" Python version, such as whatever `which python3` or `type python3` shows (e.g., if one has used an alias or export to change their user space via, say, `alias python3='/usr/bin/python3.11'`, it will show that). However, the shebangs in the toolchain invoke `/usr/bin/env python3`, which ignores the userspace declarations and will give you the PATH default `python3`. A workaround for this is to create symlinks and change your path, but this is cumbersome. 

So, this issue notes two things:
* If one has python < 3.9 in their `/usr/bin/env python3`, `./mfc.sh` will not catch this and thus
* Even if one uses exports/aliases to change the user space, MFC will still fail at `./mfc.sh test`

We need a slick way to detect this. It would also be handy if MFC ran with Python aliases.

henryleberre:
This PR closes #269 and #286, including:
- Sets the minimum Python version to 3.8. It appears to be working on PSC Bridges2.
- Adds a more thorough version check that includes checking the Python version within the generated venv, every time it is invoked. So far, we have only been checking compatibility when creating the venv.

### DIFF
--- a/mfc.sh
+++ b/mfc.sh
@@ -4,7 +4,8 @@ unset u_c
 unset u_cg
 
 MFC_PYTHON_MIN_MAJOR=3
-MFC_PYTHON_MIN_MINOR=9
+MFC_PYTHON_MIN_MINOR=8
+MFC_PYTHON_MIN_STR="$MFC_PYTHON_MIN_MAJOR.$MFC_PYTHON_MIN_MINOR"
 
 # Check whether this script was called from MFC's root directory.
 if [ ! -f "$(pwd)/toolchain/mfc.py" ]; then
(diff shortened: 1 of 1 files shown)


henryleberre:
@sbryngelson can you confirm this fixes your issue on the H100 Rogues Gallery system? It works for me. It also looks like 3.11 is the default in the `PATH` now. Tested with:

```console
$ ./mfc.sh test -j 2 -b mpirun --gpu --no-mpi
```

It also worked with MPI and `-b mpirun` on CPUs.

sbryngelson:
> @sbryngelson can you confirm this fixes your issue on the H100 Rogues Gallery system? It works for me. It also looks like 3.11 is the default in the `PATH` now. Tested with:
> 
> ```
> $ ./mfc.sh test -j 2 -b mpirun --gpu --no-mpi
> ```
> 
> It also worked with MPI and `-b mpirun` on CPUs.

Thanks @henryleberre -- That system (violet1) now has default 3.11 everywhere, but I still can't get it to work because it's missing cuFFT somehow.

sbryngelson (approved):

henryleberre:
@sbryngelson I reloaded the `nvhpc` module and that fixed the cuFFT error but introduced an MPI one.. I'm sure we can get it working. Will there be more H100s in the future?

sbryngelson:
> @sbryngelson I reloaded the `nvhpc` module and that fixed the cuFFT error but introduced an MPI one.. I'm sure we can get it working. Will there be more H100s in the future?

Yes, but they will probably be in Phoenix, not Violet1. We will have GraceHopper in the RG first.

PR #293, opened 2024-01-05
CI: Ensure tests/ is committed (#293)

henryleberre:
Closes #293.


sbryngelson:
Good idea @henryleberre though looks like you'll need to check stdout rather than an exit code.

henryleberre:
@sbryngelson correct me if am wrong, but `git diff --exit-code` should return with a non-zero exit  code if there is a diff.

This still works if someone adds a test because, by "adding a test", we mean that they added code in cases.py that generate cases. This leaves open the possibility that they did this the other way around. That they manually added a test in tests/ but not in cases.py. I can easily add a check for this. We just delete all case.py files in tests/ before we run `mfc.sh test`.

sbryngelson:
> @sbryngelson correct me if am wrong, but `git diff --exit-code` should return with a non-zero exit code if there is a diff.

Right, seems fine, but as you can see from the Phoenix runners not every git version supports this flag (or some such other issue).
 
> This still works if someone adds a test because, by "adding a test", we mean that they added code in cases.py that generate cases. This leaves open the possibility that they did this the other way around. That they manually added a test in tests/ but not in cases.py. I can easily add a check for this. We just delete all case.py files in tests/ before we run `mfc.sh test`.

I know what you mean, I think, but it isn't what I meant, I don't think. If someone creates a new test they will ultimately be committing a case.py file, a readme, and a golden file for that test -- causing a diff. If they update the goldenfiles they will likewise be causing a diff. Right?

henryleberre:
@sbryngelson I see what you meant. We check if there is a diff relative to `HEAD` (their last commit), not upstream's `master`. Therefore their own changes don't appear in the diff after you clone their branch.

Thanks for pointing out that compatibility issue with git, I will fix it shortly.

sbryngelson:
> @sbryngelson I see what you meant. We check if there is a diff relative to `HEAD` (their last commit), not upstream's `master`. Therefore their own changes don't appear in a diff right after you clone their branch.
> 
> Thanks for pointing out that compatibility issue with git, I will fix it shortly.

🤌

henryleberre:
@sbryngelson Ready for review. I fixed the Phoenix runner (changed the command to work with outdated git).

sbryngelson (approved):

PR #296, opened 2024-01-06
IBM

anshgupta1234:
This pull request adds the IBM feature to MFC. It supports:
- 2D and 3D IBM
- A variety of IBM geometries
- Support for multiphase flow and bubbles


henryleberre:
@anshgupta1234 (Warning: use at your own risk) To fix conflicts due to formatting, I would suggest:
```console
$ git branch ibm-backup origin/ibm
$ git remote add upstream git@github.com:henryleberre/MFC.git
$ git fetch upstream
$ git checkout upstream/master -- mfc.sh toolchain/indenter.py toolchain/bootstrap/format.sh
$ ./mfc.sh format
$ git stash -- mfc.sh toolchain
$ git add src
$ git commit -m temp
$ git rebase -i origin/master
... squash commits here ...
$ git rebase upstream/master
... hope there are only a few conflicts ...
$ git push -f
```

sbryngelson:
@anshgupta1234 you'll need to grab the latest toolchain formatting files to re-format your code (this is why there are so many merge errors). @henryleberre's instructions for this are above.

henryleberre:
@sbryngelson I believe your #314 PR does not account for when files are added or removed in PRs (that is, when the file list of both branches does not match).

sbryngelson:
> @sbryngelson I believe your #314 PR does not account for when files are added or removed in PRs (that is, when the file list of both branches does not match).

You're correct. I tested this when I was making the PR, but I guess I didn't consider some edge case. I will look into this, but it need not hold up this PR.

sbryngelson:
Is this PR "ready for review," or @anshgupta1234, do you need to sort out some other things first?

anshgupta1234:
I believe it should be good to go!

sbryngelson:
Pasting some discussion I had on Slack with @anshgupta1234 here for "permanence":

> 512x256x256 3D IBM vs No IBM comparison test. Viscosity on, ran for 100 timesteps on GPUs w/ 4 ranks
Without IBM: 1.125714475909869 sec/timestep
With IBM:       1.144588934878508 sec/timestep
= 1.67666% increase

> It’s GPU w/ 4 ranks 

Regarding "correctness," @anshgupta1234  has verified that the theoretical bow shock angles are correct at a specific _Re_ and _Ma_ for flow past a cylinder.

```
Re = 6.5 x 10^5
Density = 0.2199 kg / m^3
Velocity: 527.2 m/s
Sphere Diameter (used as characteristic length): 10 cm = 0.1 m
```

Example overlay of MFC and ref paper simulations (of S. Sarkar UCSD).
image

sbryngelson (approved):

sbryngelson:
@wilfonba and @henryleberre can you review this PR for obvious opportunities for refactoring? at present it adds many lines that seem redundant. obviously any bugs you spot should be flagged as well.

examples/2D_ibm_airfoil/case.py
```diff
@@ -0,0 +1,143 @@
...
+    # 'patch_icpp(2)%y_centroid'     : 1.0E-03,
+    # 'patch_icpp(2)%radius'         : 0.2E-03,
+    # # Specify the patch primitive variables 
+    # 'patch_icpp(2)%vel(1)'         : 1.0E+00,
+    # 'patch_icpp(2)%vel(2)'         : 1.0E+00,
+    # 'patch_icpp(2)%pres'           : 1.E+05,
+    # 'patch_icpp(2)%alpha_rho(1)'   : 1.22E+00,
+    # 'patch_icpp(2)%alpha(1)'       : 1.E+00,
+    # 'patch_icpp(2)%alter_patch(1)'  : 'T',
+    # # ========================================================================
```
wilfonba: Are these lines necessary?

examples/2D_ibm_airfoil/case.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # Patch: Cylinder Immersed Boundary ========================================
```
wilfonba, line 120: Is this a cylinder or an airfoil?

wilfonba:
I made a few comments above. One way to remove a few hundred lines is to generalize the buffer communication subroutines so that only one is needed rather than two. I know this works with my surface tension work, but I'm not sure how well it would work with IBM communication. Also note that the total additions are closer to 5400 once you ignore the eigen solver module. I didn't see any obvious other ways to par down the number of additions significantly.

sbryngelson:
> I made a few comments above. One way to remove a few hundred lines is to generalize the buffer communication subroutines so that only one is needed rather than two. I know this works with my surface tension work, but I'm not sure how well it would work with IBM communication. Also note that the total additions are closer to 5400 once you ignore the eigen solver module. I didn't see any obvious other ways to par down the number of additions significantly.

Got it. @anshgupta1234 can you have a look at the above? @wilfonba can you link your abstraction for MPI? 

Also, the eigensolver module was associated with another problem which I just resolved via a commit. That was a good catch, I had to rewrite that module before @anshgupta1234's fork in order to remove goto statements for formatting purposes.

wilfonba:
I haven't actually implemented a more general communication routine myself yet. I still have two separate but very similar subroutines. The data structures and types are almost identical in my case, so I can see how it would be done pretty easily. I just noticed that the IMB code is communicating integer values rather than floats. this would make a more general routine less straightforward.

sbryngelson:
@anshgupta1234 delete all of your `case.py` and `README` files in the `tests/` directory. This is an update we made and they are no longer included in version control.

sbryngelson:
> I haven't actually implemented a more general communication routine myself yet. I still have two separate but very similar subroutines. The data structures and types are almost identical in my case, so I can see how it would be done pretty easily. I just noticed that the IMB code is communicating integer values rather than floats. this would make a more general routine less straightforward.

Ok - understood. Let's keep this refactor for another time then, and do all communication refactoring at the same time (or at least most of it).

@anshgupta1234 can you confirm that both slip and no-slip work (for inviscid and viscous cases, respectively)? I think you showed me it did, but I cannot fully remember and at some point in time there was a bug. I guess the main one I'm thinking of is 3D+no-slip.

anshgupta1234:
Hi everyone! I'll address the comments in the PR shortly. As for the no-slip case, although it does work qualitatively and quantitatively for 2D IBM, it's a little weirder for 3D IBM. @sbryngelson you may remember from our conversation that the no-slip case returned a realistic result with the correct bow shock but the flow behind the sphere did not match the paper I was referencing.

I discussed the case file with @anandrdbz, fixed the Reynolds number, and everything else pretty much matched the paper. However, the flow was still different.

image
image

sbryngelson:
@anshgupta1234 I'm aware of this case but suspect it to be due to differences in inflow (turbulent versus steady) or grid resolution, given that all the other features match so well. What about the slip case in 2D/3D?

anshgupta1234:
> @anshgupta1234 I'm aware of this case but suspect it to be due to differences in inflow (turbulent versus steady) or grid resolution, given that all the other features match so well. What about the slip case in 2D/3D?

@sbryngelson Yes, those do match up!

image
image

sbryngelson:
Looks good. Thanks @anshgupta1234. I think some minor cleanup is needed like what @wilfonba mentioned, and then we can merge.

sbryngelson:
@anshgupta1234 sorry to drag this out. 

1. You have a couple of new tests golden files that don't have `goldenfile_metadata`. That metadata file should have been autogenerated when creating the goldenfile. Can you recreate/commit those by just regenerating the goldenfile?
2. Does IBM work with the 6-equation model, `model_eqns = 3`? This would be a quite useful feature if true. If it doesn't work, can you look at what is needed to add support? The velocities are all the same with the 6-equation model. This will be quite useful for our multiphase efforts. If it doesn't work but seems like a lot of effort to add, let me know.

anshgupta1234:
@sbryngelson No worries, that's my bad! I'll push  it right now. And as for the 6-equation model, yes, that does work. @anandrdbz famously said,  "it only took like 4 lines of code" (or something like that)  to do it 😂

anshgupta1234:
@sbryngelson I'm a bit unfamiliar with the 6-equation model, so correct me if I'm wrong; if we simply change `model_eqns` to 3 in `examples/2D_ibm_multiphase`, that would do the trick, right? As far as just having a 6-equation model test case.

sbryngelson:
> @sbryngelson I'm a bit unfamiliar with the 6-equation model, so correct me if I'm wrong; if we simply change `model_eqns` to 3 in `examples/2D_ibm_multiphase`, that would do the trick, right? As far as just having a 6-equation model test case.

Umm I think so. See this case for a reference: https://github.com/MFlowCode/MFC/blob/master/examples/3D_sphbubcollapse/case.py

sbryngelson:
@henryleberre this is "ready" to merge pending your review. I see the test suite GPU run failed, but I think you fixed this in one of our recent PRs. Please confirm/deny that.

src/simulation/m_rhs.fpp
```diff
@@ -1850,9 +1852,24 @@ contains
...
             call nvtxEndRange
         end do
+
+        if (ib) then
+            do l = 0, p
+                do k = 0, n
+                    do j = 0, m
+                        if (ib_markers%sf(j, k, l) /= 0) then
+                            do i = 1, sys_size
+                                rhs_vf(i)%sf(j, k, l) = 0d0
```
henryleberre, line 1865: Is it OK if this is happening on the host? I don't see any `copyin` or `copyout` near here.

henryleberre (commented):
@anshgupta1234 Could you rebase on `master` for the Benchmarking to run?

src/simulation/m_rhs.fpp
sbryngelson: I agree. the `ib` variables may all be on the device, but without a `!$acc parallel loop` this is not happening on the device.

sbryngelson:
The latest autobenchmark @anshgupta1234 @henryleberre  suggests that your new code is a factor of 1/5-times slower than the current master branch for a 3D shock droplet problem: 
```
  Case              Pre Process   Simulation   Post Process  
 ─────────────────────────────────────────────────────────── 
  3D_shockdroplet         1.00x        0.80x          1.00x  
```
This could well be a limitation of the test itself. Perhaps it needs to be re-run or ran for longer times.

anshgupta1234:
@sbryngelson Ah, wait--do I need to make a separate PR to `MFlowCode/MFC` and then rebase this  branch?

sbryngelson:
> @sbryngelson Ah, wait--do I need to make a separate PR to `MFlowCode/MFC` and then rebase this branch?

Potentially. I don't 100% understand it. @henryleberre would be more confident about the answer.

sbryngelson:
New GPU bench result:

```console
mfc: OK > (venv) Entered the Python 3.9.12 virtual environment (>= 3.8).

      .=++*:          -+*+=.        | sbryngelson3@login-phoenix-slurm-1.pace.gatech.edu [Linux]
     :+   -*-        ==   =* .      | ----------------------------------------------------------
   :*+      ==      ++    .+-       | --jobs 1
  :*##-.....:*+   .#%+++=--+=:::.   | --mpi
  -=-++-======#=--**+++==+*++=::-:. | --no-gpu
 .:++=----------====+*= ==..:%..... | --no-debug
  .:-=++++===--==+=-+=   +.  :=     | --targets pre_process, simulation, and post_process

 Comparing Bencharks: master/bench-gpu.yaml is x times slower than pr/bench-gpu.yaml.
                                                             
  Case              Pre Process   Simulation   Post Process  
 ─────────────────────────────────────────────────────────── 
  3D_shockdroplet         1.00x        0.21x          1.00x  
                                                             

```

anshgupta1234:
This is crazy! I'll look into it.

anshgupta1234:
Actually @sbryngelson , there is an alternative theory I just realized. I updated the benchmark to go from taking 100 to 500 time steps. So if it's running the simulation for 500 time steps on this PR but only for 100 on master, then that could be the issue!!

sbryngelson:
@anshgupta1234 yes this is surely it. can you post a screenshot of the nsight sys profile from the GUI of an IBM simulation without the openacc statements in that loop?

anshgupta1234:
image

Latest benchmark btw

anshgupta1234:
image

Interesting, I did the profiling on the 2D ibm case and the IB section took so little time even without the ACC loop.

sbryngelson:
@anshgupta1234 Can you post the profile file so I can look at it? What I really want to see is if the IBM loop ends up as a kernel on the device or not.

sbryngelson:
Looks good. It turns out that the IBM part of the profiled simulations was so short that it wasn't visible on the benchmarks, so removing the ACC loop and performing it on the host didn't result in a meaningful slowdown.

sbryngelson (approved):

Issue #288, opened 2024-01-02
MFC seems to rebuild all targets on `run` even if they're already been built via `build`


sbryngelson:
MFC rebuilds all targets on `./mfc.sh run` even if they've already been built via `./mfc.sh build`. Seems to be triggered via a `GLOB MISMATCH` though I haven't investigated why this occurs. Even occurs if one _immediately_ runs after a build. One option is to avoid `build` altogether though this is obviously not a very attractive idea.

It even partially rebuilds some dependencies like FFTW/HDF5.

Example snippet:
```console
[I]br011: bryngel-startup-proj/MFC $ ./mfc.sh run examples/1D_sodshocktube/case.py -t pre_process simulation -e batch -N 1 -n 8 -w "00:10:00" -p GPU -# "shb-test" -b mpirun -a phy210041p
mfc: OK > (venv) Entered the Python virtual environment.

      .=++*:          -+*+=.          bryngel@br011.ib.bridges2.psc.edu [Linux]
     :+   -*-        ==   =* .        -----------------------------------------
   :*+      ==      ++    .+-         --jobs 1
  :*##-.....:*+   .#%+++=--+=:::.     --mpi
  -=-++-======#=--**+++==+*++=::-:.   --gpu
 .:++=----------====+*= ==..:%.....   --no-debug
  .:-=++++===--==+=-+=   +.  :=       --targets pre_process and simulation
  +#=::::::::=%=. -+:    =+   *:
-----------------------------------------------------------
 .*=-=*=..    :=+*+:      -...--      $ ./mfc.sh [build, run, test, clean, count, packer]
--help

Run
  Acquiring /ocean/projects/phy210041p/bryngel/MFC/examples/1D_sodshocktube/case.py...
  Configuration:
    Input
/ocean/projects/phy210041p/bryngel/MFC/examples/1D_sodshocktube/case.py
    Job Name      (-#)  shb-test
    Engine        (-e)  batch
    Nodes         (-N)  1
    Tasks (/node) (-n)  8
    Walltime      (-w)  00:10:00
    Partition     (-p)  GPU
    Account       (-a)  phy210041p
    Email         (-@)

  Generating input files for pre_process...

    Generating pre_process.inp.
      INFO: Forwarded 32/50 parameters.

  Generating input files for simulation...

    Generating simulation.inp.
      INFO: Forwarded 32/50 parameters.

  Generating syscheck/include/case.fpp.
    INFO: Custom case.fpp file is up to date.

[...]

Issue #287, opened 2024-01-01
Specifying mpi executable/binary doesn't carry over into a batch script?


sbryngelson:
Running the following: `[I]br012: bryngel-startup-proj/MFC $ ./mfc.sh run examples/1D_sodshocktube/case.py -t pre_process simulation -e batch -N 1 -n 1 -w "00:10:00" -p RM-small -# "shb-test" -b mpirun -a phy210041p` 

on Bridges2 gives in stdout:

```console
mfc: OK > (venv) Entered the Python virtual environment.

      .=++*:          -+*+=.          bryngel@br012.ib.bridges2.psc.edu [Linux]
     :+   -*-        ==   =* .        -----------------------------------------
   :*+      ==      ++    .+-         --jobs 1
  :*##-.....:*+   .#%+++=--+=:::.     --mpi
  -=-++-======#=--**+++==+*++=::-:.   --no-gpu
 .:++=----------====+*= ==..:%.....   --no-debug
  .:-=++++===--==+=-+=   +.  :=       --targets pre_process and simulation
  +#=::::::::=%=. -+:    =+   *:
-----------------------------------------------------------
 .*=-=*=..    :=+*+:      -...--      $ ./mfc.sh [build, run, test, clean, count, packer]
--help

Run
  Acquiring /ocean/projects/phy210041p/bryngel/MFC/examples/1D_sodshocktube/case.py...
  Configuration:
    Input
/ocean/projects/phy210041p/bryngel/MFC/examples/1D_sodshocktube/case.py
    Job Name      (-#)  shb-test
    Engine        (-e)  batch
    Nodes         (-N)  1
    Tasks (/node) (-n)  1
    Walltime      (-w)  00:10:00
    Partition     (-p)  RM-small
    Account       (-a)  phy210041p
    Email         (-@)

  Generating input files for pre_process...

    Generating pre_process.inp.
      INFO: Forwarded 32/50 parameters.

  Generating input files for simulation...

    Generating simulation.inp.
      INFO: Forwarded 32/50 parameters.

  Generating syscheck/include/case.fpp.
    INFO: Custom case.fpp file is up to date.

  $ cmake --build
/ocean/projects/phy210041p/bryngel/MFC/build/no-debug_no-gpu_mpi/syscheck --target
syscheck -j 1 --config Release

[100%] Built target syscheck

  $ cmake --install
/ocean/projects/phy210041p/bryngel/MFC/build/no-debug_no-gpu_mpi/syscheck

[...]

Issue #240, opened 2023-11-26
[RFC] Gather and maintain a list of "batch file" templates for various computers


henryleberre:
Currently, we use "template" batch submission scripts to generate and submit job scripts when running `./mfc.sh run <...> -e batch`. This adds a lot of convenience and modularity but some tuning is still required to get some jobs (especially GPU ones) running on different systems. Since many of us use the same systems, it would be nice if these modifications were tracked in source control for added visibility and to make running batch jobs easier for new users. There are two options I can think of:

- We maintain, perhaps in the `misc/` folder, the templates we use for different systems.
- We add everything to current template files in the form of comments one can just remove and edit.
- We build toolchain support for various systems and run configurations. However, this does not sound neither easy, nor friendly, nor maintainable.

What are you thoughts?

sbryngelson:
I agree that something useful can and should be done here. 2. and 3. do sound hard to maintain and more complicated than necessary. 1. seems best. 

With 1., it would be useful to keep templates there, but also build in a way to automatically grab those templates via `mfc.sh` calls. Most new folks don't even know the templates exist or how they work, or where they end up. So, something like `mfc.sh run <...> -e batch -c bridges2` (`c` for computer, I suppose) that automatically grabs the appropriate `bridges2` template would be quite useful (and maintainable).

sbryngelson:
The documentation for Bridges2 with > 8 GPUs per job contradicts what works on their system. So, I'll post a working script here for a future template (> 1 GPU node, in this case 2 nodes of 8 GPUs):

```bash
#!/usr/bin/env bash
#SBATCH --job-name="shb-test"
#SBATCH --nodes=2
#SBATCH --ntasks-per-node=8
#SBATCH --cpus-per-task=1
#SBATCH --gpu-bind=verbose,closest
#SBATCH --gres=gpu:v100-16:8
#SBATCH --time=00:10:00
#SBATCH --partition="GPU"
#SBATCH --output="shb-test.out"
#SBATCH --account="phy210041p"
#SBATCH --error="shb-test.err"
#SBATCH --export=ALL
#SBATCH --mail-type="BEGIN, END, FAIL"

TABLE_FORMAT_LINE="| - %-14s %-35s - %-14s %-35s |\n"
TABLE_HEADER="+-----------------------------------------------------------------------------------------------------------+ \n"
TABLE_FOOTER="+-----------------------------------------------------------------------------------------------------------+ \n"
[...]

PR #307, opened 2024-01-14
Mako Batch/Interactive Templates & Build Fixes

### LINKED ISSUE #292: CI doesn't ensure tests/ is committed
If your run `./mfc.sh test` on `master`, you will observe a `git diff` in case files for some tests. Per @sbryngelson's request for having test case files kept in source control, we should ensure through CI that these files are committed.

The diff currently includes some Phase Change tests (@JRChreim). A few other cases also have diffs but they still describe the same cases - physically.

henryleberre:
The last PR before the Benchmarking one.

Closes #292, #240, and #287.


docs/documentation/running.md
```diff
@@ -2,7 +2,31 @@
...
+> Running simulations locally should work out of the box. On supported clusters,
+> you can append `-c <computer name>` on the command line to instruct the MFC toolchain
+> to make use of the template file `toolchain/templates/<computer name>.mako`. You can
+> browse that directory and contribute your own files. Since systems and their schedulers
+> do not have a standardized syntax to request certain resources, MFC can only provide
+> support for a restricted subset of common or user-contributed configuration options.
+>
+> Adding a new template file or modifying an existing one will most likely be required if:
+> - You are on a cluster that does not have a template yet.
+> - Your cluster is configured with SLURM and but fails when interactive jobs are
```
sbryngelson: typo 'and but'

misc/run-phoenix-release-cpu.sh
```diff
@@ -12,5 +12,4 @@ cd "$SLURM_SUBMIT_DIR"
 echo "Running in $(pwd):"
 
 . ./mfc.sh load -c p -m gpu
-./mfc.sh test -j $(nproc) -b mpirun -a
-
+./mfc.sh test -j $(nproc) -a -- -b mpirun
```
sbryngelson: can you comment on what `--` does?

toolchain/mfc/args.py
```diff
@@ -98,29 +94,31 @@ def add_common_arguments(p, mask = None):
...
-    run.add_argument(      "--dry-run",        action="store_true",                       default=False,      help="(Batch) Run without submitting batch file.")
-    run.add_argument("--case-optimization",    action="store_true",                       default=False,      help="(GPU Optimization) Compile MFC targets with some case parameters hard-coded.")
-    run.add_argument(      "--no-build",       action="store_true",                       default=False,      help="(Testing) Do not rebuild MFC.")
-    run.add_argument("--wait",                 action="store_true",                       default=False,      help="(Batch) Wait for the job to finish.")
+    run.add_argument("input",                      metavar="INPUT",              type=str,                     help="Input file to run.")
+    run.add_argument("arguments",                  metavar="ARGUMENTS", nargs="*", type=str, default=[],       help="Additional positional arguments to pass to the case file.")
+    run.add_argument("-e", "--engine",             choices=["interactive", "batch"],              type=str, default="interactive", help="Job execution/submission engine choice.")
```
sbryngelson, line 103: is there an option to use `-b mpirun` for template files or does one have to modify the template

toolchain/mfc/run/run.py
```diff
@@ -24,30 +31,67 @@ def validate_job_options() -> None:
...
+                       "--cache-control=none", "--clock-control=none"] + ARG("ncu")
 
-    input_file = input.load()
+    if ARG("nsys") is not None:
+        if not does_command_exist("nsys"):
+            raise MFCException("Failed to locate [bold green]NVIDIA Nsight Systems[/bold green] (nsys).")
 
-    engine = engines.get_engine(ARG("engine"))
-    engine.init(input_file)
+        return ["nsys", "profile", "--stats=true", "--trace=mpi,nvtx,openacc"] + ARG("nsys")
```
sbryngelson, line 46: i think i had a comment on your original PR that said that nsys profile needs to go in the opposite order against `mpirun` for whatever we do currently (for mpi jobs at least). does this fix that?

toolchain/templates/default.mako
```diff
@@ -0,0 +1,53 @@
...
+warn "Please use a different template via the $MAGENTA--computer$COLOR_RESET option."
+echo
+
+% for binpath in binpaths:
+    echo -e ":) Running $MAGENTA${binpath}$COLOR_RESET:\n"
+
+    % if not mpi:
+        ${' '.join([f"'{x}'" for x in profiler ])} "${binpath}"
+    % else:
+        if command -v jsrun > /dev/null; then
```
sbryngelson: can we not just pass `-b` in here? i know we should be using the templates but this seems like an easy improvement of sorts.

sbryngelson (changes requested):

sbryngelson:
I think this PR was going to fix the issue of not using multiple ranks with `./mfc.sh test -j X` on the Phoenix CPU runner. does it?

docs/documentation/running.md
henryleberre: Fixed! I am getting tired..

toolchain/mfc/args.py
henryleberre: I agree, I just added it back in.

toolchain/templates/default.mako
henryleberre: As per above, I agree and I just added it.

henryleberre:
> I think this PR was going to fix the issue of not using multiple ranks with `./mfc.sh test -j X` on the Phoenix CPU runner. does it?

I had forgotten to include it. Just did. I added:
```console
               --bind-to none                             \
```

So when using GT Phoenix, you just need to use `-c phoenix` while testing and/or running (interactive & batch) MFC.

misc/run-phoenix-release-cpu.sh
henryleberre: It forwards arguments to `./mfc.sh run` (called by `./mfc.sh test`). This means we no longer have to redocument all of these options as options of `./mfc.sh test`, too. We also no longer have to manually forward each option.

henryleberre: I just changed this line to:
```console
./mfc.sh test -j $(nproc) -a -- -c phoenix
```

toolchain/mfc/run/run.py
henryleberre: Noted. I just switched the order.

sbryngelson:
so the new command to test on an `mpirun` system is `./mfc.sh test -j X -- --binary mpirun --gpu`? this seems to be what I need when testing your PR on Phoenix.

henryleberre:
> so the new command to test on an `mpirun` system is `./mfc.sh test -j X -- --binary mpirun --gpu`? this seems to be what I need when testing your PR on Phoenix.
> 
> it is also re-building even though the GLOBs are the same?

For Phoenix, I would recommend just:
```console
./mfc.sh test -j X --gpu -- -c phoenix
```

I looked into #288 and it's not quite fixed yet. It is really quite strange. I think it has to do with CMake expecting its binary directory to be empty upon creation skipping GLOB checks within that directory.

sbryngelson:
> > so the new command to test on an `mpirun` system is `./mfc.sh test -j X -- --binary mpirun --gpu`? this seems to be what I need when testing your PR on Phoenix.
> > it is also re-building even though the GLOBs are the same?
> 
> For Phoenix, I would recommend just:
> 
> ```
> ./mfc.sh test -j X --gpu -- -c phoenix
> ```
> 
> But this also works now:
> 
> ```
> ./mfc.sh test -j X --gpu -- -b mpirun -f --bind-to none
> ```
> 
> I looked into #288 and it's not quite fixed yet. It is really quite strange. I think it has to do with CMake expecting its binary directory to be empty upon creation and thus skipping GLOB checks within that directory.

got it

henryleberre:
Part of the appeal of this PR (other than there being a template per system) is that:
- Templates are now used for both Batch & Interactive execution. This gives us more flexibility and makes it a lot easier to understand and modify the behavior of these template files. You no longer need to parse through `toolchain/mfc/run/engines.py`. It doesn't exist anymore!
- Instead of my handcrafted and regex-powered template engine, I switched to an all-around better solution, Mako.
- A lot of indirection was removed (through the support of `--` in argument parsing in many places).
- Refactored `system()` in `toolchain/mfc/` so that whenever we execute a command, we no longer depend on the shell. We provide a list of arguments rather a command-string. It's a lot more difficult to create a bug by introducing spaces in a command or variable name.

sbryngelson (approved):

sbryngelson:
This looks great - glad to merge once tests pass. We can also now run CPU tests much faster.

henryleberre:
@sbryngelson I just fixed the rebuild issue! Closes #288.

sbryngelson:
> @sbryngelson I just fixed the rebuild issue! Closes #288.

Thank goodness! It was very strange. What was it?

sbryngelson:
@henryleberre you have access to NCSA Delta and Purdue Anvil as well. Can you add appropriate batch templates for those computers?

henryleberre:
> > @sbryngelson I just fixed the rebuild issue! Closes #288.
> 
> Thank goodness! It was very strange. What was it?

There were two issues:
- We only generated the case-optimization FPP files while building, not also while configuring. This causes a GLOB mismatch.
- When I added their generation to the configure step, I had forgotten we also delete the binary directory before running CMake (we have good reasons for doing this). So the files were actually never there while CMake was configuring the MFC targets.

henryleberre:
> @henryleberre you have access to NCSA Delta and Purdue Anvil as well. Can you add appropriate batch templates for those computers?

Could these be added later? It might be a good idea for users to get acquainted with this way of doing things, adding Mako files as they need them.

sbryngelson:
> > @henryleberre you have access to NCSA Delta and Purdue Anvil as well. Can you add appropriate batch templates for those computers?
> 
> Could these be added later? It might be a good idea for users to get acquainted with this way of doing things, adding Mako files as they need them.

Yes that's fine. Merge when appropriate.

Issue #225, opened 2023-11-08
Test suite does not touch any `--case-optimization` cases


sbryngelson:
The current test suite does not do any case optimization cases. Is it possible to compile a single test case that uses all the case optimization parameters (bubbles, 3D, etc.) and add it as a one-off test? would this double the run-time of our test suite?

henryleberre:
This should be rather straightforward to add. It would currently require two partial rebuilds of `simulation` (once for a `--case-optimization` case and once for all others) as well as the serialization of the `--case-optimization` test case(s) - you would run into issues with running multiple of these cases concurrently. The cost is probably negligeable, however. There are was we could get around this but it would most likely be over-engineering.

sbryngelson:
`--case-optimization` could be added to the test suite matrix for one select case that has "all the case optimization features" and for a couple select runners (at least one CPU and one GPU). This seems like the most straightforward solution.

sbryngelson:
This can be closed because benchmarking runs case optimization cases, so we will know if it breaks (at least for the most part).

Issue #49, opened 2023-01-06
Use GitHub CI to continuously benchmark some example cases


sbryngelson:
Implement continuous benchmarking via CI of a set of MFC examples.

A good example: https://github.com/benchmark-action/github-action-benchmark

henryleberre:
@sbryngelson I found a workaround to make this work! It's now under `/bench` on our website. Here's it up on my fork: https://henryleberre.github.io/MFC/bench/. Which cases should we include in the benchmark? The ones on the website are just there as a proof of concept.

To make the publishing a lot easier, I'm using a new GitHub feature whereby you can publish to GitHub pages from a workflow without having the need for another branch or another repository. This removes the need for the `DOC_PUSH_URL` secret.

sbryngelson:
For this specific issue, I would consider a selection of 2D and 3D cases. Probably 2-component (`num_fluids = 2`) problems will be sufficiently general and the problem size `n,m,p` will mostly depend on the resources/runner. I think we can make do with just a few cases, which we can discuss.

So, where does it (or can it) run the benchmark? We probably want a CPU and GPU benchmark for each "test" case.

Also, consider making smaller pull requests so we can iterate more quickly.

henryleberre:
I agree. However, in this case, it would have been difficult to make _neat_ Pull Requests considering the scope of my changes. Adding support for Cray/CCE required significant and/or major changes to the Fortran code, the toolchain code, the workflows (GitHub & ORNL), the build system, and the documentation. They all tie together and adding Cray support is a single feature so there weren't many "discrete" steps in the process at which times the code would have been in an appropriate state for it to be upstreamed. I'll make sure none of my changes broke anything before submitting a Pull Request and finish this feature once that is done!

sbryngelson:
reference: https://github.com/illinois-ceesd/timing

sbryngelson:
Update: Now have self-hosted runners to do this properly without relying on Gitlab.

sbryngelson:
Adding @belericant to assignees. 

The first step is to benchmark a couple of cases in the `examples/` directory on the "self-hosted" runner (let's start in CPU mode) and append those numbers to a file that can be turned into a graph. This will append the graph with every CI call. Some care will be needed in selecting what examples to use and so on, but that will be easy, and let's worry about it later. Once a prototype is working, then getting the cases going will be simple.

[...]

PR #316, opened 2024-01-21
Handshake: Benchmarking

henryleberre:


sbryngelson:
I noticed that the outputs of the regular test suite Phoenix runners are sent to artifacts. This makes it hard to quickly find the problem (though it is probably more elegant in some other ways). Can we revert back to something that shows the stdout/err to the developer in the actions call stack? My old way of `cat test.out` was a bit hacky but worked, not sure if there is a nicer way.

henryleberre:
@sbryngelson I added that feature back and removed `3D_sphbubcollapse` from the list of benchmarked cases. This should be a (not-so minimal) MVP for the Benchmarking feature. It can be expanded upon in the future.

sbryngelson:
@henryleberre Phoenix is undergoing maintenance for the next few days so I suppose this will sit for this week.

sbryngelson:
closes https://github.com/MFlowCode/MFC/issues/225

sbryngelson:
closes https://github.com/MFlowCode/MFC/issues/49

sbryngelson (approved):

sbryngelson:
closes https://github.com/MFlowCode/MFC/pull/316#pullrequestreview-1839029903

sbryngelson:
did it just fail because of this?  `cat: .github/workflows/phoenix/bench.sh: No such file or directory`

PR #321, opened 2024-01-26
One benchmark case (short)

sbryngelson:


sbryngelson:
Update from discussion with @henryleberre : 

Some of these benchmarks will never finish because the CPU version will take forever (too many time steps). We should create a benchmark directory for the benchmark cases and give each such case the argument option for whether they are running on GPU or CPU.

sbryngelson:
This fails for seemingly multiple reasons. One is that one slurm job was stuck in the queue forever. The other is that too many files are generated (or something like this), which saturates my project disk space.

sbryngelson:
The fix (as far as I can tell) is separate benchmark cases, in a separate benchmark directory, with fewer time steps and file output dumps.

PR #322, opened 2024-01-26
Update required cmake version

sbryngelson:
We need cmake >= 3.25 for the new nvhpc compilers (nvhpc 23.11 is one such culprit).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -6,7 +6,7 @@
 #  than you might expect to find or need. Note that any variable or symbol
 #  prefixed by CMAKE_ is reserved by CMake and has special meaning.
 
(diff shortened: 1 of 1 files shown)


henryleberre:
@sbryngelson I am not sure we should do this. It is my understanding that updating `cmake_minimum_required` should only be done when we use a new CMake (language) feature. e.g: you would get a syntax error if running a prior version. CMake v3.20 should be able to build MFC in any configuration, provided the system is _nice enough_. Also, v3.25 is somewhat recent and not all systems support it.

sbryngelson:
Good to know. I'm not sure if I know of other systems with nvhpc 23.11. Once I find one I'll see if it doesn't work there as well (as Violet/RG systems).

PR #333, opened 2024-02-08
Minor addition

haochey:
Add threshold and support_width to input

### DIFF
--- a/src/pre_process/m_mpi_proxy.fpp
+++ b/src/pre_process/m_mpi_proxy.fpp
@@ -91,6 +91,7 @@ contains
                 call MPI_BCAST(patch_icpp(i)%${VAR}$, size(patch_icpp(i)%${VAR}$), MPI_DOUBLE_PRECISION, 0, MPI_COMM_WORLD, ierr)
             #:endfor
 
+            call MPI_BCAST(patch_icpp(i)%model%threshold, 1, MPI_DOUBLE_PRECISION, 0, MPI_COMM_WORLD, ierr)
             call MPI_BCAST(patch_icpp(i)%model%spc, 1, MPI_INTEGER, 0, MPI_COMM_WORLD, ierr)
 
             ! Broadcast IB variables
--- a/src/simulation/m_global_parameters.fpp
+++ b/src/simulation/m_global_parameters.fpp
@@ -491,6 +491,7 @@ contains
             mono(j)%support = 1
             mono(j)%foc_length = dflt_real
             mono(j)%aperture = dflt_real
+            mono(j)%support_width = 2.5d0
         end do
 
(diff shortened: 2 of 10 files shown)


src/pre_process/m_mpi_proxy.fpp
```diff
@@ -91,6 +91,7 @@ contains
                 call MPI_BCAST(patch_icpp(i)%${VAR}$, size(patch_icpp(i)%${VAR}$), MPI_DOUBLE_PRECISION, 0, MPI_COMM_WORLD, ierr)
             #:endfor
 
+            call MPI_BCAST(patch_icpp(i)%model%threshold, 1, MPI_DOUBLE_PRECISION, 0, MPI_COMM_WORLD, ierr)
```
wilfonba, line 94: I believe this could be added to the fypp loop above

src/simulation/m_global_parameters.fpp
```diff
@@ -491,6 +491,7 @@ contains
             mono(j)%support = 1
             mono(j)%foc_length = dflt_real
             mono(j)%aperture = dflt_real
+            mono(j)%support_width = 2.5d0
```
wilfonba, line 494: Where does 2.5 come from?

haochey: It was the hard-coded default for monopole.

src/pre_process/m_mpi_proxy.fpp
haochey: This new input is not an array unlike others in the loop above. I can add it to the loop in line 80 for normal patch inputs if this makes more sense.

src/simulation/m_global_parameters.fpp
sbryngelson: please add a comment for this line. also, why is this a new line? shouldn't the support width have been already present in the code?

src/pre_process/m_mpi_proxy.fpp
sbryngelson: yes

src/simulation/m_global_parameters.fpp
haochey: It's not in the code yet. The new input is the number of characteristic cells the support spans. It uses the new input to calculate the 'sig' which is the actual width of the support.

sbryngelson: Sounds good to me.

sbryngelson:
Was this closed intentionally?

haochey:
> Was this closed intentionally?

Sorry for any confusions. I closed it because I want to use a forced push to solve a conflict in the root branch. I made a new PR #336.

PR #334, opened 2024-02-09
Toolchain fix for older python version

wilfonba:
This fixes the problem that I had building on wingtip for python 3.8.10 thanks to @henryleberre


toolchain/mfc/bench.py
```diff
@@ -1,6 +1,6 @@
 import os, sys, uuid, subprocess, dataclasses
 
-import rich.table
+import rich.table, typing
```
henryleberre: Nitpick but `typing` should be imported on the first line with the other Python standard library dependencies.

henryleberre (changes requested):

toolchain/mfc/bench.py
wilfonba: Done

wilfonba:
No idea why this change messed with the benchmark time

sbryngelson:
@henryleberre can merge this whenever it is done

henryleberre:
Thanks @wilfonba. The increase in benchmark time is due to:
- Only running a few time steps
- Storing benchmark numbers as integers (seconds).

I believe @sbryngelson will add more cases.

sbryngelson:
@wilfonba @henryleberre yeah it's in my queue

PR #335, opened 2024-02-10
small fix to m_mpi_proxy.fpp

JRChreim:
Adding 'cv' and 'qvp' variables to m_mpi_proxy.fpp module such that parallel pre-processing runs correctly whenever those variables are called (e.g. phase change module).


sbryngelson:
run `./mfc.sh format`

PR #338, opened 2024-02-14
Creates NCSA Delta modules and mako templates for batch jobs

sbryngelson:
## Description

Creates NCSA Delta modules and mako templates for batch jobs

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [ ] Checked compilation on CPU and GPU
- [ ] Checked that tests pass
- [ ] Checked that batch jobs submit as expected

**Test Configuration**:

	* NCSA Delta

## Checklist

- [ ] I have added example cases in `examples/` that demonstrate my new feature performing as expected


henryleberre:
@sbryngelson, does this template also work for interactive jobs? i.e. are you able to run the test suite on an allocated node with `./mfc.sh test -- -c delta` ?

henryleberre:
> > @sbryngelson, does this template also work for interactive jobs? i.e. are you able to run the test suite on an allocated node with `./mfc.sh test -- -c delta` ?
> 
> It works if you first `salloc` yourself resources, which I assume is the intended use case?

Yes! I was just checking since `srun` does not always work interactively.

toolchain/modules
```diff
@@ -46,3 +46,8 @@ c-all cmake/3.23.2 cray-fftw/3.3.10.2 hdf5/1.12.1 cray-python/3.9.13.1
 c-all ninja/1.10.2 cray-mpich/8.1.23
 c-cpu
 c-gpu rocm/5.1.0 craype-accel-amd-gfx90a
+
+d     NCSA Delta
+d-all python/3.11.6
```
henryleberre: nitpick: you could add `openmpi` to `d-all`.

toolchain/templates/delta.mako
```diff
@@ -0,0 +1,54 @@
...
+echo
+
+% for target in targets:
+    ${helpers.run_prologue(target)}
+
+    % if not mpi:
+        ${' '.join([f"'{x}'" for x in profiler ])} "${target.get_install_binpath()}"
+    % else:
+        ${' '.join([f"'{x}'" for x in profiler ])}             \
+            srun \
```
henryleberre: For interactive mode, you should maybe explicitly set `-np ${nodes*tasks_per_node}`. This should not interfere with Batch jobs.

henryleberre (changes requested):

toolchain/templates/delta.mako
sbryngelson: How does one make this only happen in the interactive case?

toolchain/modules
sbryngelson: agreed. i didn't appreciate that `-all` modules are loaded _after_ the cpu/gpu specific ones, which changes the calculus on if this is acceptable.

toolchain/templates/delta.mako
henryleberre: In all batch files, we currently just add `-np ${nodes*tasks_per_node}`. This should match exactly the number of tasks requested in a `-e batch` job so _should be fine to specify_.

sbryngelson: Also, it's not really clear to me how interactive mode even works. How does this script know how many nodes and tasks it has?

henryleberre: If you needed to conditionally specify `-np` (inline) you could so some python magic like:
```
srun \
  ${f"-np {nodes*tasks_per_node}" if engine == 'interactive' else ''}
```

henryleberre: For interactive mode, it still uses the `-n` and `-N` command-line arguments. The default is $(n,N)=(1,1)$. We do not detect the number of available tasks.

henryleberre: @sbryngelson just to be clear, when we use the `nodes` and `tasks_per_node`, these are the (long) names of the command-line options.

henryleberre: See https://github.com/MFlowCode/MFC/blob/43f1865cd98eb1408428a17eec45abb9fdfd7cec/toolchain/mfc/run/run.py#L86-L93

We pass `**ARGS()` where `ARGS()` returns a dictionary of command-line options. `**` turns these key/value pairs into keyword arguments.

sbryngelson:
Update: It actually didn't work with `srun` in interactive mode. I just didn't notice the change I made. Updating now.

toolchain/templates/delta.mako
sbryngelson: but what if one does not pass these options?

henryleberre: If none are specified, it is as if the user had passed in `-N 1 -n 1` (1 task on one _node_). This makes our expressions easier to handle across batch and interactive. This only creates trouble if running in interactive mode and requesting `-N X -n 1` (where $X > 1$), since it will use $X$ tasks across all nodes it has access to, not (necessarily) spawn one task on each of the $X$ nodes. We could see if specifying `-N` and `-n` to mpirun/srun works everywhere for interactive jobs. Fortunately, this is only a problem when requesting strange configurations in interactive mode. It is not a problem for Batch mode. I am open to your thoughts and suggestions about this.

sbryngelson: ah yes. ok, got it. makes sense.

henryleberre (approved):

PR #354, opened 2024-02-23
Fix integer overflow

wilfonba:
## Description

Fixes integer overflow for simulations with more than 2^31 cells.

### Type of change

- [ ] Something else

### Scope


sbryngelson:
Something about some of these changes doesn't quite seem right to me (I could be wrong). For example, if you write `int(p)`, this shouldn't do anything if `p` is already `int`. Likewise for other quantities.

wilfonba:
`int(p)` casts an 8-byte integer `p` to a 4-byte integer. The MPI routines only accept 4-byte integers.

sbryngelson:
> `int(p)` casts an 8-byte integer `p` to a 4-byte integer. The MPI routines only accept 4-byte integers.

Got it - though I see a couple other ones, like for `s_compute_finite_difference_coefficients`.

wilfonba:
I didn't change the integer types in simulation because the global cell count is never computed. `s_compute_finite_difference_coefficients` is used in both simulation and post_process and expects a 4-byte integer.

sbryngelson:
> I didn't change the integer types in simulation because the global cell count is never computed in simulation. `s_compute_finite_difference_coefficients` is used in both simulation and post_process and expects a 4-byte integer.

Does it not automatically up (or down) cast the input precision? That would make this much easier to maintain, otherwise every time one of the "special" variables (e.g., `p`) is used in both simulation and another code (pre/post process), and the developer needs to somehow know that they need to do `int(p)`.

wilfonba:
It does not automatically cast the inputs. I'll just make simulation use 8-byte integers for `m`, `n`, and `p`, then everything will be the same. The only reason I didn't was because adding all the casting in the MPI routines would be tedious.

wilfonba:
I'll just explain it at our meeting this afternoon. It'll be easier that way.

PR #359, opened 2024-02-24
Fix integer overflow for large simulations

wilfonba:
## Description

Fixes integer overflow for simulations with more than 2^31 cells.

### Type of change

### Scope

### DIFF
--- a/src/pre_process/m_checker.f90
+++ b/src/pre_process/m_checker.f90
@@ -135,9 +135,7 @@ contains
         elseif (n == 0 .and. p > 0) then
             call s_mpi_abort('Unsupported choice of the combination of '// &
                              'values for n and p. Exiting ...')
-        elseif ((m + 1)*(n + 1)*(p + 1) &
-                < &
-                2**(min(1, m) + min(1, n) + min(1, p))*num_procs) then
+        elseif (nGlobal< 2**(min(1, m) + min(1, n) + min(1, p))*num_procs) then
             call s_mpi_abort('Unsupported choice of the combination of '// &
                              'values for num_procs, m, n and p. '// &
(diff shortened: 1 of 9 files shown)


src/pre_process/m_checker.f90
```diff
@@ -135,9 +135,7 @@ subroutine s_check_inputs()
         elseif (n == 0 .and. p > 0) then
             call s_mpi_abort('Unsupported choice of the combination of '// &
                              'values for n and p. Exiting ...')
-        elseif ((m + 1)*(n + 1)*(p + 1) &
-                < &
-                2**(min(1, m) + min(1, n) + min(1, p))*num_procs) then
+        elseif (nGlobal < 2**(min(1, m) + min(1, n) + min(1, p))*num_procs) then
```
sbryngelson, line 138: how does this evaluate when the right side of `nGlobal < 2**(min(1, m) + min(1, n) + min(1, p))*num_procs` overflows? (since presumably this is what you are checking for?) or am I missing something else. actually, what are we actually checking for?

wilfonba: `2**(min(1, m) + min(1, n) + min(1, p))*num_procs` is just `(2^num_dims)*num_procs`, it should never overflow.

wilfonba: `nGlobal` replaces `(m+1)*(n+1)*(p+1)` which overflows at `2^31`

wilfonba: The other modified check is `(weno_order^num_dims)*num_procs`, which should also never overflow. It would take 17.2m cores with WENO 5 for it to overflow.

PR #364, opened 2024-03-03
Added memory to the Delta Mako file to run batch runs

mrodrig6:
## Description

Fixes #(issue) [optional]

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

## Checklist

- [ ] I have added example cases in `examples/` that demonstrate my new feature performing as expected
- [ ] New and existing tests pass locally with my changes, including with GPU capability enabled and disabled


sbryngelson:
@mrodrig6 what exactly does this do? how do you know it works? what doesn't work in the current version? @wilfonba you ran some big jobs on Delta, did you find a problem?

wilfonba:
I ran 2.25 billion cells in batch mode on 32 nodes (128 A100s) with the current mako file. The only thing I had to change was 4 CPU cores per task. Although I guess I should have shared that i needed to do this with Spencer before now.

sbryngelson:
> I ran 2.25 billion cells in batch mode on 32 nodes (128 A100s) with the current mako file. The only thing I had to change was 4 CPU cores per task. Although I guess I should have shared that i needed to do this with Spencer before now.

So how did you go about changing this?

sbryngelson:
> I changed it manually in the autogenerated batch script. Not exactly the most optimal way to do it.

I just meant what is the literal SLURM script line

wilfonba:
`#SBATCH --cpus-per-task=4`. I think this could also be fixed in how Mauro is suggested by requesting more memory per core.

sbryngelson:
I'm still unclear on what the right thing to do here is. Someone will need to explain to me what's going on in this PR and pros/cons (to the extent they exist).

wilfonba:
Replacing `--mem=208G` with `--mem=0` will always request all memory available on a node. This will cause longer wait times for interactive jobs using less than 4 GPUs because the scheduler will wait until a node with all of its memory is available. It might also charge for all 4 GPUs. I'm not sure how Delta handles this case.

sbryngelson:
I don't see why we are specifying memory at all. SLURM should handle this on its own if you specify all other requested resources (CPU cores and GPU devices). I would be quite shocked if this was not the case.

wilfonba:
Based on @haochey slack message, DELTA's slurm setup defaults to 1G per core.

sbryngelson:
can we specify memory per core, then?

wilfonba:
`--mem-per-cpu` specifies memory per CPU core in SLURM. It's still odd that it's a problem at all though. This isn't a problem on any other machines. I'm guessing it has to do with UIUC setting a default that encourages full utilization of shared nodes.

sbryngelson:
I'm going to merge this since it seems necessary to run any meaningful jobs and I don't know what the alternative is.

sbryngelson (approved):

PR #378, opened 2024-03-22
Update getting-started.md

sbryngelson:
## Description

Fix MacOS install docs

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,12 +56,12 @@ Using [brew](https://brew.sh), install MFC's modest set of dependencies:
 brew install wget python cmake gcc@13 mpich
 ```
 You're now ready to build and test MFC!
-Clone it to a convenient directory via
+Put it to a convenient directory via
 ```console
 git clone https://github.com/mflowcode/MFC.git
 cd MFC
 ```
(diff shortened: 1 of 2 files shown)


henryleberre:
@sbryngelson I couldn't quite remember why we depended on GNU `coreutils` on macOS so I tracked it down to https://github.com/MFlowCode/MFC/commit/c95195b9c6d44fbec0dd9435fe0d6a51fa890199. (`coreutils` was added to the list before we had the website docs).

To make CI reflect the docs (& ensure they are correct), could you update `.github/workflows/test.yml` ?

sbryngelson:
The reason is likely because in CI MacOS does not have "command line tools" installed. I installed MacOS on a VM (w/ GUI) and ensured these dependencies + command line tools were sufficient. I also found that Ubuntu requires an install of `python3-venv` in another VM.

henryleberre (approved):
@sbryngelson Sounds good to me, thanks for checking. `python3-venv` is likely not needed in CI because Github preloads it in their Docker image. (see https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md)

PR #379, opened 2024-03-22
More and Longer Benchmarks

### LINKED ISSUE #363: More benchmarks and longer benchmarks
* More benchmarks are needed (3D cases that are homogeneous in 2 of the dimensions might be easiest to setup).
* Longer benchmarks are needed (more time steps, e.g.) for GPU cases, which run quickly and are thus vulnerable to small runtime differences. This can be passed into the `case.py` benchmark case via an argument that contains the device type, like `examples/weak_scaling`.

wilfonba:
## Description

This PR replaces the existing shock droplet benchmark with 5 new benchmarking cases that run for more time-steps and combined cover a large portion of the code's features.

Fixes #363 [optional]

### Type of change

### Scope

## How Has This Been Tested?

This PR should run the benchmarks to test the changes.

## Checklist


sbryngelson:
It seems that we should add an argument option to the case file that indicates if it is running from a CPU or GPU. If it's CPU, it should use fewer time steps than the GPU case (such that the benchmarks run for a somewhat similar amount of time).

wilfonba:
Information about what's enabled in each case is already given in comments in bench.yaml. I'll add a GPU vs CPU input though.

sbryngelson:
can you rename the cases to reflect (roughly) what they're checking?

sbryngelson:
* Make names shorter: e.g. weno5viscous, weno3, ibm, etc. Whatever the 'main' part of the test is. 
* Remove the `output.txt` files
* Remove the `Test5/` directory

wilfonba:
The main part of the test is what's already listed in the test names. The test names include any previously untested feature. I'm not sure how to make them any shorter and still include all of the features that each case tests. I suppose I could abbreviate everything, but then it would be less clear what's going on (ie. `weno_order = wo`, `riemann_solver = rs` etc...) I'm unsure what `Test5` directory you're referring to. I can't find it anywhere, and searching the repo doesn't show anything.

output.txt
henryleberre, line 1: extra file!

toolchain/mfc/args.py
henryleberre, line 1: You should remove the `--mode` option. You can access the info you need in the case files through the first `dict` argument - like

```console
if dict["gpu"]:
    # gpu mode
```

The `--mode` option is dangerous because the binaries that are used to execute these benchmarks do not depend on that option. If you have a `--gpu` build, the `--gpu` binaries will be used, even if you specify a CPU `--mode`.

You could set the `mem` value in accordance with `dict["gpu"]` (within a case file) but I do see value in using these benchmarks for scaling tests, where, presumably, you would like to change `--mem` through a CLI option. So keeping the `--mem` options sounds like a good idea.

.github/workflows/phoenix/bench.sh
henryleberre, line 1: Per my other comment, you can revert this change.

henryleberre (changes requested):
Thanks for adding new benchmarking cases! These will be very useful.

output.txt
wilfonba: Fixed

toolchain/mfc/args.py
wilfonba: This comment seems to contradict your following comment to revert changes to `bench.sh`. I removed the `--mode` option, but I still need an if statement to specify different memory amounts for CPUs and GPUs. I'm also confused because you suggest keeping `--mem`. I updated `bench.sh` to read
```
if ["$job_device" == "gpu"]; then
    ./mfc.sh bench --mem 8 -j $(nproc) -o "$job_slug.yaml" -- -c phoenix $device_opts -n $n_ranks
else
    ./mfc.sh bench --mem 1 -j $(nproc) -o "$job_slug.yaml" -- -c phoenix $device_opts -n $n_ranks
fi
```
though. I think this accomplishes the main thing you were asking for, which was removing `--mode`. I think calculating the memory from within the case file makes it more difficult to see exactly what's going on.

.github/workflows/phoenix/bench.sh
wilfonba: See above

toolchain/mfc/args.py
henryleberre: You got me here. That's an oversight. I changed my mind halfway through writing that comment and totally forgot about that you were changing `--mem` for cpu/gpu.

sbryngelson (approved):

sbryngelson:
is this ready to merge @wilfonba ?

wilfonba:
I need to finish using 'DICT["gpu"]' in the case file. I was in a rush trying to finish it before going out and kept running into issues because I was rushing. I'll finish it and send a message either tonight or tomorrow.

wilfonba:
I'm having issues where the git clone for hdf5 fails more than 3 times. I'll try again tomorrow and hope whatever hairball is causing this is resolved.

wilfonba:
@sbryngelson This should be ready to merge as soon as all of the tests run. I haven't verified how long the GPU benchmarking takes, but the CI will indicate if I need to make any changes.

sbryngelson:
I think the benchmarking CI doesn't work until merged, correct? Though I guess it will run benchmark on the PR, so maybe it does...

wilfonba:
Yeah. It won't actually compare anything, but I think it will still run the cases. I can also just run it manually on phoenix.

henryleberre:
@wilfonba GitHub was probably rate-limiting your clone requests.

sbryngelson:
> @wilfonba GitHub was probably rate-limiting your clone requests.

I'm not sure -- Checking the GPU logs, all the jobs completed successfully. Only the toolchain failed (seemingly). It likely failed because the master benchmarks are different cases than the PR ones. We should throw a proper exception here (if that is indeed the case).

henryleberre:
@sbryngelson It did fail for that reason, but not in the manner you expected. The code already took care of situations where the cases ran in the two compared benchmarks differ, by only comparing cases whose names are common to both benchmark result files (taking the set intersection of case names). I tested this to make sure it worked.

The error here arose from an oversight - on my end. We also check for when the `metadata` field is different between compared benchmarks. It includes both the CLI invocation and the current modes MFC was in (e.g. `--gpu`, `--mpi`, `--debug`, etc.). When they differ, we produce a warning. However, the code I wrote also immediately aborted. I fixed this and gave the warning messages some more polish.

The new output looks like (ran locally):
```
~/dev/MFC $ ./mfc.sh bench_diff ~/Downloads/logs-gpu/master/bench-gpu.yaml ~/Downloads/logs-gpu/pr/bench-gpu.yaml 
mfc: OK > (venv) Entered the Python 3.12.2 virtual environment (>= 3.8).

      .=++*:          -+*+=.        | henryleberre@Henrys-Laptop.local [Darwin]
     :+   -*-        ==   =* .      | -----------------------------------------
   :*+      ==      ++    .+-       | --jobs 1
  :*##-.....:*+   .#%+++=--+=:::.   | --mpi
  -=-++-======#=--**+++==+*++=::-:. | --no-gpu
 .:++=----------====+*= ==..:%..... | --no-debug
  .:-=++++===--==+=-+=   +.  :=     | --targets pre_process, simulation, and post_process

 Comparing Bencharks: ../../Downloads/logs-gpu/master/bench-gpu.yaml is x times slower than ../../Downloads/logs-gpu/pr/bench-gpu.
 Warning: Metadata in lhs and rhs are not equal.
  This could mean that the benchmarks are not comparable (e.g. one was run on CPUs and the other on GPUs).
  lhs:
  * Invocation: bench -j 24 -o bench-gpu.yaml -- -c phoenix --gpu -g 0 1 -n 2
  * Modes:      debug=False gpu=False mpi=True
  rhs:
  * Invocation: bench --mem 1 -j 24 -o bench-gpu.yaml -- -c phoenix --gpu -g 0 1 -n 2
  * Modes:      debug=False gpu=False mpi=True
 Warning: Cases in lhs and rhs are not equal.
  * rhs cases: ibm, viscous_weno5_sgb_mono, 5eq_rk3_weno3_hllc, hypo_hll.
  * lhs cases: 3D_shockdroplet.
  Using intersection: set() with 0 elements.
                                                  
  Case   Pre Process   Simulation   Post Process  
 ──────────────────────────────────────────────── 
                                                  

```

henryleberre (approved):

PR #380, opened 2024-03-22
changes to include maximum number of iterations for phase change

JRChreim:
## Description

This pull request includes global variable "max_iter_pc_ts" to account for the maximum number of iterations within phase change, when this module is activated. It also include OpenACC statement for the GPU compilers

### Type of change

- [X] New feature (non-breaking change which adds functionality)

### Scope

- [X] This PR comprises a set of related changes with a common goal

## How Has This Been Tested?

The test to check this functionality was to add the global variable to the appropriate modules and checking whether the output was expected or not. After it was working for the CPU compilation, with was extended (and tested) to the GPU compilation. Note that this functionality only affects the .out file, whenever phase change is activated

**Test Configuration**:

* What computers and compilers did you use to test this: Bridges2 (CPU) and Delta (GPU)

## Checklist

- [X] I have added comments for the new code
- [X] I added Doxygen docstrings to the new code
- [N/A] I have made corresponding changes to the documentation (`docs/`)
- [N/A] I have added regression tests to the test suite so that people can verify in the future that the feature is behaving as expected
- [N/A] I have added example cases in `examples/` that demonstrate my new feature performing as expected
- [X] I ran `./mfc.sh format` before committing my code
- [X] New and existing tests pass locally with my changes, including with GPU capability enabled and disabled
- [X] This PR does not introduce any repeated code (it follows the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
- [X] I cannot think of a way to condense this code and reduce any introduced additional line count

- [X] Checked that the code compiles using NVHPC compilers
- [N/A] Ran the code on either V100, A100, or H100 GPUs and ensured the new feature performed as expected (the GPU results match the CPU results)
- [N/A] Enclosed the new feature via `nvtx` ranges so that they can be identified in profiles
- [N/A] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this PR
- [N/A] Ran my code using various numbers of different GPUs (1, 2, and 8, for example) in parallel and made sure that the results scale similarly to what happens if you run without the new code/feature

### DIFF
--- a/src/common/m_phase_change.fpp
+++ b/src/common/m_phase_change.fpp
@@ -119,6 +119,9 @@ contains
 
         !$acc declare create(p_infOV, p_infpT, p_infSL, sk, hk, gk, ek, rhok)
 
+        ! assigning value to the global parameter
+        max_iter_pc_ts = 0
+
         ! starting equilibrium solver
         !$acc parallel loop collapse(3) gang vector default(present) private(p_infOV, p_infpT, p_infSL, sk, hk, gk, ek, rhok,pS, pSOV, pSSL, TS, TSOV, TSatOV, TSatSL, TSSL, rhoe, dynE, rhos, rho, rM, m1, m2, MCT, TvF)
         do j = 0, m
@@ -396,6 +399,9 @@ contains
         ! common temperature
         TS = (rhoe + pS - mQ)/mCP
 
+        ! updating maximum number of iterations
+        max_iter_pc_ts = maxval((/ max_iter_pc_ts, ns/))
+
     end subroutine s_infinite_pt_relaxation_k ! -----------------------
 
     !>  This auxiliary subroutine is created to activate the pTg-equilibrium for N fluids under pT
@@ -517,6 +523,9 @@ contains
 
         ! common temperature
         TS = (rhoe + pS - mQ)/mCP
+
+        ! updating maximum number of iterations
+        max_iter_pc_ts = maxval((/ max_iter_pc_ts, ns/))
 
     end subroutine s_infinite_ptg_relaxation_k ! -----------------------
 
--- a/src/pre_process/m_global_parameters.fpp
+++ b/src/pre_process/m_global_parameters.fpp
@@ -216,6 +216,8 @@ module m_global_parameters
     type(pres_field) :: pb
     type(pres_field) :: mv
 
+    integer :: max_iter_pc_ts !< maximum number of iterations for phase change at each time-step    
+
 contains
 
     !>  Assigns default values to user inputs prior to reading
--- a/src/simulation/m_start_up.fpp
+++ b/src/simulation/m_start_up.fpp
@@ -1058,6 +1058,10 @@ contains
                 t_step - t_step_start + 1, &
                 t_step_stop - t_step_start + 1, &
                 t_step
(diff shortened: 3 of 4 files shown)


src/common/m_phase_change.fpp
```diff
@@ -518,6 +524,9 @@ contains
         ! common temperature
         TS = (rhoe + pS - mQ)/mCP
 
+        ! updating maximum number of iterations
+        max_iter_pc_ts = maxval((/ max_iter_pc_ts, ns/))
```
henryleberre, line 528: Do you also intend for `max_iter_pc_ts` to a variable a user can set higher than `ns` if they so desire? It looks like it because the `maxval` would otherwise be redundant. If so, you should add this as an input variable and distribute it to all ranks.

src/pre_process/m_global_parameters.fpp
```diff
@@ -216,6 +216,8 @@ module m_global_parameters
     type(pres_field) :: pb
     type(pres_field) :: mv
 
+    integer :: max_iter_pc_ts !< maximum number of iterations for phase change at each time-step
```
henryleberre, line 219: Given the `maxval` above, the (user-specified) input-variable would actually be a lower-bound for this value. Could be useful to separate these two for clarity.

src/simulation/m_start_up.fpp
```diff
@@ -1058,6 +1058,10 @@ contains
                 t_step - t_step_start + 1, &
                 t_step_stop - t_step_start + 1, &
                 t_step
+                if (relax) then
+                    !$acc update host(max_iter_pc_ts)
```
henryleberre, line 1062: This looks like you are debugging your feature. As far as I can see, `max_iter_pc_ts` is constant throughout the simulation so we shouldn't need to update the host copy and print its value to stdout every iteration. Am I missing something?

henryleberre (changes requested):
@JRChreim, it looks like you are still working on this. I left a few comments and am marking this PR as a draft for now. The most pressing issue is that the variable you are adding is unused.

src/common/m_phase_change.fpp
JRChreim: @henryleberre, no. The user has no control over this variable. It is set to track the maximum number of iterations phase change takes at every time step for the whole domain. This is the reason, it auto-updates. Should I still distribute it to all ranks? It does not seem to be needed, but I might be mistaken

src/pre_process/m_global_parameters.fpp
JRChreim: This is not user-specified, as commented above

src/simulation/m_start_up.fpp
JRChreim: This is not debugging. I am outputting max_iter_pc_ts into the ".out" file. Look at the image attached for reference

image

JRChreim:
@henryleberre , I believe I have clarified all your questions. Let me know if there is anything still unsure. Thank you!

henryleberre:
@JRChreim Thanks, I must have misread where the `maxval` was located. You are right.

PR #398, opened 2024-04-15
IBM Fix

haochey:
## Description

This is a simple fix on the pre_process checker and documentation to ensure that more than one immersed boundaries can be used properly. Please note that this is not a general fix. Boundaries immersed in the interface of two patches with different initial primitive variables are not allowed. 

### Type of change

### Scope

## How Has This Been Tested?

Adding one regular patch for the 2nd immersed circle boundary with an inflow.

- [x] Two immersed 2D circle with inflow

https://github.com/MFlowCode/MFC/assets/98496194/d1e94475-736d-45f8-96b9-0c3d27047e0b

**Test Configuration**:

* What computers and compilers did you use to test this: macOS and GCC

## Checklist

### DIFF
--- a/src/pre_process/m_initial_condition.fpp
+++ b/src/pre_process/m_initial_condition.fpp
@@ -137,20 +137,6 @@ contains
                     print *, 'Processing patch', i
                 end if
 
-                !> IB Patches
-                !> @{
-                ! Spherical patch
-                if (patch_ib(i)%geometry == 8) then
-                    call s_sphere(i, ib_markers%sf, q_prim_vf, .true.)
-                    ! Cylindrical patch
-                elseif (patch_ib(i)%geometry == 10) then
-                    call s_cylinder(i, ib_markers%sf, q_prim_vf, .true.)
-
-                elseif (patch_ib(i)%geometry == 11) then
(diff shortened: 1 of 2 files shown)


sbryngelson:
with this change, how can one have 2 IBs but 1 patch?

haochey:
@sbryngelson ib_markers were defined inside `s_assign_patch_primitive_variables` previously, and that is not necessary. Ib_markers are now defined in patches. This avoids the assignment of primitive variables to the ib patches again and allows using more than one ib's with less or one regular patches.

sbryngelson:
Looks good. @anandrdbz can you review this code?

sbryngelson (approved):

anandrdbz:
I think this looks good, one last thing that you might need to do once (and this isn't being done currently even in master), is to apply the immersed BC once at the start of the simulation. Right now for most of our cases,, this is not a problem, as the patch in which immersed boundary is placed is uniform. However, if the patch is non-uniform we could potentially run into some issues at the first time step

anandrdbz:
But this can also be done in a separate PR

haochey:
> But this can also be done in a separate PR

Thanks @anandrdbz. I agree that making a separate PR to attack this non-uniform issue will be easier as it will involve changes to `src/simulation`.

PR #409, opened 2024-05-12
-1.2k LOC in s_mpi_sendrecv_variables_buffers

henryleberre:
## Description

`s_mpi_sendrecv_variables_buffers` has too much repetition and is therefore very hard to read and maintain. This PR refactors it, removing 1.2k LOC.

### Scope

## How Has This Been Tested?

- [x] Test Suite
- [x] 3D QBMM

### DIFF
--- a/src/common/m_helper.fpp
+++ b/src/common/m_helper.fpp
@@ -33,7 +33,9 @@ module m_helper
  f_cross, &
  f_create_transform_matrix, &
  f_create_bbox, &
- s_print_2D_array
+ s_print_2D_array, &
+ f_xor, &
+ f_logical_to_int
 
 contains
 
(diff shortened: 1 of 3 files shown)


sbryngelson:
When this works it will be my favorite PR to date, second to the Frontier merge and my LAPACK commit.

henryleberre:
@sbryngelson I modified the 1D_qbmm example case to run on 8 A100s in 3D with different BCs. I compared the results against master using the `./mfc.sh packer` and the results are EXACTLY the same:
```
 Comparing test/master.txt to test/master.txt:
  
   OK: The two packs are within tolerance (abs: 1.00E-12, rel: 1.00E-12).
   Average error: abs: 0.00E+00, rel: 0.00E+00.
```

This test was important because halo-exchange buffers are different when `qbmm .and. .not. polytropic` evaluates to true.

sbryngelson:
wonderful. I think the Frontier case will run before maintenance and then I will merge. I won't merge anything else in the meantime anyway since all other PRs will also be waiting on Frontier CI.

Issue #385, opened 2024-04-05
Update `Final time` in the print statement to match `performance_expectations.md` and add benchmark case to `examples/`


sbryngelson:
Update `Final time` in the print statement in simulation to match `performance_expectations.md` as ns/gp/rhs-evaluation.

Add benchmark case to `examples/` which is apparently:

```python
#!/usr/bin/env python3

import json

# Configuring case dictionary
print(json.dumps({
    # Logistics ================================================================
    'run_time_info'                : 'T',
    # ==========================================================================

    # Computational Domain Parameters ==========================================
    # For these computations, the bubble is placed at the (0,0,0)
    # domain origin. Only one octant of a full spherical bubble is
    # used, with reflecting BC's at octant boundaries acting
    # as the other bubble regions.
    'x_domain%beg'                 : 0.E+00,
    'x_domain%end'                 : 4.E-03/1.E-03,
    'y_domain%beg'                 : 0.E+00,
    'y_domain%end'                 : 4.E-03/1.E-03,
    'z_domain%beg'                 : 0.E+00,
    'z_domain%end'                 : 4.E-03/1.E-03,
    # Grid stretching is used in the all coordinate directions 
    # to minimize computational costs. The grid is coarsened
    # away from the bubble / origin
    'stretch_x'                    : 'T',
    'a_x'                          : 4.E+00,
    'x_a'                          : -1.5E-03/1.E-03,
    'x_b'                          : 1.5E-03/1.E-03,
    'stretch_y'                    : 'T',
    'a_y'                          : 4.E+00,
    'y_a'                          : -1.5E-03/1.E-03,
    'y_b'                          : 1.5E-03/1.E-03,
    'stretch_z'                    : 'T',
    'a_z'                          : 4.E+00,
    'z_a'                          : -1.5E-03/1.E-03,
    'z_b'                          : 1.5E-03/1.E-03,
    'cyl_coord'                    : 'F',
    'm'                            : 99,
    'n'                            : 99,
    'p'                            : 99,
    'dt'                           : 0.2E-09,
    't_step_start'                 : 0,
    't_step_stop'                  : 100,
    't_step_save'                  : 100,
    # ==========================================================================
    
    # Simulation Algorithm Parameters ==========================================
    # Only two patches are necesssary, the background liquid and the
    # gas bubble
    'num_patches'                  : 2,
    # Use the 6 equation model
    'model_eqns'                   : 2  ,
    # 6 equations model does not need the K \div(u) term                    
[...]

Issue #299, opened 2024-01-10
Real-time simulation performance monitoring needs improvement


sbryngelson:
Real-time simulation performance monitoring needs improvement

* This code: https://github.com/MFlowCode/MFC/blob/4833971bf24f4f039f7443a780712b8231434114/src/simulation/m_time_steppers.fpp#L310-L314 
is repeated for each time stepper verbatim. Needs DRY principle.

* The metric of seconds is not particularly helpful and confusing to compute. The metric should be grid point per second per RHS evaluation. 
	* This should show up as a replacement for "Final time" here as well: https://github.com/MFlowCode/MFC/blob/4833971bf24f4f039f7443a780712b8231434114/src/simulation/m_start_up.fpp#L946 (and likely elsewhere).

* The metrics shown in the docs here https://github.com/MFlowCode/MFC/blob/master/docs/documentation/expectedPerformance.md should be updated accordingly.

sbryngelson:
Related to PR #411

sbryngelson:
This was closed by #432

PR #411, opened 2024-05-13
Fix for Issue #385

BucketofJava:
## Description

A fix for issue #385, updating time to be recorded in ns/gp/rhs. Also adding benchmark case.py to ./examples.

Related to https://github.com/MFlowCode/MFC/issues/299

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

- [x] Ran benchmark cases

**Test Configuration**:

Tested locally on WSL
Lenovo L380 Yoga
Intel(R) Core(TM) i5-8250U CPU @ 1.60GHz 

## Checklist


examples/benchmark/case.py
sbryngelson, line 1: Why is this file added?

src/simulation/m_start_up.fpp
```diff
@@ -1127,11 +1127,11 @@ contains
             if (num_procs == 1) then
                 time_final = time_avg
                 io_time_final = io_time_avg
-                print *, "Final Time", time_final
+                print *, "Final Time", time_final*1000000000, " ns/gp/rhs"
             else
                 time_final = maxval(proc_time)
                 io_time_final = maxval(io_proc_time)
-                print *, "Final Time", time_final
+                print *, "Final Time", time_final*1000000000, " ns/gp/rhs"
```
sbryngelson: it looks like this print statement doesn't need to go in the `if` or `else` condition and can just go after

examples/benchmark/case.py
henryleberre: I think this is because of how you worded #385. We could now maybe substitute it with the `./mfc.sh bench` feature. Thoughts?

src/simulation/m_start_up.fpp
henryleberre: Additionally, you can rewrite `1000000000` as `1.0d9` (should be more readable).

examples/benchmark/case.py
sbryngelson: @henryleberre I should probably re-read my own issue before critiquing the PR fixing it... I totally forgot about this. 

OK so it does make sense to add the case that is used for "performance expectations" to the benchmark suite I think (or still think, apparently). 

What do you mean by "We could now maybe substitute it with the `./mfc.sh bench` feature"? I assume running this command runs the four (now five?) benchmark cases, but perhaps that's not the case? I haven't recently run it myself actually.

src/simulation/m_start_up.fpp
sbryngelson: agree

sbryngelson:
Needs to run `./mfc.sh format` @BucketofJava

examples/benchmark/case.py
henryleberre: I hadn't really considered runtime.. I just took a look at the page on the website and this approach seems very reasonable to me, for the purposes I think this serves. With that in mind, it probably wouldn't be all that useful to provide a case-by-case breakdown for each benchmark configuration. I would probably suggest then that the case be added to the `benchmarks` folder.

sbryngelson: One can only compare to the expected performance page if you are running very close to the same case, so we should include it in that sense. However, this PR does not do this fully, and the case file needs a proper name.

BucketofJava:
Updated the code according to the comments; For the benchmark case, I moved it to the benchmarks folder, named it "default" as a placeholder and added it to the yml.

sbryngelson:
You are failing the `Pretty` CI because you didn't follow my comment above (or in the PR template). How long does it take to run this benchmark case? On CPU? GPU?

sbryngelson:
This PR doesn't actually compute the other things I mention in the issue:
```

Update: Probably want to switch both the code and the .md file in the docs to be ns per grid-point per equation per rhs evaluation (I updated the docs to have this already but without the "per-equation" part).
```

BucketofJava:
Sorry, I did not see the per-equation part before. I have run ./mfc.sh format with every commit, however I did not run it with -j ${nproc} as in the test case, this commit has done this, so hopefully the issue should be fixed.

sbryngelson:
@BucketofJava if you change the docs markdown page to have _per equation_ then you also need to change all of the performance numbers in that table (dividing them by the number of equations). I think this means dividing them all by 8?

sbryngelson:
PR merges are on hold until benchmark CI is working (issue #419)

sbryngelson:
@BucketofJava or someone... can you force sync this fork with MFC root master so I can merge?

sbryngelson:
@AiredaleDev can you assist with finishing off this PR?

AiredaleDev:
Sure thing -- it merged cleanly with master with no issue. I'd like to push the changes to this PR but I need access to @BucketofJava 's fork, so I made another PR instead: #432

sbryngelson:
Superseded by #411

PR #414, opened 2024-05-14
Fix for issue #395

### LINKED ISSUE #395: CI Benchmarking reports opposite results
The CI Benchmarker reports `master is X times faster than PR` (or something like this)  but in reality it is actually reporting `master is X times slower than PR`. This is a lhs vs rhs swap issue in `bench.py`.

okBrian:
## Description

Fixed the clarity issue for the CI benchmark comparison message when comparing a PR and master.

### Example:

* Before
```
Comparing Bencharks: master/bench-cpu.yaml is x times slower than pr/bench-cpu.yaml.
```
* Expected After
```
Comparing Benchmarks:
    1.5x indicates pr/bench-cpu.yaml is 1.5-times as fast as master/bench-cpu.yaml (so pr/bench-cpu.yaml is faster than master/bench-cpu.yaml).
    0.5x indicates pr/bench-cpu.yaml is 0.5-times as fast as master/bench-cpu.yaml (so pr/bench-cpu.yaml is slower than master/bench-cpu.yaml).
```

Fixes #395 

### Type of change

### Scope

## How Has This Been Tested?

Not sure how to run the diff benchmark comparison without the github runner

**Test Configuration**:

Compiled on a 

Ubuntu 22.04.4 using Docker
Intel i7-1165G7

## Checklist


sbryngelson:
@henryleberre can confirm but I think this might be backwards. lhs -> master and rhs -> PR. So, 0.5x indicates the PR is faster than master and 1.5x indicates the PR is slower. 

Code:
https://github.com/MFlowCode/MFC/blob/64d9677e58f26a157e207afbdcc705e5cab90c7f/toolchain/mfc/bench.py#L108-L122

and:
https://github.com/MFlowCode/MFC/blob/f23fceb103412864a9846fd75fdc13261699e147/toolchain/mfc/args.py#L135-L138

and:
https://github.com/MFlowCode/MFC/blob/f23fceb103412864a9846fd75fdc13261699e147/.github/workflows/bench.yml#L38

henryleberre:
Let's consider an example:

```console
$ ./mfc.sh bench_diff lhs rhs
```

where `lhs` and `rhs` are YAML files. By the way, "lhs" and "rhs" just stand for "left-hand side" and "right-hand side". We could maybe opt for better names.

Say that for a given example case, you have:
```
     simulation
lhs:    1.0s
rhs:    0.5s 
```

So, $\frac{\text{lhs}}{\text{rhs}}=\frac{1.0}{0.5}=2$. So $\frac{\text{lhs}}{\text{rhs}}$ is the speedup going from `lhs` to `rhs`. In other words, `lhs` is what you are comparing `rhs` to. This implies that the text I had written and the text @okBrian wrote are both correct, although mine was worded in a way that might be confusing. It was like "LHS is X times SLOWER than RHS" rather than "RHS is X times FASTER than LHS" (@okBrian's version).

As a sidenote, there are a lot of `cons.write` calls that follow each other. These can be joined to be more readable like:
```python
cons.write(f"""\
Line 1
Line 2
""")

sbryngelson:
There should just be a line of text above the results that says "numbers < 1 indicate the PR is faster than master, > 1 indicate PR is slower" (if that is actually the case)

sbryngelson:
PR merges are on hold until benchmark CI is working (issue #419)

henryleberre:
@okBrian Why did you revert [1e30c18](https://github.com/MFlowCode/MFC/pull/414/commits/1e30c186929e8702799fb1e1d64747e8431dcf3a) ?

okBrian:
The past few tests have either failed or have the wrong output so I'm just making sure that's not the issue.

sbryngelson:
@okBrian please sync your branch with master, which I just merged to (that's why your current CI tests are failing)

sbryngelson (approved):

PR #421, opened 2024-05-16
Modify Templating to Work with Omniperf and Rocprof

wilfonba:
## Description

Changes how profiling is added to the template so that profiling works with AMDs tools

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

I tested this with `./mfc.sh run examples/1D_sodshocktube` -n 1 --nsys` on Wingtip and with `./mfc.sh run examples/1D_sodshocktube -n 1 --omni` on Frontier.

## Checklist


toolchain/mfc/run/run.py
```diff
@@ -102,7 +102,7 @@ def __generate_job_script(targets, case: input.MFCInputFile):
         case=case,
         MFC_ROOTDIR=MFC_ROOTDIR,
         qsystem=queues.get_system(),
-        profiler=__profiler_prepend(),
+        profiler=' '.join(__profiler_prepend()),
```
henryleberre: This should probably become:
```python
profiler=' '.join([ shlex.quote(_) for _ in __profiler_prepend()])
```
so that arguments that contain whitespace are not split, and those with special characters are not misinterpreted.

toolchain/templates/bridges2.mako
```diff
@@ -39,9 +39,9 @@ echo
     ${helpers.run_prologue(target)}
 
     % if not mpi:
-        (set -x; ${' '.join([f"'{x}'" for x in profiler ])} "${target.get_install_binpath(case)}")
```
henryleberre, line 42: `set -x` is pretty useful. If this is only an issue on Frontier, it would be nice to keep it on other systems. There might also be a way of fixing the issue there, I have not looked into it.

henryleberre (changes requested):

toolchain/templates/bridges2.mako
wilfonba: I'm not sure I see the value of switching to a less portable tool when this approach works. This won't just be a problem on Frontier either, it will be a problem on all AMD machines, and I don't imagine this will go away anytime soon...

henryleberre: Just to be clear, I was only talking about the `set -x` part. The rest of the changes are great. It is my understanding that `set -x` only causes trouble on Frontier, not any other system. Correct me if I'm wrong!

toolchain/mfc/run/run.py
```diff
@@ -102,7 +102,7 @@ def __generate_job_script(targets, case: input.MFCInputFile):
         case=case,
         MFC_ROOTDIR=MFC_ROOTDIR,
         qsystem=queues.get_system(),
-        profiler=__profiler_prepend(),
+        profiler=' '.join([ shlex.quote(_) for _ in __profiler_prepend()])
```
henryleberre: Actually I think `shlex.join(__profiler_prepend())` would do the trick :). I just learned about its existence.

toolchain/templates/bridges2.mako
wilfonba: Presumably we'll get access to more AMD machines in the future that will use the same profiling software as Frontier. I don't imagine AMD machines are going away any time soon.

sbryngelson: did @henryleberre and @wilfonba come to a consensus on this?

sbryngelson:
I'm happy with this when @henryleberre is. 

I would like the GPU benchmark to pass before merging, but all GPU Phoenix cases are failing right now because they cannot get on a node.

toolchain/templates/default.mako
```diff
@@ -33,28 +33,28 @@ warn "Consider using a different template via the $MAGENTA--computer$COLOR_RESET
...
+            (set -x; j${profiler}   \
                 jsrun --nrs          ${tasks_per_node*nodes}      \
                       --cpu_per_rs   1                            \
                       --gpu_per_rs   ${1 if gpu else 0}           \
                       --tasks_per_rs 1                            \
                       ${' '.join([f"'{x}'" for x in ARG('--') ])} \
                       "${target.get_install_binpath(case)}")
         elif [ "$binary" == "srun" ]; then
-            (set -x; ${' '.join([f"'{x}'" for x in profiler ])}  \
+            (et -x; {profiler}  \
```
henryleberre: Missing an "s" to "set"

toolchain/templates/default.mako
```diff
@@ -33,28 +33,28 @@ warn "Consider using a different template via the $MAGENTA--computer$COLOR_RESET
     ${helpers.run_prologue(target)}
 
     % if not mpi:
-        (set -x; ${' '.join([f"'{x}'" for x in profiler ])} "${target.get_install_binpath(case)}")
+        (${profiler} "${target.get_install_binpath(case)}")
     % else:
         if [ "$binary" == "jsrun" ]; then
-            (set -x; ${' '.join([f"'{x}'" for x in profiler ])}   \
+            (set -x; j${profiler}   \
```
henryleberre: Extra "j" (typo)

henryleberre (changes requested):

toolchain/templates/bridges2.mako
henryleberre: Looks like @wilfonba added these back in, even to the frontier template.

sbryngelson: as long as it works...

toolchain/templates/default.mako
wilfonba: oopsies

toolchain/templates/default.mako
wilfonba: oopsies

toolchain/templates/bridges2.mako
wilfonba: Yes. The problem was me copying the set with the profiler prepend and not leaving it at the beginning of of the command. It was a misunderstanding of what set was doing.

toolchain/templates/default.mako
```diff
@@ -33,28 +33,28 @@ warn "Consider using a different template via the $MAGENTA--computer$COLOR_RESET
     ${helpers.run_prologue(target)}
 
     % if not mpi:
-        (set -x; ${' '.join([f"'{x}'" for x in profiler ])} "${target.get_install_binpath(case)}")
+        (${profiler} "${target.get_install_binpath(case)}")
```
henryleberre, line 36: I think you also forgot to add `set -x;` here

henryleberre: I think you also forgot to add back `set -x` here.

wilfonba: oopsies. Fixed and did a more careful check on my own before pushing this time.

henryleberre (approved):
lgtm

PR #422, opened 2024-05-17
Silo: Reference official LLNL/Silo repository

henryleberre:
## Description

Currently, Silo is fetched from my fork - which is not ideal. This had to be done to include some of my fixes for macOS and Frontier. Although the former set of changes was upstreamed, there is no *non-broken* release that includes it.. We therefore reference the commit hash from my PR. The Frontier hotfix cannot be upstreamed because CMake users are not supposed to manually specify `CMAKE_MODULE_PATH`s at configure-time. I cannot recall the specifics of the issue, but there may be a more permanent solution available. In the meantime, we manually patch the one-line hotfix in.

### Type of change

- [x] Something else

### Scope

### DIFF
--- a/toolchain/dependencies/CMakeLists.txt
+++ b/toolchain/dependencies/CMakeLists.txt
@@ -59,14 +59,16 @@ endif()
 
 # SILO
 if (MFC_SILO)
+    find_package(Git REQUIRED)
+
     # If we are using the CCE, HDF5 is not built, and we wish to find
     # the system's cray-hdf5. Otherwise, we point SILO to find HDF5 in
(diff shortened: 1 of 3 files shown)


sbryngelson (approved):

henryleberre:
@sbryngelson Looking at the logs, I found a bug I had mistakenly (re)introduced. When `./mfc.sh test` invokes `./mfc.sh run` for each test, it wasn't passing the `--no-build` option. So it was attempting to rebuild, resulting in it reinstalling the (same) binaries. This can cause a few race conditions that cause spurious failures when testing with multiple threads. It's also inefficient. I pushed a fix in this PR.

Issue #396, opened 2024-04-12
Delta GPUs results give NaNs in the viscous sub-grid bubble benchmark case whereas Phoenix GPUs do not


sbryngelson:
Delta GPUs results give NaNs in the viscous sub-grid bubble benchmark case whereas Phoenix GPUs do not.

This is regardless of "memory size" (I've checked 4gb).

I've tested A100s and A40s on Delta, both give the issue discussed further on Slack.

I tested A100s and V100s on Phoenix, both of which _do not_ give the issue.

Both computers use NVHPC 22.11.

Error is this:

```console
 [ 40%]  Time step      358 of 901 @ t_step = 357
 [ 40%]  Time step      359 of 901 @ t_step = 358
 [ 40%]  Time step      360 of 901 @ t_step = 359
Warning: ieee_inexact is signaling
ERROR STOP NaN(s) in timestep output.
 NaN(s) in timestep output.            0            0            0            1
             0          360          198           99           99
--------------------------------------------------------------------------
Primary job  terminated normally, but 1 process returned
a non-zero exit code. Per user-direction, the job has been aborted.
--------------------------------------------------------------------------
```

One can run this case via something like
`./mfc.sh run benchmarks/viscous_weno5_sgb_mono/case.py 4 -t pre_process simulation -c delta --gpu`
if you are already on a node with GPUs and have loaded the appropriate modules.

sbryngelson:
Update: This issue is associated with `parallel_IO = 'T'`. I witnessed it again on a Rogues Gallery GH200 chip w/ NVHPC 24.1.

sbryngelson:
I'm not sure if this is still "broken" or not.

henryleberre:
@sbryngelson I'm pretty sure `--case-optimization` is enabled by default in `bench`.

wilfonba:
The logs indicate that case optimization is enabled on Phoenix for the benchmarking. There's recompilation of code in cases that I would expect to see recompilation due to case optimization.

sbryngelson:
Nevermind, you're both right and it fails with and without case optimization on Delta (and presumably other computers).

PR #425, opened 2024-05-20
Mock PR to test benchmarking

wilfonba:
This is a mock PR that checks the benchmarking PR. It does the following:

- Adds a 60 second delay to pre_process for the hypo_hll case
- Adds a print statement to the 5eq_rk3_weno3_hllc case file to cause failure in input generation
- Calls s_mpi_abort in simulation for the viscous_weno5_sgb_mono case
- Allows the IBM case to run to completion

### DIFF
--- a/benchmarks/5eq_rk3_weno3_hllc/case.py
+++ b/benchmarks/5eq_rk3_weno3_hllc/case.py
@@ -172,7 +172,7 @@ Nt = AS * SF
 
 # total simulation time - s. Note that tend >= tendA
 tend = Nt * dt
-
+print*, "THIS CAUSES AN ERROR"
 # Configuring case dictionary
 print(json.dumps({
     # Logistics ================================================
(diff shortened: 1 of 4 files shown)


sbryngelson (approved):

wilfonba:
I would expect all of the test suites to fail on the first bubbles case since I added an MPI_Abort if bubbles = T. I expected the benchmarks to handle this though since it runs the 4 cases independently. Looking at the CI it appears that neither benchmark actually got a node and ran at all. It could be the case that I needed to introduce a seg fault rather than an MPI_Abort, but I'll have to see what happens when the benchmarks actually run to know.

sbryngelson:
Please sync this branch with master

sbryngelson:
I'd like to see this check get to the point where it gets to point where it actually prints benchmark results (presumably with the aforementioned failures).

sbryngelson:
The only thing I really want to be checked now is if the benchmark cases run on Delta/non-Phoenix computers.

wilfonba:
I'll do a run on delta

sbryngelson:
I just did. Check the issue for the latest at #396

sbryngelson:
this can be closed due to #497 thanks @ChrisZYJ

PR #426, opened 2024-05-20
Test benchmark in benign conditions

sbryngelson:
This tests that benchmarking works as expected when there is no meaningful change to the underlying code.


sbryngelson:
checks out, closing.

PR #407, opened 2024-05-10
IB-DB Treatment

haochey:
## Description

This PR adds the treatment for cases where immersed boundaries touch the domain boundaries. 

### Type of change

Please delete options that are not relevant.

- [x] Bug fix (non-breaking change which fixes an issue)

### Scope

- [x] This PR comprises a set of related changes with a common goal

If you cannot check the above box, please split your PR into multiple PRs that each have a common goal.

## How Has This Been Tested?

Use 2D rectangle and circle as well as 3D cylinder and sphere IBs touching the domain boundary w/ shock wave and high amplitude acoustic wave.

https://github.com/MFlowCode/MFC/assets/98496194/1feab706-34a2-4f03-91f4-f0c6e389aee1

https://github.com/MFlowCode/MFC/assets/98496194/4595cf58-730a-494b-bca4-550672b076c7

https://github.com/MFlowCode/MFC/assets/98496194/6d3589e2-d4ed-4b3e-a124-4773ba67f357

https://github.com/MFlowCode/MFC/assets/98496194/c62b4723-cf31-4acf-afc0-3fd23507506a

https://github.com/MFlowCode/MFC/assets/98496194/caec1540-e27b-46f6-9767-af65aa19b1b3

https://github.com/MFlowCode/MFC/assets/98496194/ae8feb88-a9d2-4fb6-be94-13c797623428

https://github.com/MFlowCode/MFC/assets/98496194/26a3efc3-002b-4430-aed0-f324901e5c7f

**Test Configuration**:

Sample 2D cases: 149*49

```
    'time_stepper'                 : 3,
    'weno_order'                   : 5,
    'weno_eps'                     : 1.E-16,
    'weno_Re_flux'                 : 'F',
    'weno_avg'                     : 'F',
    'avg_state'                    : 2,
    'mapped_weno'                  : 'F',
    'null_weights'                 : 'F',
    'mp_weno'                      : 'F',
    'riemann_solver'               : 2,
    'wave_speeds'                  : 1,
    'bc_x%beg'                     : -6,
    'bc_x%end'                     : -6,
    'bc_y%beg'                     : -15,
    'bc_y%end'                     : -15,
    'ib'		                   : 'T',
    'num_ibs'                      :  1, 
    'patch_icpp(1)%geometry'       : 3,
    'patch_icpp(1)%x_centroid'     : x1/2, 
    'patch_icpp(1)%y_centroid'     : x2/2,
    'patch_icpp(1)%length_x'       : x1,
[...]

PR #430, opened 2024-05-22
Remove grid stretch parameters from simulation inputs

haochey:
## Description

This PR removes the grid stretch parameters `xyz_{ab}` from the simulation input parameters in `case_dicts.py`.

### Type of change

### Scope

## How Has This Been Tested?

It builds successfully on my local MacOS. Simulations with stretched grid are not crashing.

* What computers and compilers did you use to test this: MacOS 12.2.1

### DIFF
--- a/toolchain/mfc/run/case_dicts.py
+++ b/toolchain/mfc/run/case_dicts.py
@@ -123,7 +123,7 @@ for cmp in ["x", "y", "z"]:
     SIMULATION.append(f'bc_{cmp}%ve1')
     SIMULATION.append(f'bc_{cmp}%ve2')
     SIMULATION.append(f'bc_{cmp}%ve3')
-    for prepend in ["domain%beg", "domain%end", "a", "b"]:
+    for prepend in ["domain%beg", "domain%end"]:
         SIMULATION.append(f"{cmp}_{prepend}")
 
 for wrt_id in range(1,10+1):


sbryngelson:
are you sure simulation doesn't need grid stretching parameters? it's easy to check... I recommend just changing to notation so we don't repeat letters

haochey:
> are you sure simulation doesn't need grid stretching parameters? it's easy to check... I recommend just changing to (IBM) notation so we don't repeat letters

Hi Spencer, I just made a couple tests for IB touching DBs on stretched girds. All the configurations are the same as the test listed in [PR#407](https://github.com/MFlowCode/MFC/pull/407) except for the stretching configuration: 

```
    'stretch_x'                : 'T',
    'a_x'                          : 7,
    'x_a'                          : 0.5*x1,
    'x_b'                          : 1.5*x1,  
    'stretch_y'                : 'T',
    'a_y'                          : 3,
    'y_a'                          : 0.9*x2,
    'y_b'                          : 1.1*x2, 
```

https://github.com/MFlowCode/MFC/assets/98496194/40cbfc1a-b0ca-4f24-8299-c6fc2501a939

 It looks fine to me. The current simulation stage is not using any stretching parameters and the domain sizes are the only additional parameters it needs for IBM.

sbryngelson:
@haochey, thank you - this seems fine to me, though I'm a little confused about the whole situation. @lee-hyeoksu said you added `{xyz}_{ab}` to the case parameter input file for the domain boundary treatment. What are those parameters used for? 

If they are used for something related to IBs, you presumably cannot use both those _and_ domain stretching simultaneously (you would have to pick one or the other). Right?

haochey:
@sbryngelson Thanks for the message.  `xyz_{ab}` are not used for IBs. I believe they are just used in the pre_process stage. 

When I try to add `xyz_{domain%beg}` and `xyz_{domain%end}`, I mistakenly added `xyz_{ab}` to the simulation inputs in `case_dicts.py` as @lee-hyeoksu mentioned. 

Therefore, simply deleting them from the `case_dicts.py` will fix @lee-hyeoksu's issue and won't affect the domain boundary treatment.

sbryngelson:
got it. thanks @haochey. can you sync your fork/branch with master so it runs benchmarking properly and then i can merge

haochey:
It looks like my fork/branch has already synchronized with master. I am not sure why the benchmarking is not running properly.

sbryngelson:
I see - I am re-running them.

PR #436, opened 2024-05-25
Fix type in default.mako

wilfonba:


sbryngelson:
idk why exactly but you have a merge conflict... probably a function of how you are submitting this PR with stale old commits along for the ride

PR #437, opened 2024-05-25
Improve error reporting for case files.

### LINKED ISSUE #424: Validate case dictionaries
Currently, we have a file in the toolchain named [toolchain/mfc/run/case_dicts.py](toolchain/mfc/run/case_dicts.py) that lists all the keys a case dictionary can have for each target. This is done by programmatically generating the lists on the fly. We need to do this because each target has its own set of inputs (keys) it accepts. If you specify one that it does not know about, it will fail to load the input data.

However, we do not check data-types. This can cause trouble when Fortran tries to read values from the `.inp` files we generate. Depending on the compiler, the error message can be very misleading, sometimes pointing to innocent lines in the `.inp` files.

To mitigate this issue, we could define a schema in Python for these case dictionaries and use them to validate the JSON we get from cases. There are a few libraries that do this. One of them is `jsonschema`. It might not be the most appropriate.

There might be a better way of doing this I cannot think of (perhaps with a larger-scoped change).

AiredaleDev:
## Description

If a case file's parameter is assigned a value of the wrong type, MFC will now correctly report the mismatching field. Previously, it would point to the next parameter, even if it typechecks.

Fixes #424

### Type of change

- New feature (non-breaking change which adds functionality)

### Scope

## How Has This Been Tested?

- [x] Assertions that ensure that typechecking logic covers all parameters
- [x] Modifying and running a case file with (in)correct values.

**Test Configuration**:

My personal laptop:
> Ryzen 7 7840U
> GCC 11.4.0

## Checklist

- [x] I have made corresponding changes to the documentation (`docs/`) -- (I caught a typo while grepping through the codebase: Loginal -> Logical)
- [x] I cannot think of a way to condense this code and reduce any introduced additional line count -- I considered an automated solution to get the types from the Fortran codebase that uses `grep` and `awk`, but it seemed too brittle and complicated.


sbryngelson:
This _seems_ right to me, but the rightful person to review this is @henryleberre... nice job @AiredaleDev

AiredaleDev:
Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?

AiredaleDev:
> This _seems_ right to me, but the rightful person to review this is @henryleberre... nice job @AiredaleDev

Thanks! Henry and I reviewed it last night, there's only the one outstanding thing I just discovered before it should be ready to merge.

sbryngelson:
> Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?

Can you be a bit more precise... how long are we talking? 

and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?

AiredaleDev:
> > Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?
> 
> Can you be a bit more precise... how long are we talking?
> 
> and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?

Roughly a minute or two, though it's only for `./mfc.sh test` since it generates a lot of `MFCInputFile`s `./mfc.sh build` and `./mfc.sh run` are fine because they validate at most one case file.

sbryngelson:
> > > Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?
> > 
> > 
> > Can you be a bit more precise... how long are we talking?
> > and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?
> 
> Roughly a minute or two, though it's only for `./mfc.sh test` since it generates a lot of `Case`s. `./mfc.sh build` and `./mfc.sh run` are fine because they validate at most one case file.

That is a lot. Can we switch to type checking at every test initialization rather than all at once in the beginning? That would probably fix it... distributing 1 minute over 150 tests makes it negligible. Need to be careful with the multithreading going on, through `./mfc.sh test -j X`

AiredaleDev:
> > > > Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?
> > > 
> > > 
> > > Can you be a bit more precise... how long are we talking?
> > > and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?
> > 
> > 
> > Roughly a minute or two, though it's only for `./mfc.sh test` since it generates a lot of `Case`s. `./mfc.sh build` and `./mfc.sh run` are fine because they validate at most one case file.
> 
> That is a lot. Can we switch to type checking at every test initialization rather than all at once in the beginning? That would probably fix it... distributing 1 minute over 150 tests makes it negligible. Need to be careful with the multithreading going on, through `./mfc.sh test -j X`

We appear to be already doing that *in addition* to the initial check. Removing these means we'll have to go back to waiting until after MFC is built to validate these inputs, but I think that's fine.

sbryngelson:
> > > > > Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?
> > > > 
> > > > 
> > > > Can you be a bit more precise... how long are we talking?
> > > > and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?
> > > 
> > > 
> > > Roughly a minute or two, though it's only for `./mfc.sh test` since it generates a lot of `Case`s. `./mfc.sh build` and `./mfc.sh run` are fine because they validate at most one case file.
> > 
> > 
> > That is a lot. Can we switch to type checking at every test initialization rather than all at once in the beginning? That would probably fix it... distributing 1 minute over 150 tests makes it negligible. Need to be careful with the multithreading going on, through `./mfc.sh test -j X`
> 
> We appear to be already doing that _in addition_ to the initial check. Removing the initial check means we'll have to go back to waiting until after MFC is built to validate the case inputs, but I think that's fine since the alternative is sequentially validating all 150 inputs. Each test has a unique file that it reads iirc, so multithreading should behave fine.
> 
> I also just discovered we could validate only the changed parameters upfront, but the test runner is already going to run the check used during a normal run and the one it uses for generating its inputs, so it would ultimately do redundant work.

I don't know what you mean by it already does that, can you be more specific?

MFC is only built once before the test suite runs.

I think trying to get the toolchain to only check for changed inputs is more complex than appropriate for this.

AiredaleDev:
> > > > > > Adding the check in the constructor significantly extends the pre-build time for the test suite, enough for the user to think something is wrong (validating json is apparently quite expensive). For performance reasons, I may have to remove the check in the constructor, is there a way we can run fewer checks?
> > > > > 
> > > > > 
> > > > > Can you be a bit more precise... how long are we talking?
> > > > > and I guess you mean it expands the time between initiating `./mfc.sh build` and it actually starting to build? or `./mfc.sh run` and it actually starting to run?
> > > > 
> > > > 
> > > > Roughly a minute or two, though it's only for `./mfc.sh test` since it generates a lot of `Case`s. `./mfc.sh build` and `./mfc.sh run` are fine because they validate at most one case file.
> > > 
> > > 
> > > That is a lot. Can we switch to type checking at every test initialization rather than all at once in the beginning? That would probably fix it... distributing 1 minute over 150 tests makes it negligible. Need to be careful with the multithreading going on, through `./mfc.sh test -j X`
> > 
> > 
> > We appear to be already doing that _in addition_ to the initial check. Removing the initial check means we'll have to go back to waiting until after MFC is built to validate the case inputs, but I think that's fine since the alternative is sequentially validating all 150 inputs. Each test has a unique file that it reads iirc, so multithreading should behave fine.
> > I also just discovered we could validate only the changed parameters upfront, but the test runner is already going to run the check used during a normal run and the one it uses for generating its inputs, so it would ultimately do redundant work.
> 
> I don't know what you mean by it already does that, can you be more specific?
> 
> MFC is only built once before the test suite runs.
> 
> I think trying to get the toolchain to only check for changed inputs is more complex than appropriate for this.

Agreed. If I were to go about it, I'd have to move check to two unintuitive places to perform a full check in the case of `run` and only a partial, upfront check in the case of `test` (call `validate` on the `mods` parameter of `TestCase`'s constructor). Moving the check into `MFCInputFile`'s constructor saves us the upfront minute, so I'll call it done.

sbryngelson:
If I understand correctly, you have it setup so it builds _all_ of the test cases and then checks _all_ of the parameters for _all_ the tests before actually starting running the tests, whereas the existing (pre-PR) implementation didn't check the parameters of the test cases at all? or did the pre-PR implementation only test some of them somehow? I'm not more confused about the nature of the issue #424

AiredaleDev:
> If I understand correctly, you have it setup so it builds _all_ of the test cases and then checks _all_ of the parameters for _all_ the tests before actually starting running the tests, whereas the existing (pre-PR) implementation didn't check the parameters of the test cases at all? or did the pre-PR implementation only test some of them somehow? I'm now more confused about the nature of the issue #424

Okay, I'm not explaining it right. The parameters were initially validated in the Fortran codebase, but the solution was quite hacky and would often point to benign parameters. In this PR, each run now checks all of its parameters using the schema, which provides clear error reporting, and then launches MFC if everything is okay -- this is true regarding normal runs and test suite.

I unintentionally made a change that would do as you described, which introduced the minute-long delay in the test suite.  I fixed this delay with my most recent commits.

sbryngelson:
ah - got it! I understand now.

henryleberre:
@AiredaleDev @sbryngelson I think the best approach here is to define a method under `Case` called `validate()`. Code that makes use of `Case` objects should have the burden of calling it as a sanity check.

henryleberre:
Also @AiredaleDev, I am to blame for the inclusion of (most of) the unused variables you found (`chem_wrt`, `spec_pp`, ...). These were unintentionally added when upstreaming some changes from my branch that is adding chemistry.

AiredaleDev:
> @AiredaleDev @sbryngelson I think the best approach here is to define a method under `Case` called `validate()`. Code that makes use of `Case` objects should have the burden of calling it as a sanity check.

@henryleberre Normally I would agree, but if I apply this sanity check to all code that constructs a `Case` (i.e. also a `TestCase`) then this line in the test runner forces the user to wait almost one whole minute before it begins. Apparently `jsonschema` is a bit slow.

```py
# In toolchain/mfc/test/test.py
def test():
    # pylint: disable=global-statement, global-variable-not-assigned
    global nFAIL

    # This is the line I have to work around.
    # It runs before anything in parallel happens, so all 150 cases are checked sequentially.
    cases = [ _.to_case() for _ in list_cases() ]

# In toolchain/mfc/test/case.py
    def to_case(self) -> TestCase:
        dictionary = self.mods.copy()
        if self.path:
            dictionary.update(input.load(self.path, self.args).params)

            for key, value in dictionary.items():
                if not isinstance(value, str):
                    continue

                for path in [value, os.path.join(os.path.dirname(self.path), value)]:
                    path = os.path.abspath(path)
                    if os.path.exists(path):
                        dictionary[key] = path
                        break
        # This code "should" check the dictionary, but if we check the dictionary here
        # we wait ~45 seconds before the test runner starts.
        return TestCase(self.trace, dictionary, self.ppn, self.rebuild)
```

The test runner will eventually call `run` anyway, which ends up reading and validating each case dictionary in parallel before launching the simulation (i.e. each concurrent run of MFC checks its own parameters.)

I added a `validate` method with the last commit, but it's only used in the same place its body was used before. Regarding @sbryngelson 's request, I added a case to `examples` that intentionally triggers the error reporting when run.

sbryngelson:
hi @AiredaleDev no need to actually include broken code in version control, just 'proof' or 'verification' that if you do something like this your code identifies it.

sbryngelson:
not sure I quite follow @AiredaleDev -- I know `test` is just wrapper around `run` at some level, so I think @henryleberre is saying that everything that creates a case, which would be any `run` call, including those invoked through `test`, gets its case file validated. I don't think `test` should have any special machinery for this but rather invoke it at the `run` level?

henryleberre:
> > Code that makes use of `Case` objects should have the burden of calling it as a sanity check.
> 
> @henryleberre Normally I would agree, but if I apply this sanity check to all code that constructs a `Case` (i.e. also a `TestCase`) then this line in the test runner forces the user to wait almost one whole minute before it begins. Apparently `jsonschema` is a bit slow.
> 
> The test runner will eventually call `run` anyway, which ends up reading and validating each case dictionary in parallel before launching the simulation (i.e. each concurrent run of MFC checks its own parameters then runs if they are correct.) I discovered this by adding a print statement next to the validation code, and if you'd like to check this yourself, add a print statement to the newly added `validate` method.
> 
> ```python
> # In toolchain/mfc/test/test.py
> def test():
>     # pylint: disable=global-statement, global-variable-not-assigned
>     global nFAIL
> 
>     # This is the line I have to work around.
>     # It runs before anything in parallel happens, so all 150 cases are checked sequentially.
>     cases = [ _.to_case() for _ in list_cases() ]
> 
> # In toolchain/mfc/test/case.py
>     def to_case(self) -> TestCase:
>         dictionary = self.mods.copy()
>         if self.path:
>             dictionary.update(input.load(self.path, self.args).params)
> 
>             for key, value in dictionary.items():
>                 if not isinstance(value, str):
>                     continue
> 
>                 for path in [value, os.path.join(os.path.dirname(self.path), value)]:
>                     path = os.path.abspath(path)
>                     if os.path.exists(path):
>                         dictionary[key] = path
>                         break
>         # This code "should" check the dictionary, but if we check the dictionary here
>         # we wait ~45 seconds before the test runner starts.
>         return TestCase(self.trace, dictionary, self.ppn, self.rebuild)
> 
> # In toolchain/mfc/test/test.py
> def test():
>     # ...
> 
>     sched.sched(sched.Task(..., func=handle_func, ...)
> 
> # Called by handle_func
> def _handle_func():
>    # ...
>    case.run(...) # case : TestCase
> 
> class TestCase(case.Case):
>     # ...
>     def run(...):
>         # ...
> 
>         # We just end up calling `./mfc.sh run` for each test anyway.
>         command = [
>             mfc_script, "run", filepath, *rebuild, *tasks, *case_optimization,
>             *jobs, "-t", *target_names, *gpus_select, *ARG("--")
>         ]
> 
>         return common.system(command, print_cmd=False, text=True, capture_output=True)
> ```
> 
> Regarding @sbryngelson 's request, I added a case to `examples` that intentionally triggers the error reporting.

You're right. Sorry if I was unclear. I believe we agree on this. When I said that:
> Code that makes use of `Case` objects should have the burden of calling it as a sanity check.

I did not intend to imply that code that makes use of Case objects should always call `validate`. This decoupling of validation & construction is intended to facilitate this. You only run `validate` when it is deemed reasonable / necessary. I agree we shouldn't call it in `./mfc.sh test` and just let the spawned `./mfc.sh run` processes do the validation.

AiredaleDev:
> not sure I quite follow @AiredaleDev -- I know `test` is just wrapper around `run` at some level, so I think @henryleberre is saying that everything that creates a case, which would be any `run` call, including those invoked through `test`, gets its case file validated. I don't think `test` should have any special machinery for this but rather invoke it at the `run` level?

This is how it behaves currently.

> I did not intend to imply that code that makes use of Case objects should always call validate. This decoupling of validation & construction is intended to facilitate this. You only run validate when it is deemed reasonable / necessary. I agree we shouldn't call it in ./mfc.sh test and just let the spawned ./mfc.sh run processes do the validation.

Ah, I see! I'll go in and address that real quick.

sbryngelson:
@henryleberre can approve and merge this PR when it's ready

AiredaleDev:
> hi @AiredaleDev no need to actually include broken code in version control, just 'proof' or 'verification' that if you do something like this your code identifies it.

Okay, will remove the broken case and show you the logs:

toolchain/mfc/run/case_dicts.py
```diff
@@ -1,5 +1,45 @@
...
+    'perturb_flow', 'perturb_sph', 'qbmm', 'ib', 'mapped_weno', 'mp_weno',
+    'weno_avg', 'weno_Re_flux', 'prim_vars_wrt', 'alt_soundspeed', 
+    'null_weights', 'probe_wrt', 'Monopole', 'integral_wrt', 'cu_mpi', 
+    'adap_dt', 'schlieren_wrt', 'alpha_rho_wrt', 'rho_wrt', 'mom_wrt', 
+    'vel_wrt', 'flux_wrt', 'E_wrt', 'pres_wrt', 'alpha_wrt', 'kappa_wrt',
+    'gamma_wrt', 'heat_ratio_wrt', 'pi_inf_wrt', 'pres_inf_wrt',
+    'cons_vars_wrt', 'c_wrt', 'omega_wrt', 'qm_wrt', 'vel_profile',
+    'mixture_err',
+}
+STRINGS = {"case_dir"}
```
henryleberre: @AiredaleDev I think you can just initialize these as empty now.

toolchain/mfc/run/case_dicts.py
```diff
@@ -1,5 +1,45 @@
...
+
+
+def add_param(param_name: str, stage_list: list[str], type_set: set[str]):
+    '''Arguments:
+    stage_list -- List containing parameters used for a particular stage. 
+                  One of COMMON, PRE_PROCESS, SIMULATION, POST_PROCESS.
+    type_set   -- Set containing parameters of a certain type. Currently
+                  one of INTEGERS, REALS, LOGICALS, STRINGS.'''
+    stage_list.append(param_name)
+    type_set.add(param_name)
```
henryleberre: Perhaps now is a good time to remove the `COMMON` dictionary. You can modify this function such that you can specify multiple `stage_list`s.

toolchain/mfc/run/case_dicts.py
```diff
@@ -20,175 +60,221 @@
...
+    if param in INTEGERS:
+        _properties[param] = {"type": "integer"}
+    elif param in REALS:
+        _properties[param] = {"type": "number"}
+    elif param in LOGICALS:
+        _properties[param] = {"enum": ["T", "F"]}
+    elif param in STRINGS:
+        _properties[param] = {"type": "string"}
+    else:
+        print(f'WARNING: Found parameter {param} without type!')
```
henryleberre: I think we should throw an error here instead of just a warning.

toolchain/mfc/test/test.py
```diff
@@ -99,6 +99,7 @@ def test():
 
     for case in cases:
         if case.rebuild:
+            case.validate_params()
```
henryleberre: I think you should move this to `build()`.

toolchain/mfc/test/test.py
```diff
@@ -192,6 +193,7 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
             if not os.path.exists(silo_filepath):
                 silo_filepath = os.path.join(case.get_dirpath(), 'silo_hdf5', 'p_all', 'p0', f'{t_step}.silo')
 
+            case.validate_params()
```
henryleberre: Any particular reason for checking this here?

AiredaleDev:
@henryleberre Just a reminder that this should be ready to merge!

toolchain/mfc/run/case_dicts.py
AiredaleDev: Many of these parameters are not covered by the rest of the code. I cannot just iterate through the lists as we discussed since they are sorted by "target" and not by type.

toolchain/mfc/test/test.py
AiredaleDev: None in particular -- will see if I can get away without it.

toolchain/mfc/run/case_dicts.py
AiredaleDev: I ultimately came up with another, better way of representing the types -- you now only have to type them once, resulting in less overall code -- this function is now replaced with a dictionary assignment

toolchain/mfc/run/case_dicts.py
AiredaleDev: 100% agree

toolchain/mfc/test/test.py
AiredaleDev: Moved!

toolchain/mfc/run/case_dicts.py
AiredaleDev: This actually prompted me to refactor one more time since I found a way I could easily regex-replace what I currently have into something simpler.

henryleberre (approved):
Thanks for making these changes! lgtm.

sbryngelson (approved):

PR #443, opened 2024-05-29
Revert MacOS runner - new one unstable?

sbryngelson:
This is a mock PR to test CI. Related to https://github.com/MFlowCode/MFC/issues/444 and my humble attempt to work around it for now.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -21,15 +21,15 @@ jobs:
     name: Github
     strategy:
       matrix:
-        os:    ['ubuntu', 'macos']
+        os:    ['ubuntu', 'macos-13']
         mpi:   ['mpi']
         debug: ['debug', 'no-debug']
         intel: [true, false]
         exclude:
-          - os:    macos
+          - os:    macos-13
             intel: true
 
-          - os:    macos
+          - os:    macos-13
             debug: debug
             intel: false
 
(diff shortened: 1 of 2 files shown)


sbryngelson:
I've tried the macos-12,13,14 runners and they all grab gcc13.3.0, which as far as I can tell is the only relevant change to these runners in the past few days. Strangely, gcc14 works fine with fftw on my machine, though I don't know if I ever checked gcc13.3.0 (perhaps only an older minor version).

henryleberre:
@sbryngelson I also tried to fix this but couldn't a) reproduce it locally or b) understand why this error occurs. I have a suspicion it's related to it picking the wrong system headers but I am not quite sure. This might be related to Github updating their runners, but I'm not quite sure. It did start failing when brew updated GCC from v13.2 to v13.3 but that makes no difference on my mac. I could only find one question online about someone having this error (not while building fftw) and it wasn't very helpful.

sbryngelson:
> @sbryngelson I also tried to fix this but couldn't a) reproduce it locally or b) understand why this error occurs. I have a suspicion it's related to it picking the wrong system headers but I am not quite sure. This might be related to Github updating their runners, but I'm not quite sure. It did start failing when brew updated GCC from v13.2 to v13.3 but that makes no difference on my mac. I could only find one question online about someone having this error (not while building fftw) and it wasn't very helpful.

I've had the same experience you describe. Something else must have changed.

sbryngelson:
Closing this PR in favor of https://github.com/MFlowCode/MFC/pull/469

PR #447, opened 2024-05-30
Temporary GNU 13 Silo Build Fix

henryleberre:
## Description

This hotfix gets Silo building on GNU 13. It should also pass the `-a` test suite. Let's perpetuate the tradition of hacky C code. This fix isn't pretty, but it should be correct, given the types we are working with.


sbryngelson:
approved so long as it passes CI and builds locally on gcc13+14

sbryngelson (approved):

PR #448, opened 2024-06-01
GPU-Aware MPI on OLCF Frontier and Combined weak- & strong-scaling case

### LINKED ISSUE #405: Frontier GPU-aware MPI
Support and document GPU-aware MPI on Frontier.

* If not oversubscribing the GPUs, make sure we run with `-c 7 --gpus-per-task=1 --gpu-bind=closest`. Because the NIC is connected directly to the GPU, affinity is very important.
	* It seems like we _aren't_ doing some of this per the current Frontier template: https://github.com/MFlowCode/MFC/blob/master/toolchain/templates/frontier.mako

* Since we are already building everything in, it should just be a matter of setting the runtime flags.
	* Runtime flag to set: `MPICH_GPU_SUPPORT_ENABLED=1`

henryleberre:
## Description

- Adds support for GPU-Aware MPI on OLCF Frontier. Closes #405.
- Adds a combined and configurable weak- & strong-scaling case.
- Adds scripts used to submit scaling runs on Frontier and export results.
- Modifies case argument parsing to support non-positional arguments, too.
  - Required the removal of the `-f` (`--flags`) option that became obsolete with the addition of template files for running cases and submitting jobs.
  - The new format is `./mfc.sh run <path/to/case.py> <regular mfc.sh arguments> -- <case arguments>
  - To be clear, "non-positional arguments" are arguments that start with a "-". They are called non-positional because they are identified by name, not by which argument number they are (their position in the list).

### Type of change

- [x] Modifies existing behavior

### Scope

## How Has This Been Tested?

- [x] Ran strong scaling experiments on OLCF Frontier with and without `cu_mpi`.
- [x] GitHub CI


sbryngelson:
Very good. I think we need not hold `.csv` data in version control, though everything else looks great on a quick look. Several runners are failing but I have not investigated that either.

sbryngelson:
Also, I think we need to rename `cu_mpi` $\to$ `rdma_mpi` or `gpu_mpi` or `gpudirect` or something (since `cu` was for cuda)

henryleberre:
Thanks @sbryngelson. The Frontier runner is changing because of a fix for batch execution. I'll get it working with interactive execution, too.

henryleberre:
I _think_ I fixed all the issues.

sbryngelson:
Seems like something isn't quite right. Also you can use the existing plots if you like they are fine, I am just tweaking them a bit for other purposes.

sbryngelson:
I made this figure but you don't have to use it for anything 
[strong_scaling_frontier.pdf](https://github.com/user-attachments/files/15530096/strong_scaling_frontier.pdf)
[strong_scaling_frontier.pdf](https://github.com/user-attachments/files/15530098/strong_scaling_frontier.pdf)

PR #461, opened 2024-06-09
coverage only runs when meanful files are changed (duplicates code fr…

sbryngelson:
…om `test.yml`). This removes the other `on:` commands because I couldn't figure out the right syntax.

## Description

Changed codecov to only run when meaningful files are changed.

This code probably needs improvement from @henryleberre and/or @okBrian. It duplicates paths from `test.yml` and gits rid of the `push` and `workflow_dispatch` events because I couldn't figure out the right syntax for them.

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

- [x] https://rhysd.github.io/actionlint/

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -1,6 +1,18 @@
 name: Coverage Check
 
-on: [push, pull_request, workflow_dispatch]
+on:
+  pull_request:
+    paths:
+      - '**.f90'
+      - '**.fpp'
+      - '**.py'
+      - '**.yml'
+      - 'mfc.sh'
+      - 'golden.txt'
+      - 'golden-metadata.txt'
+      - 'CMakeLists.txt'
+      - 'requirements.txt'
+
 
 jobs:
   run:


okBrian:
Instead of using GitHub built-in path, we can use something like this https://github.com/dorny/paths-filter . This should reduce possible repeated code since we don't need to use path on each pull request, workflow dispatch, and push, but instead have a single filter step.
Here's a example I created that worked on my branch:
```
name: Coverage Check

on: [push, pull_request, workflow_dispatch]

jobs:
  run:
    name: Coverage Test on CodeCov
    runs-on: "ubuntu-latest"
    steps:
      - name: Setup Ubuntu
        run: |
            sudo apt update -y
            sudo apt install -y tar wget make cmake gcc g++ python3 python3-dev "openmpi-*" libopenmpi-dev
          
      - name: Checkouts
        uses: actions/checkout@v4

      - name: Filters
        uses: dorny/paths-filter@v3
        id: filter
        with: 
          # tests/ should encompass the golden and golden-metadata.txt
          filters: |
            root:
              - 'mfc.sh'
              - 'CMakeLists.txt'
            fortran:
              - '**/*.f90'
              - '**/*.fpp'
            python:
              - '**/*.py'
              - 'toolchain/requirements.txt'
            yml:
              - '**/*.yml'
            tests:
              - 'tests/**'
        
      - name: Build
        run: /bin/bash mfc.sh build -j $(nproc) --gcov

      - name: Test
        run: /bin/bash mfc.sh test -a -j $(nproc)

      - name: Upload coverage reports to Codecov
        uses: codecov/codecov-action@v4.0.1
        with:
          token: ${{ secrets.CODECOV_TOKEN }}
```
You can even create custom filters like for documentation.

sbryngelson:
So, will this kill the job if none of the files in the filter are modified? If so, great!

I do see that this is only used in `coverage.yml`, but we would want to use the same filter in `tests.yml`. Can we do that (so as to not duplicate code)?

okBrian:
Yea, we can create a filter file so we can reuse a filter in multiple tests as seen here 
https://github.com/getsentry/sentry/blob/2ebe01feab863d89aa7564e6d243b6d80c230ddc/.github/workflows/backend.yml#L40
https://github.com/getsentry/sentry/blob/2ebe01feab863d89aa7564e6d243b6d80c230ddc/.github/file-filters.yml

sbryngelson:
perfect can you add to this PR for this?

okBrian:
Okay I'll add it to this one

okBrian:
I don't think I have the permissions to add to this pull request since I can't even fetch the changes from your fork, assuming I am doing this correctly, so I will just make another pull request when I finish.

sbryngelson:
closing in favor of #468

PR #468, opened 2024-06-10
Coverage only runs runs when meaningful files are changed

okBrian:
## Description

Added https://github.com/dorny/paths-filter to CI based on @sbryngelson Pull Request #461 
The Filter is modulated on purpose for future use as you can specify which grouping you want as seen in test.yml & coverage.yml

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

- [x] Edited the CMakeFile, then tested to make sure the filter caught the changes, and the tests would run.
- [x] Edited no files and sent changes and commented away some parts of the filter to ensure the test would be skipped. 

**Test Configuration**:

Github Actions on my fork

## Checklist

### DIFF
--- a/.github/file-filter.yml
+++ b/.github/file-filter.yml
@@ -27,7 +27,7 @@ yml: &yml
 checkall: &checkall
   - *fortran_src
   - *python_src
-  # - *cmakelist
+  - *cmakelist
   - *tests
   - *scripts
-  # - *yml
+  - *yml
--- a/.github/file-filter.yml
+++ b/.github/file-filter.yml
@@ -1,0 +1,18 @@
+# This is used by the action https://github.com/dorny/paths-filter
+# Lists all files to be filtered for changes to run a CI
+
+checkall: &checkall
+  - 'mfc.sh'
+  - 'CMakeLists.txt'
+  - '**/*.f90'
+  - '**/*.fpp'
+  - '**/*.py'
+  - 'toolchain/requirements.txt'
+  - '**/*.yml'
+  - 'tests/**'
+
+coverage: &coverage
+  - *checkall
+
+test: &test
+  - *checkall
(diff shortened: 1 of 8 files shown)


sbryngelson:
can be reviewed and merged by @henryleberre

sbryngelson (approved):

henryleberre:
Is there a reason you can't use the `paths:` feature like we do for `on: [push]` ?

okBrian:
The main reason I suggested using https://github.com/dorny/paths-filter was because the path applied to pull request, push, and workflow_dispatch. 
I don't think their is a way to do something like this
```
on: 
  [push, pull_request, workflow_dispatch]
      paths:
      - '**.f90'
      - '**.fpp'
      - '**.py'
      - '**.yml'
      - 'mfc.sh'
      - 'golden.txt'
      - 'CMakeLists.txt'
      - 'requirements.txt'
```
If we can't do something similar to above we are repeating a lot of code, which is something Spencer was trying to change in his pr I think.
Also since the path filter can read from file all CI can share the same file (like coverage and test) and just filter what it needs.

sbryngelson:
indeed the above syntax doesn't work, though I do notice that you have to repeat your filtering statement in all of the workflows in `tests.yml` ...

okBrian:
This is true, but in the future if we have different jobs in the same files that want to check different files we can specify that it in the conditional like in this example: https://github.com/GoogleChrome/web.dev/blob/3a57b721e7df6fc52172f676ca68d16153bda6a3/.github/workflows/lint-workflow.yml#L26 . This feature might be useful if we add more complex tests.

henryleberre:
I was assuming that Github supported YAML anchors.. but they don't. https://github.com/actions/runner/issues/1182 has been open since 2021 and has 1k upvotes. The last comment in that thread gives an example of how I was imagining using it.

PR #469, opened 2024-06-12
Fixed MacOS CI with no-debug flag only

### LINKED ISSUE #444: FFTW Broken on MacOS runner CI
FFTW Broken on MacOS runners

See: https://github.com/MFlowCode/MFC/actions/runs/9290749453/job/25567695171?pr=443

This is new, I suspect has something to do with updated MacOS runners.

Both of the runners are new, and they may have changed the meaning of `macos-latest`
* https://github.com/actions/runner-images/blob/main/images/macos/macos-14-Readme.md
* https://github.com/actions/runner-images/blob/main/images/macos/macos-14-arm64-Readme.md

See the table here:
https://github.com/actions/runner-images

and the updated MacOS runners here, released just as this change happened:
https://github.com/actions/runner-images/releases

Notably, it seems that they upgraded the runners to GCC13.3.0 from 13.2.0. Perhaps this is the issue.

Update:

I've tried the macos-12,13,14 runners and they all grab gcc13.3.0 and fail, which as far as I can tell is the only relevant change to these runners in the past few days. Strangely, gcc14 works fine with fftw on my machine, though I don't know if I ever checked gcc13.3.0 (perhaps only an older minor version).

okBrian:
## Description

Fixed the MacOS CI specifically with the no-debug flag
Unsure whats wrong with the debug one

Fixes #444 

### Type of change

### Scope

## How Has This Been Tested?

- [x] check if MacOS --no-debug would build w/o my lines
- [x] check if MacOS --no-debug would build with my lines

**Test Configuration**:

GitHub Runners

## Checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,8 +49,8 @@ jobs:
           echo "CC=gcc-14"      >> $GITHUB_ENV
           echo "CXX=g++-14"     >> $GITHUB_ENV
           echo "FC=gfortran-14" >> $GITHUB_ENV
-          export CPATH=/opt/homebrew/include
-          export LIBRARY_PATH=/opt/homebrew/lib
+          export CPATH="/opt/homebrew/include":$CPATH
+          export LIBRARY_PATH="/opt/homebrew/lib":$LIBRARY_PATH
           
       - name: (MacOS) Build OpenMPI
         if:   matrix.os == 'macos' && matrix.mpi == 'mpi'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -49,6 +49,8 @@ jobs:
           echo "CC=gcc-14"      >> $GITHUB_ENV
           echo "CXX=g++-14"     >> $GITHUB_ENV
           echo "FC=gfortran-14" >> $GITHUB_ENV
+          export CPATH=/opt/homebrew/include
+          export LIBRARY_PATH=/opt/homebrew/lib
           
       - name: (MacOS) Build OpenMPI
         if:   matrix.os == 'macos' && matrix.mpi == 'mpi'


sbryngelson:
hmm. do you want to prepend the existing paths instead of replacing them?
```bash
	export CPATH=/opt/homebrew/include
	export LIBRARY_PATH=/opt/homebrew/lib
```
to
`export CPATH="/opt/homebrew/include":$CPATH`

This might fix the debug issue.

(I may be messing up the exact syntax).

sbryngelson:
There's also the "recommendations" here:

https://apple.stackexchange.com/questions/40704/homebrew-installed-libraries-how-do-i-use-them

okBrian:
I did some more experimentation here's what I figured out.
* mpich is not the issue
* gcc 13.3 doesn't build with my changes at all
* gcc 14 builds correctly with my changes but requires the pre-release: 20240611.1 MacOS image
  * There is no current support in changing or choosing a pre-release image 
* gcc 14 doesn't build w/o my changes
*  brew auto links libraries with brew link and unlink so its unlikely that's the cause
Heres the branch I tested everything on https://github.com/okBrian/MFC/tree/mac

sbryngelson:
> I did some more experimentation here's what I figured out.
> 
> * mpich is not the issue
> * gcc 13.3 doesn't build with my changes at all
> * gcc 14 builds correctly with my changes but requires the pre-release: 20240611.1 MacOS image
>   
>   * There is no current support in changing or choosing a pre-release image
> * gcc 14 doesn't build w/o my changes
> * brew auto links libraries with brew link and unlink so its unlikely that's the cause
>   Heres the branch I tested everything on [okBrian/MFC@`mac`](https://github.com/okBrian/MFC/tree/mac?rgh-link-date=2024-06-12T19%3A34%3A09Z)

This is all very believable to me. I guess this is a draft PR to be merged when the MacOS runner is updated? Do we have an approximate known date for that? What about `--debug`?

okBrian:
--debug and no-debug have no affect on the results since --no-debug worked on my actions and debug worked on this repo. I also tested this in my mac branch. The only difference between the test that failed and passed I could find was the MacOS Image.

sbryngelson:
Understood. Do we have timeline for new MacOS image or how to specify it?

okBrian:
From what I searched you can't switch to a pre-release image specifically. For the release data/timeline I'll make a post here https://github.com/actions/runner-images/discussions since I can't find it anywhere.

okBrian:
Before I asked in discussion I just read a bit more in their README.
"Usually, image deployment takes 2-3 days, and documentation in the main branch is only updated when deployment is finished."
The 20240611.1 has the tag "Deploying" in their README so I think in 2-3 days it should be up and running over the previous image.

okBrian:
Hey, the MacOS Image is fully deployed so assuming the test pass this should be it.

PR #481, opened 2024-06-22
Add WENO5 Variants

ChrisZYJ:
## Description

This PR adds WENO-Z (Borges, et al., 2008) and TENO (Fu et al., 2016) options on top of the existing WENO-JS and WENO-M schemes. WENO-Z is less dissipative and less costly than WENO-M, and is much less dissipative than WENO-JS while being similar in speed. TENO has even better spectral properties than WENO-Z, but can be less robust in extreme scenerios. 

Please refer the the updates documentation for more details. 

### Type of change

### Scope

## How Has This Been Tested?

- [x] 1D Shu-Osher Comparison at Nx=200
The 1D Shu-Osher test results are compared against those of Fu et al. (2016). Note that $C_T = 10^{-6}$ is used for TENO in our test as opposed to $C_T = 10^{-5}$ stated in the paper. The 'exact' solution is calculated using WENO-JS at Nx=2000, as suggested by Fu et al. (2016). The case files are included in the updated examples folder. 

1D_shu_osher_comparison
1D_shu_osher_comparison_zoomed
**Fu et al. (2016)**
image

- [x] 1D Shu-Osher Comparison at Nx=1000
The same test but at higher resolution to show convergence. 

1D_shu_osher_comparison_1000
1D_shu_osher_comparison_1000_zoomed

- [x] 2D Riemann Problem
The case files are the same as those in the example folder, with WENO-JS changed to WENO-M, WENO-Z, and TENO. The results show slight differences due to the chaotic nature of the simulation, and the better spectral properties are evident in the more prominent swirls at the same resolution (TENO > WENO-Z > WENO-M > WENO-JS).

**WENO-JS** (on MFC Website - Example Cases)
2D_riemann_test_wenojs
**WENO-M**
2D_riemann_test_wenom
**WENO-Z**
2D_riemann_test_wenoz
**TENO**
2D_riemann_test_teno

- [x] 2D Shock-Droplet Interaction
The case files are the same as those in the example folder, with WENO-M changed to WENO-Z and TENO respectively. The results are simliar to each other and to the original result using WENO-M on the MFC website within expectations. 

**WENO-M** (on MFC Website - Example Cases)
2D_shockdroplet_WENOM_2
**WENO-Z**
2D_shockdroplet_WENOZ_2
**TENO**
2D_shockdroplet_TENO

**Test Configuration**:

  * 1D tests: Ubuntu 22.04.4 LTS; GNU v11.4.0
  * 2D tests: Richardson

## Checklist

- [x] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled (Note: not for AMD hardware)

- [x] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this PR [nsys_report_wenoz.txt](https://github.com/user-attachments/files/15934366/nsys_report_wenoz.txt)

**16M points**
GPU_performance

## Comments
* New tests are generated for WENO-Z and TENO, for weno_order=3,5 and dim=1,2,3 where applicable. The golden files for WENO-M and MP_WENO are regenerated due to an improvement in how cases files are generated (the parameters are now optional), but all tests are performed before and after the regeneration to ensure that the results match. All tests before and after are passed on CPU (locally on Ubuntu and on Richardson), and on GPU (Delta).
* The current implementation for WENO-M (if statement within WENO-JS) is improved using case-optimization, which is necessary for adding new WENO variants without affecting the performance. The attached performance report shows that the performance of the new code is virtually identical to the original code for WENO-JS and WENO-M, and marginally better at 8 GPUs. The performance of WENO-Z and TENO matches the expectations according to the theory and other papers (WENO-JS ~ WENO-Z > TENO >> WENO-M). 
* I only have access to Delta, which has Nvidia GPUs, but the changes to the GPU code are minimal and have very simple logic. Extensive tests have been performed on those GPUs to ensure the results are correct and the performance is as expected.


sbryngelson:
This is an example of a great PR! Good use of added tests, toolchain involvement, proof that it works as intended, and only the required minimal code additions. Will review it a bit more thoroughly tomorrow.

src/simulation/m_global_parameters.fpp
```diff
@@ -487,7 +495,7 @@ contains
         mpp_lim = .false.
         time_stepper = dflt_int
         weno_eps = dflt_real
-        mapped_weno = .false.
+        teno_CT = 1d-6
```
sbryngelson: the default value should be `dflt_real` so that the code crashes if the user doesn't set this parameter but tries to use `teno`.

ChrisZYJ: My intention is for the code to run with teno without explicitly setting teno_CT, since the default value is most commonly used. However, I'm open to changing it if you believe it would improve consistency with the rest of the codebase.

sbryngelson: right now standard MFC practice is to require the user to set the variables that impact the method. for example, `weno_eps` should be required to be supplied by the user for "normal WENO." 

What you would do is have a statement that says something in `m_checker`. So we probably have something like 
```
if (weno_eps == dflt_real and weno_order > 1) then 
	print "weno is used, but weno_eps is not set. aborting"
	stop
end if
```

and you would have something very similar for teno.

sbryngelson: note that checking for `==` on a float is bad programming practice, but i think we still do it (to my knowledge). if we do, do you mind opening a github issue on the topic? something like "replace checks that see if two float variables are equal via `==` with something resiliant to floating point errors/roundoff".

ChrisZYJ: I see. I'll take the appropriate changes. Thanks!

ChrisZYJ: Yep I'll create the == issue

sbryngelson: > right now standard MFC practice is to require the user to set the variables that impact the method. for example, `weno_eps` should be required to be supplied by the user for "normal WENO."
> 
> What you would do is have a statement that says something in `m_checker`. So we probably have something like
> 
> ```
> if (weno_eps == dflt_real and weno_order > 1) then 
> 	print "weno is used, but weno_eps is not set. aborting"
> 	stop
> end if
> ```
> 
> and you would have something very similar for teno.

Actually this should say 
`print "weno is used, but weno_eps is not set. a typical value of weno_eps is 1e-6. aborting"`

PR #483, opened 2024-06-22
Pre_Proccess, Added Intent, Comments, and Formatting

### LINKED ISSUE #478: Pre process

okBrian:
## Description

Added Intent to all subroutine arguments that didn't already have it
Added Missing @ param comments on subroutines
Formatted Code to be a bit more uniform inline with my how I formatted the Common Directory

Fixes #478 

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

Tested On Github Actions, and Tested Locally

**Test Configuration**:

Ubuntu 22.04.4 LTS, gcc14
Github Actions

## Checklist

### DIFF
--- a/src/pre_process/m_checker.f90
+++ b/src/pre_process/m_checker.f90
@@ -2,8 +2,6 @@
 !!@file m_checker.f90
 !!@brief Contains module m_checker
 
-#define UNSUPPORTED_MESSAGE(f, s) "Unsupported combination of values of " + f + " and " + s + "Exiting ..."
-
 !> @brief The purpose of the module is to check for compatible input files
 module m_checker
 
--- a/src/pre_process/m_checker.f90
+++ b/src/pre_process/m_checker.f90
@@ -19,7 +19,7 @@ module m_checker
 
 contains
 
-    subroutine s_check_inputs()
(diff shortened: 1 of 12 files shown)


src/pre_process/m_checker.f90
```diff
@@ -2,6 +2,8 @@
 !!@file m_checker.f90
 !!@brief Contains module m_checker
 
+#define UNSUPPORTED_MESSAGE(f, s) "Unsupported combination of values of " + f + " and " + s + "Exiting ..."
```
sbryngelson: how is this used?

sbryngelson:
looks good. i just found one minor thing to comment on. 

on reviewing this PR I'm reminded that i find our use of `!--------------` comments after each subroutine statement or function rather silly and a waste of text/space.

Could you do a grep/macro/whatever to delete all of these in a new PR?

src/pre_process/m_checker.f90
okBrian: I was testing something I must've forgotten to remove it.

okBrian:
I can remove all the `!--------------------` in a new pr.

PR #497, opened 2024-07-01
Fix Benchmark & Add Monopole Checks & Add Monopole Tests

ChrisZYJ:
## Description

1. Fixes issue #396 by adding missing parameter in `viscous_weno5_sgb_mono` case file. 
2. Add monopole checks and make some parameters compulsory
3. New tests for monopole delay and all monopole supports (except for 6)

### Explaination

The original benchmark `viscous_weno5_sgb_mono` did not specify the monopole support type, so it defaulted to 1. However, support = 1 is not valid for 3D simulations and triggered none of the if-statements in the `f_delta` function within the monopole module, resulting in `f_delta` being undefined. I'm not entirely sure why it only started failing on Frontier and Phoenix in my PR #495, but it's probably due to the floating-point numbers being checked differently.

The modified benchmark file should simulate what it intends to simulate, with the computational domain and parameters remaining the same. The only changes are adding `Mono(1)%support' : 4` and a rotation of the domain to align the z-axis with the requirements of monopole support 4.

Now, I've made monopole support mandatory through checks (along with other monopole parameters) to prevent this from happening again. I've also added test cases for monopoles as requested (support 6 for cylindrical coordinates cannot be run, so no check is implemented).

### Type of change

### Scope

## How Has This Been Tested?

- [x] Passes all tests locally
- [x] Passed all tests on Delta
- [x] Benchmark runs successfully on Delta
- [x] Benchmark runs successfully on Delta after merging with #495

**Test Configuration**:

  * Ubuntu 22.04.4 LTS; GNU v11.4.0
  * Delta

## Checklist

- [x] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled (not for AMD hardware)

Performance tests are not done as the changes are strictly in the parameter checking and initialization stages, and only involves assigning values to variables.


src/simulation/m_monopole.fpp
```diff
@@ -69,19 +69,39 @@ contains
...
+                dir(i) = 1d0
+            else
+                dir(i) = mono(i)%dir
+            end if
+            if (f_is_default(mono(i)%delay)) then
+                delay(i) = 0d0
+            else
+                delay(i) = mono(i)%delay
+            end if
+            if (f_is_default(mono(i)%delay)) then
```
sbryngelson: typo here, should be `if (f_is_default(mono(i)%support_width)) then` I presume

sbryngelson:
Good catch!

src/simulation/m_monopole.fpp
ChrisZYJ: Sorry for the typo! Just fixed it

PR #502, opened 2024-07-03
Fix 1D_impact, 1D_qbmm, and 2D_5wave_quasi1D

okBrian:
## Description

Fixed examples 1D_impact, 1D_qbmm, and 2D_5wave_quasi1D

### Type of change

### Scope

## How Has This Been Tested?

- [x] Tested on my Local Machine on 1 Node

**Test Configuration**:

Ubuntu 22.04.4 LTS, gcc14

## Checklist

### DIFF
--- a/examples/1D_impact/case.py
+++ b/examples/1D_impact/case.py
@@ -13,7 +13,7 @@ mydt    = Tend/(1.*Nt)
 # Configuring case dictionary
 print(json.dumps({
                     # Logistics ================================================
-                    'case_dir'                     : '\'.\'',                   
+                    # 'case_dir'                     : '\'.\'',                   
                     'run_time_info'                : 'T',                       
                     # ==========================================================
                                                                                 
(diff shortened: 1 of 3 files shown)


okBrian:
I am not entirely sure these have affected what the examples are actually trying to show. These are just the changes the examples needed for it to work on my local machine.

sbryngelson (approved):

sbryngelson:
This looks right to me. I think `case_dir` or whatever is a deprecated input variable. You could check to see if it gets used anywhere in the actual code. Can you remove it from the toolchain (and any remaining input files) if it doesn't? @okBrian

sbryngelson:
Also delete, not comment, the `case_dir` lines

sbryngelson:
@henryleberre how does the code know what the `case_dir` is? it looks like if it needs to be specified in the `case.py` file and passed into `NAMELIST` in the source, but I don't see anywhere in the toolchain where it's auto-generating the `case_dir` parameter/value or even setting a default? I must be missing something because everything is working as intended, but still curious.

Also, if `case_dir` is not to be specified in `case.py`, then it should be removed from `CASE_DICTS`, right?

PR #508, opened 2024-07-05
Documentation typos

ChrisZYJ:
## Description

All `%`'s are missing on the Case Files page. Fixed by replacing `%` within literal backtick expressions by `%%`. Also fixed some other typos / formatting. 

### Type of change

### Scope

### DIFF
--- a/docs/documentation/case.md
+++ b/docs/documentation/case.md
@@ -100,7 +100,7 @@ feature, detecting GPU pointers and performing RDMA accordingly.
 
 | Parameter                | Type    | Description                      |
 | ---:                     | :----:  |          :---                    |
-| `x[y,z]_domain%beg[end]` | Real    | Beginning [ending] of the $x$[y,z]-direction domain    |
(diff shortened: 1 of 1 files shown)


sbryngelson:
Thanks @ChrisZYJ! It took me way to long to figure this out, but the solution is to use two percent signs to get one real one. x%%y in a literal backtick expression = `x%y` on the website

ChrisZYJ:
Thanks! I would never have thought of that on my own

sbryngelson:
> Thanks! I would never have thought of that on my own

Me either! Some careful Googling found the trick for me. You can build the documentation on your local machine via `./mfc.sh build -t documentation` to try many different things or just check things are working as expected.

sbryngelson (approved):

ChrisZYJ:
Thanks for the tips on how to build the pages locally - that makes debugging much easier. 

Just found a few more issues: while `a%%b` -> `a%b` works, `a%%b[x]` -> `a%%b[x]` introduces an extra `%` instead. Same for `a%%` -> `a%%`. Fixing that now

PR #514, opened 2024-07-09
Simplify `./mfc.sh clean` to behave as most people expect.

### LINKED ISSUE #503: `./mfc.sh clean` no longer works, change to do the obvious thing
running `./mfc.sh clean` just triggers an error on my end (see below). Also, `./mfc.sh --help` triggers are not useful help message. It seems like it doesn't work as intended anymore. I recommend replacing it with a function that deletes `build/` and call it a day.

```console
shb-m1pro-3: Downloads/mfc $ ./mfc.sh clean
mfc: OK > (venv) Entered the Python 3.12.3 virtual environment (>= 3.8).

      .=++*:          -+*+=.        | spencer@shb-m1pro-3.local [Darwin]
     :+   -*-        ==   =* .      | ----------------------------------
  :*##-.....:*+   .#%+++=--+=:::.   | --jobs 1
  -=-++-======#=--**+++==+*++=::-:. | --mpi --no-gpu --no-debug --no-gcov --no-unified
 .:++=----------====+*= ==..:%..... | --targets pre_process, simulation, and post_process

╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│ /Users/spencer/Downloads/mfc/toolchain/main.py:66 in <module>                                    │
│                                                                                                  │
│   63 │   │                                                                                       │
│   64 │   │   __print_greeting()                                                                  │
│   65 │   │   __checks()                                                                          │
│ ❱ 66 │   │   __run()                                                                             │
│   67 │                                                                                           │
│   68 │   except MFCException as exc:                                                             │
│   69 │   │   cons.reset()                                                                        │
│                                                                                                  │
│ /Users/spencer/Downloads/mfc/toolchain/main.py:51 in __run                                       │
│                                                                                                  │
│   48                                                                                             │
│   49                                                                                             │
│   50 def __run():                                                                                │
│ ❱ 51 │   {"test":   test.test,     "run":        run.run,          "build":      build.build,    │
│   52 │    "clean":  build.clean,   "bench":      bench.bench,      "count":      count.count,    │
│   53 │    "packer": packer.packer, "count_diff": count.count_diff, "bench_diff": bench.diff      │
│   54 │   }[ARG("command")]()                                                                     │
│                                                                                                  │
│ /Users/spencer/Downloads/mfc/toolchain/mfc/build.py:307 in clean                                 │
│                                                                                                  │
│   304                                                                                            │
│   305 def clean(targets = None, case: input.MFCInputFile = None):                                │
│   306 │   targets = get_targets(list(REQUIRED_TARGETS) + (targets or ARG("targets")))            │
│ ❱ 307 │   case    = case or input.load(ARG("input"), ARG("arguments"), {})                       │
│   308 │                                                                                          │
│   309 │   cons.print(__generate_header("Clean", targets))                                        │
│   310 │   cons.print(no_indent=True)                                                             │
│                                                                                                  │
│ /Users/spencer/Downloads/mfc/toolchain/mfc/state.py:62 in ARG                                    │
│                                                                                                  │
│   59 │   if dflt is not None:                                                                    │
│   60 │   │   return dflt                                                                         │
│   61 │                                                                                           │
│ ❱ 62 │   raise KeyError(f"{arg} is not an argument.")                                            │
│   63                                                                                             │
│   64 def ARGS() -> dict:                                                                         │
│   65 │   # pylint: disable=global-variable-not-assigned                                          │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
KeyError: 'arguments is not an argument.'


ERROR: An unexpected exception occurred: 'arguments is not an argument.'

./mfc.sh: line 49: 17445 Terminated: 15          python3 "$(pwd)/toolchain/main.py" "$@"

mfc: ERROR > mfc.py finished with a 143 exit code.
```

AiredaleDev:
## Description

Fixes #503

`./mfc.sh clean` is functional once again. It now behaves like most other build scripts' `clean` command -- delete all artifacts.

### Type of change

- [ ] Something else

### DIFF
--- a/mfc.sh
+++ b/mfc.sh
@@ -36,6 +36,8 @@ elif [ "$1" '==' "docker" ]; then
     shift; . "$(pwd)/toolchain/bootstrap/docker.sh"  $@; exit 0
 elif [ "$1" '==' "venv" ]; then
     shift; . "$(pwd)/toolchain/bootstrap/python.sh"  $@; return
+elif [ "$1" '==' "clean" ]; then
+    rm -rf "$(pwd)/build"; exit 0
 fi
 
 mkdir -p "$(pwd)/build"
--- a/mfc.sh
+++ b/mfc.sh
@@ -36,8 +36,6 @@ elif [ "$1" '==' "docker" ]; then
     shift; . "$(pwd)/toolchain/bootstrap/docker.sh"  $@; exit 0
 elif [ "$1" '==' "venv" ]; then
     shift; . "$(pwd)/toolchain/bootstrap/python.sh"  $@; return
-elif [ "$1" '==' "clean" ]; then
-    rm -rf "$(pwd)/build"; exit 0
 fi
 
 mkdir -p "$(pwd)/build"
--- a/toolchain/main.py
+++ b/toolchain/main.py
@@ -48,9 +48,9 @@ def __checks():
 
 
 def __run():
-    {"test":       test.test,        "run":        run.run,          "build":  build.build,
-     "bench":      bench.bench,      "count":      count.count,      "packer": packer.packer, 
-     "count_diff": count.count_diff, "bench_diff": bench.diff
+    {"test":   test.test,     "run":        run.run,          "build":      build.build,
+     "clean":  build.clean,   "bench":      bench.bench,      "count":      count.count,
+     "packer": packer.packer, "count_diff": count.count_diff, "bench_diff": bench.diff
     }[ARG("command")]()
 
 
--- a/toolchain/main.py
+++ b/toolchain/main.py
@@ -48,9 +48,9 @@ def __checks():
 
 
 def __run():
-    {"test":   test.test,     "run":        run.run,          "build":      build.build,
-     "clean":  build.clean,   "bench":      bench.bench,      "count":      count.count,
-     "packer": packer.packer, "count_diff": count.count_diff, "bench_diff": bench.diff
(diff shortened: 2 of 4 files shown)


sbryngelson (approved):

henryleberre (changes requested):
Thanks for fixing the `ARG("arguments")` bug. I see where you two are coming from here and I do think this is a worthwhile change. Notably, because the use-cases where `./mfc.sh clean` would actually be useful were incredibly rare (compiler bugs, missing libraries, etc.). Most of the time, you just want to configure CMake again from scratch.

I would see the value in adding some additional functionality to this command, where it doesn't remove **all** staging directories, but only those being used for the current build configuration and/or specified targets and case.

However, it's probably not worth our time right now, considering I don't even use that command in its current state. I always `rm -rf build` when in doubt (exactly what you are doing here!). For a minor compromise in flexibility, we gain a lot in ease-of-use and I am all for it.

The way I see it now is that, should your build environment not change, you should never have to call `./mfc.sh clean`. In the event it does, restart from scratch.

To fully remove the remnants from the current `./mfc.sh clean` implementation, you should, 
- In args.py, remove all the optional flags/arguments for the clean command. Running `./mfc.sh clean -h` should be practically empty.
- Remove the `clean()` methods from build.py (and elsewhere if applicable).
- Instead of hardcoding `build` as a relative path, you should use `MFC_BUILD_DIR` from common.py.

sbryngelson:
Will merge once tests pass.

PR #515, opened 2024-07-09
Add CFL Based Adaptive Time-Stepping

### LINKED ISSUE #298: Adaptive time stepping with embedded Runge-Kutta and/or CFL
Adaptive time stepping with embedded Runge-Kutta. RK23 most likely - computing RHSs for RK3 then using them to compute the update for RK2 and RK3 (2nd and 3rd order accurate versions of RK), then the difference between them is a measure of time step error. Involves a tuning parameter that has to be added to the inputs.

wilfonba:
## Description

This PR adds constant CFL time-stepping for all physics. The new feature is compatible with the existing ability to restart cases with additional patches.

Fixes #298 

### Type of change

### Scope

## How Has This Been Tested?

- [x] Test A: This test is a 1D sodshock tube ran with gfortran 14.1.0 on an Apple M2 processor. The figure shows the results of a constant CFL simulation overlayed over a constant dt simulation.
asdf
- [x] Test B: This test is a 2D Riemann problem ran with gfortran 14.1.0 on an Apple M2 procesor. The video shows the constant CFL result on the left and the constant dt simulation on the right. Some minor differences are expected.
https://github.com/MFlowCode/MFC/assets/48168887/95bc7182-f013-49c5-8951-960ed68574ea
- [x] Test C: This test shows a simulation that is restarted with a new patch. It was ran with gfortran 14.1.0 on an Apple M2 processor.
https://github.com/MFlowCode/MFC/assets/48168887/fad28d4f-bec4-455e-a6a8-795db5724f36
- [x] Test D: This test  is a 3D shock droplet ran with NVHPC 22.11 on Pace Phoenix with Nvidia V100 GPUs. The left panel shows the constant CFL result and the right panel shows the constant dt result.
https://github.com/MFlowCode/MFC/assets/48168887/882cf2bc-a5dc-4ef5-a25e-400c3ac3016d

## Checklist

- [x] Ran the code on MI200+ GPUs and ensure the new features performed as expected (the GPU results match the CPU results)
- [x] Compared ns/gp/eq/rhs
       V100 - cfl_adap_dt (3.320) constant dt (3.311) 8M Cells on 1 V100
       M2 - cfl_adap_dt (66.645) constant dt (65.657) 1M Cells on 1 core


src/common/m_checker_common.fpp
```diff
@@ -58,14 +58,18 @@ contains
...
-        if (t_step_start < 0) then
-            call s_mpi_abort('t_step_start must be non-negative. Exiting ...')
-        elseif (t_step_stop <= t_step_start) then
-            call s_mpi_abort('t_step_stop must be greater than t_step_start. '// &
-                             'Exiting ...')
-        elseif (t_step_save > t_step_stop - t_step_start) then
-            call s_mpi_abort('t_step_save must be less or equal to '// &
-                             '(t_step_stop - t_step_start). Exiting ...')
+        if (cfl_dt) then
+
```
sbryngelson: `if (.not. cfl_dt) then`

src/simulation/m_global_parameters.fpp
```diff
@@ -486,11 +497,18 @@ contains
 
         dt = dflt_real
 
+        cfl_dt = .false.
+        cfl = dflt_real
+
         t_step_start = dflt_int
         t_step_stop = dflt_int
         t_step_save = dflt_int
         t_step_print = 1
```
sbryngelson, line 507: `dflt_int`?

docs/documentation/case.md
```diff
@@ -372,6 +358,16 @@ Details of implementation of viscosity in MFC can be found in [Coralic (2015)](r
 | `wave_speeds`          | Integer | Wave-speed estimation: [1] Direct (Batten et al. 1997); [2] Pressure-velocity* (Toro 1999)	 |
 | `weno_Re_flux`         | Logical | Compute velocity gradient using scaler divergence theorem	 |
 | `weno_avg`          	 | Logical | Arithmetic mean of left and right, WENO-reconstructed, cell-boundary values |
+| `dt`                   | Real    | Time step size |
+| `t_step_start`         | Integer | Simulation starting time step |
+| `t_step_stop`          | Integer | Simulation stopping time step |
+| `t_step_save`          | Integer | Frequency to output data |
+| `t_step_print`         | Integer | Frequency to print the current step number to standard output (default 1) |
+| `cfl_dt`               | Logical | CFL based adaptive time-stepping |
+| `cfl`                  | Real    | Specified CFL value |
```
sbryngelson: maybe `cfl_target`? otherwise, this could be confused with the CFL number for the initial condition (having the code compute `dt` from `dx/dy/dz` and the speed of sound).

docs/documentation/case.md
```diff
@@ -372,6 +358,16 @@ Details of implementation of viscosity in MFC can be found in [Coralic (2015)](r
...
+| `n_start`              | Integer | Simulation starting save file |
```
sbryngelson: what does `Simulation starting save file` mean?

docs/documentation/case.md
```diff
@@ -372,6 +358,16 @@ Details of implementation of viscosity in MFC can be found in [Coralic (2015)](r
...
+| `t_save`               | Integer | Time duration between data output |
```
sbryngelson: this isn't a duration but a number of steps if it's an integer...

src/simulation/m_time_steppers.fpp
```diff
@@ -915,6 +938,133 @@ contains
 
     end subroutine s_adaptive_dt_bubble
 
+    subroutine s_compute_dt()
```
sbryngelson, line 915: should this not call a separate subroutine called `s_compute_cfl()`? if so, then we could add a feature where instead of specifying the time step size you specify the CFL and it computes step size for you (even without using adaptive time step sizes).

src/simulation/m_global_parameters.fpp
wilfonba: The CFL is between 0 and 1 no?

src/simulation/m_time_steppers.fpp
wilfonba: That's exactly what this subroutine does. Calculates a time-step given a cfl number

wilfonba: This can be accomplished by calling this subroutine once at the beginning of the simulation instead of at every time-step, but yes.

src/common/m_checker_common.fpp
wilfonba: There are supposed to be user input checks here. Oops

src/simulation/m_global_parameters.fpp
wilfonba: Stable CFL numbers are between 0 and 1

src/simulation/m_time_steppers.fpp
wilfonba: This can be accomplished by calling this subroutine once at the first time step. I'll add it.

docs/documentation/case.md
wilfonba: Does "Save file from which to start simulation" make sense.

sbryngelson: so it's for restart purposes? if so, that should go in the description. and if so, is there a reason this is different from the other restart parameters/documentation we have?

src/simulation/m_time_steppers.fpp
sbryngelson: indeed

docs/documentation/case.md
wilfonba: `t_step_start` isn't compatible with adaptive time-stepping because there's no explicit connection between # of time-steps and when the data is dumped. As it stands, `n_start` serves the same purpose as `t_step_start`, but is compatible with adaptive time-stepping. The documentation for restarting case files addresses this.

sbryngelson:
@wilfonba this codecov thing is amazing. It shows what lines you added that aren't covered in the test suite! Not that you have to cover 100% of the new lines, but you might be able to see what cases aren't covered that you could easily add.

wilfonba:
Yeah. I'm sure I can up the coverage a decent amount without too many additional tests. Apparently I need to fix GPUs too

sbryngelson:
Can you run an `./mfc.sh test -l` and copy/paste the new test cases here so I know what they are?

wilfonba:
```
  0501B3DA   1D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
  AE02324F   1D -> Bodyforces -> cfl_adap_dt=T                                      
  D80F2162   2D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
  121D4ECA   2D -> Bodyforces -> cfl_adap_dt=T                                      
  128954AD   3D -> Cylindrical -> cfl_adap_dt=T                                     
  07C54EDD   3D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
  B54BB9D8   3D -> Bodyforces -> cfl_adap_dt=T 
  37FA2CEF   1D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T  
  6B1AD553   2D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T  
  53A15FFC   3D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T
```

wilfonba:
This list was created with #518 . I figured it would be more clear this way.

sbryngelson:
> ```
>   0501B3DA   1D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
>   AE02324F   1D -> Bodyforces -> cfl_adap_dt=T                                      
>   D80F2162   2D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
>   121D4ECA   2D -> Bodyforces -> cfl_adap_dt=T                                      
>   128954AD   3D -> Cylindrical -> cfl_adap_dt=T                                     
>   07C54EDD   3D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_adap_dt=T  
>   B54BB9D8   3D -> Bodyforces -> cfl_adap_dt=T 
>   37FA2CEF   1D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T  
>   6B1AD553   2D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T  
>   53A15FFC   3D -> Viscosity -> Bubbles -> QBMM -> bubble_model=3 -> cfl_const_dt=T
> ```

Perfect, thanks

sbryngelson:
@wilfonba is this ready for review/merge?

src/simulation/m_time_steppers.fpp
```diff
@@ -892,6 +910,127 @@ contains
...
+                    vel_sum = 0d0
+                    do i = 1, num_dims
+                        vel_sum = vel_sum + vel(i)**2d0
+                    end do
+
+                    pres = q_prim_vf(E_idx)%sf(j, k, l)
+
+                    E = gamma*pres + pi_inf + 5d-1*rho*vel_sum + qv
+
+                    H = (E + pres)/rho
```
sbryngelson: this only works for the 5eqn model without bubbles I believe. perhaps you should be using some other already-existing subroutine to get these quantities? (or creating your own that can be used elsewhere, this computation is repeated several times, e.g., in Riemann solvers and m_data_output).

src/simulation/m_time_steppers.fpp
```diff
@@ -892,6 +910,127 @@ contains
...
+
+                    ! Compute mixture sound speed
+                    call s_compute_speed_of_sound(pres, rho, gamma, pi_inf, H, alpha, vel_sum, c)
+
+                    if (grid_geometry == 3) then
+                        if (k == 0) then
```
sbryngelson: i suspect this logic already exists somewhere else in the code?

src/simulation/m_time_steppers.fpp
```diff
@@ -892,6 +910,127 @@ contains
...
+                        end if
+                    end if
+
+                    if (p > 0) then
+                        !3D
+                        if (grid_geometry == 3) then
+                            max_dt(j, k, l) = cfl_target*min(dx(j)/(abs(vel(1)) + c), &
+                                                             dy(k)/(abs(vel(2)) + c), &
+                                                             fltr_dtheta/(abs(vel(3)) + c))
+                        else
```
sbryngelson: i suspect this logic already exists somewhere else in the code?

src/simulation/m_time_steppers.fpp
```diff
@@ -892,6 +910,127 @@ contains
...
+        ix%end = m; iy%end = n; iz%end = p
+        call s_convert_conservative_to_primitive_variables( &
+            q_cons_ts(1)%vf, &
+            q_prim_vf, &
+            gm_alpha_qp%vf, &
+            ix, iy, iz)
+
+        ! Computing Stability Criteria at Current Time-step ================
+        !$acc parallel loop collapse(3) gang vector default(present) private(alpha_rho, vel, alpha, Re, fltr_dtheta, Nfq)
+        do l = 0, p
```
sbryngelson, line 942: this is only for inviscid computations. what about the viscous CFL?

src/simulation/m_time_steppers.fpp
wilfonba: These quantities are calculated in the same way when computing the CFL in `s_write_runtime_info`, so I'm not sure if this is correct. I copy and pasted much of this subroutine.

src/simulation/m_time_steppers.fpp
wilfonba: This logic is copied from `m_data_outpu` where the CFL is calculated given a specific delta t, but computing a delta t given a CFL number requires separate logic. At east as far as I can tell.

src/simulation/m_time_steppers.fpp
wilfonba: See comment on previous requested change.

src/simulation/m_time_steppers.fpp
wilfonba: I'll add the viscous CFL

wilfonba: I've added this. Here's a `run_time_info` file showing the limiting factor switching between viscous and inviscid CFL being the limiting factor.
[run_time.txt](https://github.com/user-attachments/files/16901167/run_time.txt)

src/simulation/m_time_steppers.fpp
sbryngelson: if you're copy/pasting then shouldn't the logic be extracted to avoid duplicating code?

src/simulation/m_time_steppers.fpp
```diff
@@ -995,23 +997,52 @@ contains
...
+
+                        if (any(Re_size > 0)) then
+                            vcfl_dt = cfl_target*(dx(j)**2d0)/minval(1/(rho*Re))
+                        else
+                            vcfl_dt = dflt_vcfl_dt
+                        end if
+
                     end if
+
+                    max_dt(j, k, l) = min(icfl_dt, vcfl_dt)
```
sbryngelson: can you just put an `if` statement here to only use `icfl_dt` if there is a Reynolds number, otherwise use the `min(..,..)` statement you have? this also avoids your if/else above and the use of `dflt_vcfl_dt`

src/simulation/m_time_steppers.fpp
sbryngelson: if you're copy/pasting then shouldn't the logic be extracted to avoid duplicating code?

src/simulation/m_time_steppers.fpp
sbryngelson: i see. maybe we need two subroutines that can be used everywhere: (1) compute CFL from `dt` and (2) compute `dt` from CFL.

src/simulation/m_time_steppers.fpp
wilfonba: I copy and pasted the structure and then changed the code in the structure. Calculating CFL(dt) is fundamentally different from calculating dt(CFL).

sbryngelson: i see. see my other comment then
> i see. maybe we need two subroutines that can be used everywhere: (1) compute CFL from `dt` and (2) compute `dt` from CFL.

src/simulation/m_time_steppers.fpp
wilfonba: Good point

src/simulation/m_time_steppers.fpp
sbryngelson: i also don't trust the CFL number computed in runtime_info. It might be sufficient for an output file, but this seems to avoid some edge cases I mentioned?

src/simulation/m_time_steppers.fpp
wilfonba: Ok

src/simulation/m_time_steppers.fpp
wilfonba: What would you suggest naming a subroutine that contains these lines? There's no obvious name for what these lines of code do to me. What about these routines would make you not trust them? The 1species to_mixture_variables` routine accounts for bubbles, and `compute_speed_of` sound takes `model_eqns` and bubbles into account.

sbryngelson: compute_cfl?

wilfonba: I interpreted you suggestion as abstracting these lines of code into a subroutine 
```                 
                    do i = 1, num_fluids
                        alpha_rho(i) = q_prim_vf(i)%sf(j, k, l)
                        alpha(i) = q_prim_vf(E_idx + i)%sf(j, k, l)
                    end do

                    if (bubbles) then
                        call s_convert_species_to_mixture_variables_bubbles_acc(rho, gamma, pi_inf, qv, alpha, alpha_rho, Re, j, k, l)
                    else
                        call s_convert_species_to_mixture_variables_acc(rho, gamma, pi_inf, qv, alpha, alpha_rho, Re, j, k, l)
                    end if

                    do i = 1, num_dims
                        vel(i) = q_prim_vf(contxe + i)%sf(j, k, l)
                    end do

                    vel_sum = 0d0
                    do i = 1, num_dims
                        vel_sum = vel_sum + vel(i)**2d0
                    end do

                    pres = q_prim_vf(E_idx)%sf(j, k, l)

                    E = gamma*pres + pi_inf + 5d-1*rho*vel_sum + qv

                    H = (E + pres)/rho
```

wilfonba: `compute_cfl_from_dt` and `compute_dt_from_cfl` would both call whatever these lines are abstracted into

sbryngelson: ultimately, computing the CFL (or dt from CFL) requires the speed of sound, the mesh spacing, and the time step size. so we need a routine that can get the speed of sound from whatever variables we have at present, regardless of multiphase model. this should already be encapsulated in the `s_compute_speed_of_sound` routine if I understand correctly. If it isn't, then can you explain to me what's going on

sbryngelson: but you are (obviously) correct tht compute_cfl_from_dt and the reverse would call some of these lines of code, but it looks like we are doing this in multiple places? even if we are only doing it in one place, it still deserves its own routine.

wilfonba: `s_compute_speed_of_sound` accounts for the model being used as well as sub-grid bubbles. I'll just throw these into a subroutine call `s_get_cell_state`.

sbryngelson: it looks like in the end you are trying to get `H`, which is actually repeated code from the Riemann solvers as well. so we should be able to use the same code in several places.

sbryngelson: `s_compute_enthalpy`, I suppose. similarly, we have a subroutine called `s_compute_pressure` (or some such) that is used in a similar way.

wilfonba: Splitting it into `s_compute_pressure` and `s_compute_enthalpy` makes sense. Though `s_compute_enthalpy` will be a 1 line subroutine.

wilfonba: I think `s_compute_mixture_variables` actually makes the most sense, since this is what all these lines ultimately do.

sbryngelson: whatever makes most sense sounds good to me

wilfonba: Well, there's already a subroutine called that. I guess I'll do `s_compute_cell_state`

sbryngelson: but it isnt the cell state per se, you are just computing the enthalpy, right? perhaps you can create a function called `compute_enthalpy` that can all other routines as above, including the `compute_mixture_variables`

wilfonba:
Most all of the requested changes have been addressed. I feel that updating `m_riemann_solvers` would be better suited for a separate PR to adhere to the "This PR comprises a set of related changes with a common goal" checkbox. I've already kind of exceeded this with my changes to `m_data_output`, so I'd like to minimize the violation.

sbryngelson:
> Most all of the requested changes have been addressed. I feel that updating `m_riemann_solvers` would be better suited for a separate PR to adhere to the "This PR comprises a set of related changes with a common goal" checkbox. I've already kind of exceeded this with my changes to `m_data_output`, so I'd like to minimize the violation.

Please open an issue detailing what you didn't do, then, please.

wilfonba:
All the tests pass here but when I build on Wingtip for GPUs I get an error for some reason, even on a fresh clone of my fork. I'll update here when I fix that

Issue #498, opened 2024-07-02
CI Batch jobs that don't get nodes don't trigger failure stop code


sbryngelson:
CI Batch jobs that don't get nodes (due to queue wait times or down nodes or whatever) don't trigger a failure stop code. For example,

```
Run bash .github/workflows/frontier/submit.sh .github/workflows/frontier/test.sh 
Submitted batch job 2070639
Error: The operation was canceled.
```
and
```
Run cat test-gpu.out
cat: test-gpu.out: No such file or directory
Error: Process completed with exit code
```

from https://github.com/MFlowCode/MFC/actions/runs/9748333690/job/26903120742

sbryngelson:
I guess the problem is 
```yaml
  self:
    name: Self Hosted
    if: github.repository == 'MFlowCode/MFC' && needs.file-changes.outputs.checkall == 'true'
    needs: file-changes
    continue-on-error: true
```
in the `test.yml` file?

sbryngelson:
These people seem somewhat confused by the situation. Continue-on-error should be able to be enabled without giving a green check to failed jobs: https://github.com/orgs/community/discussions/15452

PR #517, opened 2024-07-11
Fail a CI self-hosted job if frontier doesn't get a node

sbryngelson:
Fail a CI self-hosted job if frontier doesn't get a node

Related to my comments on the issue here:

https://github.com/MFlowCode/MFC/issues/498

Let me know if you think there is a better way.


sbryngelson:
Works. Merging

PR #519, opened 2024-07-12
Fix 3D IBM Infinite CFL Number on GPUs

Sam-Briney:
## Description

Fixes 3D IBM infinite CFL number when running on GPUs. These modifications were suggested by @haochey and @anandrdbz via slack

### Type of change

### Scope

## How Has This Been Tested?

- [x] Tested by running the 3D_ibm example on two A100 GPUs
visit0000

**Test Configuration**:

The test was run using Nvidia A100 GPUs on HiPerGator at UF
`python/3.8, nvhpc/24.5, openmpi/4.1.6, cmake/3.26.4`

## Checklist
- [x] I ran ```./mfc.sh format``` before committing my code
- [x] This PR does not introduce any repeated code (it follows the [DRY])

### DIFF
--- a/src/simulation/m_start_up.fpp
+++ b/src/simulation/m_start_up.fpp
@@ -1417,6 +1417,10 @@ contains
             !$acc update device(palpha_eps, ptgalpha_eps)
         end if
 
+        if (ib) then
+            !$acc update device(ib_markers%sf)
+        end if
+
     end subroutine s_initialize_gpu_vars
 
     subroutine s_finalize_modules
--- a/src/simulation/m_time_steppers.fpp
+++ b/src/simulation/m_time_steppers.fpp
@@ -303,14 +303,6 @@ contains
 
         call s_compute_rhs(q_cons_ts(1)%vf, q_prim_vf, rhs_vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
 
-        if (ib .and. t_step == 1) then
-            if (qbmm .and. .not. polytropic) then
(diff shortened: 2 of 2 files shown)


sbryngelson:
I'm confused by the removal of these:

```fortran
        if (ib .and. t_step == 1) then
            if (qbmm .and. .not. polytropic) then
                call s_ibm_correct_state(q_cons_ts(1)%vf, q_prim_vf, pb_ts(1)%sf, mv_ts(1)%sf)
            else
                call s_ibm_correct_state(q_cons_ts(1)%vf, q_prim_vf)
            end if
        end if
```

If you remove all of these then the whole subroutine `s_ibm_correct_state` can be removed because these are the only places it gets called. What exactly is the purpose of this routine (or removing it?) @anandrdbz @Sam-Briney

Sam-Briney:
@sbryngelson I think it gets called in a few other places as well. For example: https://github.com/MFlowCode/MFC/blob/78fd0c51735f1c514d6b804531278b6dd3a3ef8f/src/simulation/m_time_steppers.fpp#L690

I believe it was getting called twice instead of just once at the first time step.

sbryngelson:
Ah I see, you're correct @Sam-Briney. Does this fix the issue with the visualization inside the IBM being incorrect on >1 GPU?

Sam-Briney:
@sbryngelson It does not fix the visualization issue. I thought it might be better to make a more focused PR to make this functionality available in the short term

sbryngelson (approved):

sbryngelson:
Makes sense. Merging.

PR #522, opened 2024-07-14
Acoustic Source: Fix Equations & Add Transducer Arrays & Refactor

ChrisZYJ:
## Description

- Added 2D/2D-axisym transducer array support type
- Added 3D transducer array ring support type
- Modified Supports 1-4 for 1D/2D/3D for consistency
- Fixed 3D spherical support code that breaks momentum into components (wrong angle_z)
- Fixed 2D cylindrical support mass source correction term
- Fixed Gaussian mass source correction term for all cylindrical and spherical supports (was absent)
- Fixed assymetry in large 2D/2D-axisym/3D transducers (was using small angle approx; needed aperture size << focal length)
- Corrected the wrong 3D cylindrical case to 2D-axisym
- Added dipole functionality
- Renamed monopole -> acoustic_source & Refactored subroutine for clarity

### Type of change

### Scope

## How Has This Been Tested?

- [x] Passed all tests using CPU locally
- [x] Passed all tests using GPU locally
- [x] Passed all tests using GPU on Delta
- [x] Support 7: 2D Transducer array with 4 elements:

https://github.com/user-attachments/assets/55e5f34d-ae79-4af0-b502-74686e39d164

- [x] Support 11: 3D Transducer array with 2 layers and 18 elements:

https://github.com/user-attachments/assets/062d2b76-7186-4a26-bf90-a2ba495b358b

- [x] Check momentum correction terms by comparisons with Maeda's paper
image
 
 - 3D Spherical Transducer
image

 - 2D Cylindrical Transducer
image

* Note that the power of the correction factor for the 2D cylindrical transducer case uses an empirical factor of 0.85, which improves upon the 0.5 factor suggested in Maeda's paper. 

  - Gaussian pulse
image

- [x] Symmetry for large transducers (Before vs After)
  - 2D Cylindrical Transducer
image
  - 3D Spherical Transducer
image

  - Ubuntu 22.04.4 LTS; GNU v11.4.0
  - Ubuntu 24.04 LTS; NVHPC 23.11.0

## Checklist

- [x] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled (only NVHPC)

[support3_nsys.txt](https://github.com/user-attachments/files/16340030/support3_nsys.txt)
[support11_18elements_nsys.txt](https://github.com/user-attachments/files/16340031/support11_18elements_nsys.txt)
  - 8M points: 2 GPUs: 157 sec; 4 GPUs: 87 sec

## Note
- Golden files are regenerated for Acoustic and Bubbles, as Bubbles is also dependent on Acoustic implementations. The simulation duration in the Acoustic tests are significantly increased so waves can propagate a considerable distance, and this increases the error and requires a slight increase in tolerance for Acoustic. No changes are made to tolerance for Bubbles.


src/simulation/m_monopole.fpp
```diff
@@ -165,119 +180,107 @@ contains
...
+
+                        if (bubbles) then
+                            if (mpp_lim .and. (num_fluids > 2)) then
                                 !$acc loop seq
                                 do ii = 1, num_fluids
                                     myRho = myRho + myalpha_rho(ii)
                                     n_tait = n_tait + myalpha(ii)*gammas(ii)
                                     B_tait = B_tait + myalpha(ii)*pi_infs(ii)
                                 end do
+                            else if (num_fluids > 2) then
```
sbryngelson: So long as you are changing things, it seems like this part of the code that computes the speed of sound `sound` should either be computed by `s_compute_speed_of_sound` or, if that isn't the right routine for some reason, an abstracted-out subroutine

src/simulation/m_global_parameters.fpp
```diff
@@ -420,6 +420,7 @@ module m_global_parameters
     logical :: monopole !< Monopole switch
     type(mono_parameters), dimension(num_probes_max) :: mono !< Monopole parameters
     integer :: num_mono !< Number of monopoles
+    real(kind(0d0)), parameter :: acoustic_spatial_support_width = 2.5d0 !< Spatial support width of the monopole, used in f_source_spatial
```
sbryngelson: i think `m_constants` in `common/` holds parameters?

src/simulation/m_global_parameters.fpp
```diff
@@ -615,22 +616,27 @@ contains
...
+            mono(j)%wavelength = dflt_real
+            mono(j)%frequency = dflt_real
+            mono(j)%gauss_sigma_dist = dflt_real
+            mono(j)%gauss_sigma_time = dflt_real
+            mono(j)%npulse = dflt_real
             mono(j)%dir = dflt_real
-            mono(j)%npulse = dflt_int
-            mono(j)%pulse = dflt_int
-            mono(j)%support = dflt_int
+            mono(j)%delay = dflt_real
```
sbryngelson: I think we agreed that we could default this to zero. Otherwise, we have to change all the existing input `case.py` files to have `delay -> 0`, right?

src/simulation/m_monopole.fpp
```diff
@@ -165,119 +180,107 @@ contains
...
-                                s1 = f_g(the_time, sound, const_sos, q, term_index)* &
-                                     f_delta(j, k, l, loc_mono(:, q), length(q), q, angle, angle_z)
-                            end if
+                        s2 = f_g(the_time, sound, q, term_index)* &
+                             f_delta(j, k, l, loc_mono(:, q), length(q), q, angle, ratio_x_r, ratio_y_r, ratio_z_r)
 
+                        if (support(q) == 5 .or. support(q) == 7) then
+                            term_index = 1
+                            s1 = f_g(the_time, sound, q, term_index)* &
+                                 f_delta(j, k, l, loc_mono(:, q), length(q), q, angle, ratio_x_r, ratio_y_r, ratio_z_r)
```
sbryngelson: much of the code in `m_monopole` never gets covered in the test suite, but perhaps easily could. check codecov

src/simulation/m_monopole.fpp
```diff
@@ -370,44 +389,52 @@ contains
         real(kind(0d0)) :: hxnew_cyl, hynew_cyl
         real(kind(0d0)) :: sig
         real(kind(0d0)) :: f_delta
+        real(kind(0d0)) :: angle, ratio_x_r, ratio_y_r, ratio_z_r
```
sbryngelson: how about `ratio_r(1:3)` insteady of `x/y/z`

src/simulation/m_monopole.fpp
```diff
@@ -16,7 +16,7 @@ module m_monopole
```
sbryngelson: rename from `m_monopole` to `m_acoustic_src`, perhaps?

src/simulation/m_monopole.fpp
ChrisZYJ: You are right. I kept it unchanged, but should have switched to using `s_compute_speed_of_sound`. I've checked and they produce the same result. It'll need a few more lines before calling `s_compute_speed_of_sound`, and I can abstract out all these into a subroutine.

src/simulation/m_global_parameters.fpp
ChrisZYJ: yes I will change it

src/simulation/m_global_parameters.fpp
ChrisZYJ: Regarding this, I've set the default delay to 0 for sine wave, but not for Gaussian pulse:

```
            if (f_is_default(mono(i)%delay)) then ! m_checker guarantees mono(i)%delay is set for pulse = 2 (Gaussian)
                delay(i) = 0d0 ! Defaults to zero for sine and square waves
            else
                delay(i) = mono(i)%delay
            end if
```

This is because it doesn't make sense for Gaussian pulse's t0 to default to zero - it'll become a half-Gaussian pulse. The original implementation used t0=3.5*sigt, where sigt is the sigma_t for the Gaussian bell curve. This makes sense but is very arbitrary, so I removed the default and enforced that the user must input a value. 

By only defaulting the sine wave's delay to zero, the existing input case.py files will still work. (I don't think any existing tests or examples uses Gaussian pulse)

src/simulation/m_monopole.fpp
ChrisZYJ: I'll need many tests, but I'm leaving them to the end, as I'l also regenerate some of the tests

src/simulation/m_monopole.fpp
ChrisZYJ: I kind of prefer ratio_x_r because it's very explicit and clear that it's a ratio between x and r. But I can change it if you think `ratio_r(1:3)` suits the code better

src/simulation/m_monopole.fpp
ChrisZYJ: Yes! This is one of the next steps. 

Wanted to ask you if I should also change the `Monopole` option in case file to `acoustic_src` or `acoustic_source`, so we can change everything in the documentation together? This would be much cleaner but it also means changing all existing case files containing `Monopole` though.

ChrisZYJ:
@sbryngelson Thanks for reviewing the code so carefully and the helpful comments! This is an early draft to verify that the equations work and I'm still making many changes to improve the code

ChrisZYJ:
I have one more question - can I break support 5 (transducer) into support 5 for 2D transducer and support 6 for 3D transducer? Currently support 5 is not documented, so I doubt anyone is using it. Support 6 is not working in the first place. By breaking each dimension into a different case, it'll be more consistent with the rest of the case file parameters. 

image

sbryngelson:
@ChrisZYJ this sounds good! Thanks

src/simulation/m_monopole.fpp
ChrisZYJ: @sbryngelson Would you like me to perform this change? Thanks!

src/simulation/m_monopole.fpp
sbryngelson: this sounds like a plan!

src/simulation/m_monopole.fpp
sbryngelson: oh, i didn't even realize that's what it was supported to be. perhaps a better name is still needed then.

src/simulation/m_monopole.fpp
sbryngelson: sounds good

src/simulation/m_global_parameters.fpp
sbryngelson: hmm. my understanding of how i implemented the delay feature was that it doesn't "turn on" the source term until a certain time is reached... of course a Gaussian wave is infinitely "wide," so one is forced to create a Gaussian with some fixed standard deviation/width. does that make any sense?

src/simulation/m_monopole.fpp
sbryngelson: i think `acoustic_source` is preferable, even though it is a 'breaking change' of sorts. a quick find/replace on the examples and toolchain directories should take care of most of the changes.

ChrisZYJ: That's what I think too! Will change all relevant places

src/simulation/m_monopole.fpp
ChrisZYJ: Yes I changed it to xyz_to_r_ratios(1:3)

src/simulation/m_global_parameters.fpp
ChrisZYJ: You are right that a Gaussian pulse is infinitely wide and we need to cut it off somewhere. 3.5 is a very reasonable choice, but the user should understand and have control over the parameter. A Gaussian pulse in an acoustic simulation is most commonly used to find the impulse response, in which case an insufficient delay can cause a large discontinuity that shows up as oscillations in the frequency domain. The threshold for "sufficient delay" is determined by the specific simulation setup and the required accuracy of the computed transfer function, so I think it is crucial for anyone using it to know exactly where the pulse is cut off, and a default value shouldn't be used.

sbryngelson: seems reasonable to me

ChrisZYJ:
@sbryngelson May I know why the Frontier compilation failed? I can't figure it out from the build output on Github. Thanks!

ChrisZYJ:
Also, it appears that one test case fails only for Intel compilers, which is weird. Do you have any idea on that too? Thank you!

sbryngelson:
@ChrisZYJ not sure what's going on on Frontier. could be some allocatable issue. i will have a quick look. the build output was uninformative for me so I reran it. 

For Intel compilers we sometimes see a slight different in compiler optimizations (for example, if the goldenfiles were built using gfortran but you're testing against ifort), resulting in different tolerances that can be expected. 

It's possible that we just need a slightly higher tolerance for certain types of tests. We have done this before (you can find the modified tolerance in the toolchain).

ChrisZYJ:
Thanks! I'll try again with a higher tolerance. It would be nice if I can test it locally, but I couldn't get MFC running on windows. Will direct the question to the Slack channel

sbryngelson:
@ChrisZYJ looking into the compiler issue, but in the meantime it looks like you need to rebase your PR? not sure of the exact steps for this, but right now it's showing you modifying files from master that are actually the current state of master. e.g., `.github/workflows/coverage.yml`. this makes seeing the actual changes you made challenging.

ChrisZYJ:
@sbryngelson Sorry for the oversight. I've just done a rebase and github should now only be showing my changes now.

ChrisZYJ:
It seems like the ifort test has failed again despite increasing the tolerance by 50x. I'm still trying to install ifort locally and will try to fix the problem if I succeed.

sbryngelson:
you can load Intel compilers on a bunch of different clusters, including Delta

ChrisZYJ:
I tested it on Delta using ifort but couldn't reproduce the problem. Very weird. Almost all tests fail there - I needed to increase the tolerance from 1e-12 to 1e-8, then everything passes. And there's nothing special about `2F58A83F: 1D -> Acoustic Source -> Square -> Frequency` - the tolerance it needs is the same as other tests. 

I might just increase the tolerance for only square waves and try the GitHub test again.

ChrisZYJ:
Square waves have finally passed ifort tests! Sorry for the multiple commits - I couldn't reproduce the compiler-specific issue anywhere else. This was likely caused by arithmetic discrepancies between compilers, resulting in an offset in time and a max-norm error after applying the square wave function.

sbryngelson (approved):

sbryngelson:
@henryleberre requesting your review of the source code before merging

PR #525, opened 2024-07-17
New Example CI + Fixes to some broken example!

### LINKED ISSUE #474: Add `examples/` into CI
Make each example case sufficiently small such that it can run on one processor for at least a few time steps. Then, add a few time steps of each example case to CI and set I/O to write at the end. Then check output for NaNs/Inftys.

We grow the size of the test suite this way, though it isn't as thorough a test as there will be no goldenfiles (though I suppose we could add them).

okBrian:
## Description

Added New Example CI, - Average runtime - 30 minutes

Why MacOS runner?
From my initial testing its much faster than the ubuntu images.  Some examples ran over two times faster on the MacOS runner

Why 15625, (125, 125), or (25, 25, 25) for cell boundaries? 
Some 3D examples break if m < 25, and this seems like a good spot to put examples so they don't take too long 

Fixes #474 

### Type of change
### Scope
## How Has This Been Tested?
- [x] Github Actions Run all example with CI 
- [x] Run all examples locally 

**Test Configuration**:
gcc 14, Ubuntu 22,04.4 LTS
Github Action Macos Runner
## Checklist


okBrian:
I'm not exactly sure whats wrong with the test suite, but I think if I add the --generate flag they should be okay. Do we want the test suite to even test for examples because that would add another 30 minutes or more to them. Also I need to remove the -r flag from the example suites but it looks like it worked fine.

sbryngelson:
@okBrian we can add this as a separate GitHub workflow (Examples-test.yml, Example Smoke Test).

sbryngelson:
the `-r` `--generate` things I don't understand at the moment

sbryngelson:
@okBrian It's important that these cases run and produce output that isn't nonsense. So you can run the "example suite" with `./mfc.sh test —a` (+ your flags, I guess?), where `-a` already exists. ' This runs the files through `post_process` and then checks for NaNs/Infinity. 

If `parallel_IO` is not true then I'm not sure if `-a` works or not. You may need to turn it on via your toolchain tricks.

henryleberre:
> I'm not exactly sure whats wrong with the test suite, but I think if I add the --generate flag they should be okay. Do we want the test suite to even test for examples because that would add another 30 minutes or more to them. Also I need to remove the -r flag from the example suites but it looks like it worked fine.

If you specify `--generate` it will compare the pre-process and simulation results against itself (not with the reference values from this PR). So you shouldn't pass `--generate`.

Whether we should or should not use the `-r` (`--relentless`) flag is debatable but we have currently opted not to use it so we should probably keep it that way. The main advantage is that we don't spend CI resources running more tests if one already failed and the workflow is marked as failed right away when the first test case fails.

toolchain/mfc/test/test.py
```diff
@@ -180,7 +181,7 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
                 raise MFCException(f"Test {case}: {msg}")
 
     if ARG("test_all"):
-        case.delete_output()
+        # case.delete_output()
```
sbryngelson: might want to put this back?

okBrian: It breaks the 1D bubble cases. Not sure why maybe its deleting the D folder I'm creating.

okBrian:
I am expecting the code to not work with --no-mpi so thats another thing to debug... I think everything with mpi should pass hopefully.

toolchain/mfc/test/test.py
sbryngelson: perhaps associated with current bug in CI that if it has to 'retry' a simulation, it will try to create a directory that already exists

sbryngelson:
The quicker it is fixed the quicker the CI problems become someone else's, not yours ;)

sbryngelson:
merge master into this PR

sbryngelson:
this won't work on Phoenix until @henryleberre merges #581 and you pull in the module changes for Phoenix (we moved to RHEL9 nodes).

okBrian:
Oh my bad I thought they were done based on the most recent commit.

henryleberre:
Well, they should be working but there's an issue with @sbryngelson's Python installation. I'm pretty sure it's because he deleted bits from his `~/.local` folder. The CI is complaining that a "pip" executable cannot be found in `~/.local/bin/pip`. @sbryngelson I put my `.local` in `/storage/coda1/p-sbryngelson3/0/shared/local-for-spencer` so you can copy what's missing.

sbryngelson:
> Well, they should be working but there's an issue with @sbryngelson's Python installation. I'm pretty sure it's because he deleted bits from his `~/.local` folder. The CI is complaining that a "pip" executable cannot be found in `~/.local/bin/pip`. @sbryngelson I put my `.local` in `/storage/coda1/p-sbryngelson3/0/shared/local-for-spencer` so you can copy what's missing.

I actually never deleted anything in my `.local` directory, just suggested you try it as a debug measure.

sbryngelson:
You can see that this works by looking here: https://github.com/MFlowCode/MFC/actions/runs/10498547818/job/29083688602

okBrian:
Hey is GT Phoenix fully migrated or should I keep waiting?

henryleberre:
@okBrian _It should be working now &trade;_. You should just need to rebase on upstream.

sbryngelson:
@okBrian you can sync with master now, it should work

sbryngelson:
Phoenix is down so none of the Phoenix CI will pass for now... will reboot those when they come back online

sbryngelson:
@okBrian Phoenix CI is back online, but now you will need to rebase as @henryleberre made a change for Frontier CI as well as the packer golden files.

henryleberre:
Also, @okBrian you need to regenerate the golden files for your new cases. Alternatively, I can run a script for you that strips the `<x> <y> <z>` entries for you. It's easiest if you regenerate.

sbryngelson:
@okBrian phoenix runners are back up, though this branch needs to be synced with master

sbryngelson:
@okBrian FYI the runners are taking a long time to run GPU cases because the queue is very full after the Phoenix restart/reboot. You might move onto other issues while waiting.

sbryngelson:
@okBrian you should be able to login to Frontier to see why your CI is failing. It is failing at build (link) time. It might be due to a change you made in the `src/`, or it could be due to the flags you added to CMake. It might be easier to debug Frontier by changing the flags and rerunning CI or just doing it on your own after login into Frontier and building there.

sbryngelson:
@okBrian, I also noticed that it failed in some cases with GNU and Intel compilers. This suggests that the flags might not be doing what they should (or you didn't add enough). One way to be sure you "have enough" is just using `-O0` for all `--strict` builds and ensuring that passes. If that doesn't work, then there's something else awry.

CMakeLists.txt
```diff
@@ -186,13 +207,15 @@ elseif ((CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC") OR (CMAKE_Fortran_COMPILER_
     if (CMAKE_BUILD_TYPE STREQUAL "Debug")
         add_compile_options(
             $<$<COMPILE_LANGUAGE:Fortran>:-O0>
+            -C
```
sbryngelson: i'm not sure this syntax works. you could double check by looking at the cmake output of a specific run. i think you need to add
`  $<$<COMPILE_LANGUAGE:Fortran>:-<flag>>` on separate lines.

okBrian:
With the most recent commits these are the problematic examples
MacOS, MPI, Debug, False
2D -> Example Viscous Error: 2.51E-02 
1D -> Example -> sodHypo Error: 1.01E-04

Ubuntu MPI. no-Debug, True
1D -> Example -> hypo_2materials Error: 1.74E-04
2D -> Example -> ibm_cfl_dt Error: 1.51E-04
1D -> Example -> sodHypo: Error: 1.01E-04
2D -> Example -> viscous Error: 2.51E-02
2D -> 1 Fluid(s) -> IBM Error: 1.33E-10
2D -> 1 Fluid(s) -> Viscous -> IBM Error: 2.21E-10
2D -> 1 Fluid(s) -> Viscous -> IBM -> model_eqns=3 Error: 2.21E-10
2D -> 2 Fluid(s) -> IBM Error: 5.95E-10

Ubuntu Mpi, Debug, False
2D -> Example -> ibm_cfl_dt Error: 1.65E-04

GT CPU Pheonix
2D -> Example -> laplace_pressure_jump -> NaN's detected in Case

CMakeLists.txt
okBrian: I will fix this in my next commit. Thanks

sbryngelson:
@okBrian can you visualize the output of a few of these cases (compared to the cases that are 'working')? I suspect that, given how diverse they are, some of them are just numerically unstable and eventually produce very large numbers that different compilers handle differently.

okBrian:
Remaking PR

PR #527, opened 2024-07-17
Update modules for Carpenter

lee-hyeoksu:
## Description

On Carpenter, the current set of modules causes warnings possibly due to compatibility issue among old compilers. Thus, this PR update the modules so that no warning appears.

Also, on carpenter,  `module purge` was not executed. So I edited the code in `toolchain/bootstrap/modules.sh`to execute `module purge` properly .

### Type of change

- [x] Something else

### Scope

## How Has This Been Tested?

- test suite passed
- tested with `examples/2D_shockbubble` for both interactive and batch jobs and worked well (no more warnings!)

### DIFF
--- a/.github/workflows/coverage.yml
+++ b/.github/workflows/coverage.yml
@@ -40,12 +40,11 @@ jobs:
         run: /bin/bash mfc.sh test -a -j $(nproc)
 
       - name: Upload coverage reports to Codecov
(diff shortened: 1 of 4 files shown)


sbryngelson:
Made a minor change, adding comments. Otherwise, it seems fine.

sbryngelson (approved):

PR #530, opened 2024-07-19
Does benchmarking work?

sbryngelson:
A test for benchmark CI.


sbryngelson:
@henryleberre this seems to be working now. not sure what was going on before. closing...

PR #531, opened 2024-07-19
Fix NVTX ranges

### LINKED ISSUE #501: Nsight Systems profiling NVTX blocks are broken
For NVIDIA GPU Nsight Systems profiling, the NVTX blocks were, at some point, broken. It is unclear why, though it seems like it should be easy to track down. An example is in the image below for a 3D case.

Broken (current):
Screenshot 2024-07-03 at 14 27 18

Working (old):
Screenshot 2023-05-05 at 09 57 57

AiredaleDev:
## Description

The boundaries for the NVTX ranges for WENO and the Riemann solver were broken during the process of adding MI200 support.

Fixes #501 

### Type of change

### Scope

## How Has This Been Tested?

- [x] Profile the 3D performance test for 10 time steps, yield the same profile as the one attached to this PR.
[fixed_prof.zip](https://github.com/user-attachments/files/16301332/fixed_prof.zip)

**Test Configuration**:

* Rogue's Gallery UWing Nodes (GH200) -- NVHPC 24.1

## Checklist

- [x] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this [PR](url)

### DIFF
--- a/src/simulation/m_rhs.fpp
+++ b/src/simulation/m_rhs.fpp
@@ -731,6 +731,8 @@ contains
         integer :: i, j, k, l, q, ii, id !< Generic loop iterators
         integer :: term_index
 
+        call nvtxStartRange("Compute_RHS")
+
         ! Configuring Coordinate Direction Indexes =========================
         ix%beg = -buff_size; iy%beg = 0; iz%beg = 0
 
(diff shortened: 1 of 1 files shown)


sbryngelson:
This fixes it! Looking at the new ranges, I think it would be useful to wrap everything within `s_compute_rhs` in an overarching `nvtx` range called `Compute_RHS` (below the `time_step scale`, but above `RHS-WENO`, for example). you can just put it at the beginning and end of the subroutine.

sbryngelson (approved):

PR #538, opened 2024-07-28
Low mach number correction for HLLC Riemann solver

lee-hyeoksu:
## Description

The HLLC Riemann solver is known to have an accuracy issue at the low Mach number limit. The new feature addresses this issue using two different methods (Chen et al. 2022 and Thornber et al. 2008)

### Type of change

### Scope

## How Has This Been Tested?

- [x] 2D_GreshoVortex for `Ma = 0.001` (added to `examples`)

**Test Configuration**: 2D Gresho vortex is an exact solution to the incompressible N-S equations. The flow field should remain the same unless there is a numerical dissipation. The figure attached below shows the simulation results at Mach number 0.001 with `low_Mach = 0` (no correction), `low_Mach = 1` (Chen et al. 2022) and `low_Mach = 2` (Thornber et al. 2008). Both correction methods work well.

1. Richardson (cpu)
2. Bridges2 (cpu)
3. Delta (cpu/A100x4)

Picture1

## Checklist

I don't have access to MI200+ GPUs.

[nsys.zip](https://github.com/user-attachments/files/16409275/nsys.zip)

I don't have access to machines that work with omniperf profile, I guess. 

For this one, should I turn on `low_Mach` feature on `examples/3D_performance_test` and compare with `low_Mach = 0 (off)` results for 1/2/8 GPUs?


lee-hyeoksu:
@sbryngelson Could you re-run the benchmark test? It failed with exit code 1 without any useful messages. If it is really a problem, do you have any idea what went wrong? Thanks!

examples/2D_GreshoVortex/case.py
```diff
@@ -0,0 +1,96 @@
...
+    'cons_vars_wrt'                 :'T',
+    'prim_vars_wrt'                 :'T',
+    'parallel_io'                   :'T',
+    'fd_order'                      : 1,
+    'omega_wrt(3)'                  :'T',
+    # ==========================================================================
+
+    # Patch 1 ==================================================================
+    'patch_icpp(1)%geometry'        : 7,
+    'patch_icpp(1)%hcid'            : 203,
```
sbryngelson, line 78: what `hcid` does this correspond to? shouldn't it be in the source code somewhere?

hyeoksu-lee: Ah the hardcoded id for Gresho vortex was added to the MFC in my previous PR (not sure which one though). You can find it in `pre_process/include/2dHardcodedIC.fpp`. I should not have included it in the previous PR since it was irrelevant but it was mistakenly included in the previous one.

sbryngelson: no problem, thanks

PR #544, opened 2024-07-29
Upstream CheMFC infrastructure

henryleberre:
## Description

Many changes are needed to implement chemistry and some are quite significant. This PR upstreams the _core_ set of modifications that relate to code infrastructure. The PR also includes some non-chemistry, nice-to-have, updates and features.

### Scope

## How Has This Been Tested?

- [x] Local OSX `./mfc.sh test`.
- [x] GitHub CI.

What is important here is not the correctness of the chemistry, but that these additions don't break existing code.

### DIFF
--- a/misc/fpp_to_fypp.sh
+++ b/misc/fpp_to_fypp.sh
@@ -1,0 +1,6 @@
+#!/bin/bash
+
+for file in $(find src -type f | grep -Ev 'autogen' | grep -E '\.fpp$'); do
+    echo "$file"
+    mv "$file" "$(echo "$file" | sed s/\.fpp/\.fypp/)"
+done
--- a/src/common/include/macros.fpp
+++ b/src/common/include/macros.fpp
@@ -140,17 +140,19 @@
 #endif
 #:enddef
 
-#:def PROHIBIT(*args)
-    #:set condition = args[0]
-    #:if len(args) == 1
-        #:set message = '""'
-    #:else
-        #:set message = args[1]
-    #:endif
+#:def PROHIBIT(condition, message = None)
     if (${condition}$) then
-        call s_prohibit_abort("${condition}$", ${message}$)
+        call s_prohibit_abort("${condition}$", ${message or '""'}$)
     end if
 #:enddef
 
 #define t_vec3   real(kind(0d0)), dimension(1:3)
 #define t_mat4x4 real(kind(0d0)), dimension(1:4,1:4)
+
+#:def ASSERT(predicate, message = None)
+    if (.not. (${predicate}$)) then
+        call s_mpi_abort("${_FILE_.split('/')[-1]}$:${_LINE_}$: "// &
+                         "Assertion failed: ${predicate}$. " &
+                         //${message or '"No error description."'}$)
+    end if
+#:enddef
--- a/src/pre_process/m_data_output.fpp
+++ b/src/pre_process/m_data_output.fpp
@@ -97,7 +97,7 @@ contains
         character(LEN=len_trim(t_step_dir) + name_len) :: file_loc !<
             !! Generic string used to store the address of a particular file
 
-        integer :: i, j, k, l, r !< Generic loop iterator
+        integer :: i, j, k, l, r, c !< Generic loop iterator
         integer :: t_step
 
         real(kind(0d0)), dimension(nb) :: nRtmp         !< Temporary bubble concentration
@@ -108,6 +108,8 @@ contains
 
         real(kind(0d0)) :: nR3
         real(kind(0d0)) :: ntmp
+
+        real(kind(0d0)) :: rhoYks(1:num_species) !< Temporary species mass fractions
 
         t_step = 0
 
@@ -240,14 +242,27 @@ contains
 
                     open (2, FILE=trim(file_loc))
                     do j = 0, m
+
+                        if (chemistry) then
+                            do c = 1, num_species
+                                rhoYks(c) = q_cons_vf(chemxb + c - 1)%sf(j, 0, 0)
+                            end do
+                        end if
+
                         call s_convert_to_mixture_variables(q_cons_vf, j, 0, 0, rho, gamma, pi_inf, qv)
 
                         lit_gamma = 1d0/gamma + 1d0
 
-                        if (((i >= cont_idx%beg) .and. (i <= cont_idx%end)) &
-                            .or. &
-                            ((i >= adv_idx%beg) .and. (i <= adv_idx%end)) &
-                            ) then
+                        if ((i >= chemxb) .and. (i <= chemxe)) then
+                            write (2, FMT) x_cb(j), q_cons_vf(i)%sf(j, 0, 0)/rho
+                        else if (((i >= cont_idx%beg) .and. (i <= cont_idx%end)) &
+                                 .or. &
+                                 ((i >= adv_idx%beg) .and. (i <= adv_idx%end)) &
+                                 .or. &
+                                 ((i >= chemxb) .and. (i <= chemxe)) &
+                                 .or. &
+                                 ((i == tempxb)) &
+                                 ) then
                             write (2, FMT) x_cb(j), q_cons_vf(i)%sf(j, 0, 0)
                         else if (i == mom_idx%beg) then !u
                             write (2, FMT) x_cb(j), q_cons_vf(mom_idx%beg)%sf(j, 0, 0)/rho
@@ -258,7 +273,7 @@ contains
                                 q_cons_vf(E_idx)%sf(j, 0, 0), &
                                 q_cons_vf(alf_idx)%sf(j, 0, 0), &
                                 0.5d0*(q_cons_vf(mom_idx%beg)%sf(j, 0, 0)**2.d0)/rho, &
-                                pi_inf, gamma, rho, qv, pres)
+                                pi_inf, gamma, rho, qv, rhoYks, pres)
                             write (2, FMT) x_cb(j), pres
                         else if ((i >= bub_idx%beg) .and. (i <= bub_idx%end) .and. bubbles) then
 
@@ -735,9 +750,12 @@ contains
     subroutine s_initialize_data_output_module
         ! Generic string used to store the address of a particular file
         character(LEN=len_trim(case_dir) + 2*name_len) :: file_loc
+        character(len=15) :: temp
+        character(LEN=1), dimension(3) :: coord = (/'x', 'y', 'z'/)
 
         ! Generic logical used to check the existence of directories
         logical :: dir_check
+        integer :: i
 
         if (parallel_io .neqv. .true.) then
             ! Setting the address of the time-step directory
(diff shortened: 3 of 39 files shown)


misc/fpp_to_fypp.sh
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+for file in $(find src -type f | grep -Ev 'autogen' | grep -E '\.fpp$'); do
```
sbryngelson, line 3: as nervous as it makes me for 'breaking changes' it seems like we should do this (actually apply it)? thoughts?

henryleberre: we could, it wouldn't exactly be a breaking change though. Git usually handles renaming files pretty well. The upside is that vscode's Modern Fortran extension would work a lot better.

sbryngelson: yeah that's the upside. actually, it's the only upside, but since everyone is using vscode except for me it seems worthwhile. we can probably apply this in a separate PR. it would require toolchain find/replace.

henryleberre: I agree. What would need find/replace-ing in the toolchain?

sbryngelson:
Can you add a brief summary of the added features/code cleaning? I realize this is a bit tedious but scrolling through the changed code is harder...

henryleberre:
List of changes:
- (**Main**) MFC <-> Pyrometheus compile-time integration (case setup, builds, code-gen, etc.)
- (**Main**) Parts of `m_chemistry`, some separation of logic for chemistry/non-chemistry cases at compile-time, variable buffers for chemistry, and some early logic like EoS, reactions, and post_processing.
- (**Minor**) Creation of `./mfc.sh spelling`.
- (**Minor**) Multi-threaded `misc/viz.py`.
- (**Minor**) Creation of `m_finite_differences`.
- (**Minor**) Updated `indices.dat` to include the cons/prim variables, including mass fractions.
- (**Minor**) Fixed `s_rectangle` patch for immersed boundaries. Note: I forget how I encountered this. Might have been with a test.
- (**Minor**) Explicitly zero-out `rhs_vf` at the start of `s_compute_rhs`.
- (**Minor**) Fixed the computation `time_avg` in `m_rhs` so that it works in 1D and 2D (not just 3D).
- (**Minor**) Multi-threaded `pip install` if you specify `-j` (could speed up Cantera builds).
- (**Minor**) Created `./mfc.sh spelling` (called by CI).

\dots and I think that's most of them. You can wait for the GT runners to come back up to merge this.

sbryngelson:
> List of changes:
> 
> * (**Main**) MFC <-> Pyrometheus compile-time integration (case setup, builds, code-gen, etc.)
> * (**Main**) Parts of `m_chemistry`, some separation of logic for chemistry/non-chemistry cases at compile-time, variable buffers for chemistry, and some early logic like EoS, reactions, and post_processing.
> * (**Minor**) Creation of `./mfc.sh spelling`.
> * (**Minor**) Multi-threaded `misc/viz.py`.
> * (**Minor**) Creation of `m_finite_differences`.
> * (**Minor**) Updated `indices.dat` to include the cons/prim variables, including mass fractions.
> * (**Minor**) Fixed `s_rectangle` patch for immersed boundaries. Note: I forget how I encountered this. Might have been with a test.
> * (**Minor**) Explicitly zero-out `rhs_vf` at the start of `s_compute_rhs`.
> * (**Minor**) Fixed the computation `time_avg` in `m_rhs` so that it works in 1D and 2D (not just 3D).
> * (**Minor**) Multi-threaded `pip install` if you specify `-j` (could speed up Cantera builds).
> * (**Minor**) Created `./mfc.sh spelling` (called by CI).
> 
> \dots and I think that's most of them. You can wait for the GT runners to come back up to merge this.

Fancy. Will do a little code review and merge once runners are up.

src/pre_process/m_data_output.fpp
```diff
@@ -767,15 +785,40 @@ contains
...
+            write (1, '(I3,A20,A20)') i, "\rho u_"//coord(i - momxb + 1), "u_"//coord(i - momxb + 1)
+        end do
+        do i = E_idx, E_idx
+            write (1, '(I3,A20,A20)') i, "\rho U", "p"
+        end do
+        do i = advxb, advxe
+            write (temp, '(I0)') i - contxb + 1
+            write (1, '(I3,A20,A20)') i, "\alpha_{"//trim(temp)//"}", "\alpha_{"//trim(temp)//"}"
+        end do
+        if (chemistry) then
```
sbryngelson, line 823: if there are physics involved or some other features then there could be other equations. subgrid bubbles is one such example.

src/simulation/m_rhs.fpp
```diff
@@ -1028,7 +1077,8 @@ contains
         end if
         call cpu_time(t_finish)
         if (t_step >= 4) then
-            time_avg = (abs(t_finish - t_start) + (t_step - 4)*time_avg)/(t_step - 3)
+            ncells = max((ix%end - ix%beg), 1)*max((iy%end - iy%beg), 1)*max((iz%end - iz%beg), 1)
```
sbryngelson: I actually already fixed this here: https://github.com/MFlowCode/MFC/blob/4c3e34def7726d004fe69f4fe35c5ef80eb2d0c8/src/simulation/m_start_up.fpp#L1155 

But since it's in a different module, this doesn't overwrite my fix and will give wrong answers. Please revert this change.

sbryngelson: My change is actually necessary to get grind times that make sense (using m_glb instead of local domain sizes), hence me re-doing the performance table.

henryleberre:
@sbryngelson I did but it needed some modifications to work with other compilers. I'm fixing the new errors now.

henryleberre:
@sbryngelson You haven't approved the PR yet. Let me know if you want anything changed.

src/simulation/m_global_parameters.fpp
```diff
@@ -525,6 +534,10 @@ contains
             teno = .false.
         #:endif
 
+        chem_params%advection = .true.
```
sbryngelson: shouldn't these all default to false so user has to activate them? also we don't usually put derived type assignments here

src/simulation/m_global_parameters.fpp
```diff
@@ -1035,6 +1048,16 @@ contains
             grid_geometry = 3
         end if
 
+        if (chemistry) then
+            chemistry_idx%beg = sys_size + 1
```
sbryngelson, line 1071: What if something else takes up this index? By setting `chemistry_idx%beg = sys_size + 1`, you are asserting that chemistry will be the last set variable (ever), so any additional physics cannot be added.

src/simulation/m_mpi_proxy.fpp
```diff
@@ -2152,19 +2226,24 @@ contains
         ! Send/Recv
         #:for rdma_mpi in [False, True]
             if (rdma_mpi .eqv. ${'.true.' if rdma_mpi else '.false.'}$) then
+                p_send => c_divs_buff_send(0)
+                p_recv => c_divs_buff_recv(0)
+
                 #:if rdma_mpi
```
sbryngelson: is `rdma_mpi` checked to be working in any of the tests for mi250x or nvidia GPUs? seems like it should be

src/common/include/macros.fpp
```diff
@@ -117,3 +117,19 @@
 
 #define t_vec3   real(kind(0d0)), dimension(1:3)
 #define t_mat4x4 real(kind(0d0)), dimension(1:4,1:4)
+
+#:def ASSERT(predicate, message = None)
```
sbryngelson, line 152: this seems perhaps useful but you only use it in a few places in the code (for chemistry). can you perhaps at least point out where else it should be used in the code so that is makes sense to keep?

src/common/include/macros.fpp
```diff
@@ -117,3 +117,19 @@
...
+    if (.not. (${predicate}$)) then
+        call s_mpi_abort("${_FILE_.split('/')[-1]}$:${_LINE_}$: "// &
+                         "Assertion failed: ${predicate}$. " &
+                         //${message or '"No error description."'}$)
+    end if
+#:enddef
+
+#:def PROHIBIT(predicate, message = None)
```
sbryngelson: same as the assert comment

src/simulation/m_global_parameters.fpp
henryleberre: I don't see how this is true. This is the pattern:

```fortran
if (feature) then
    feature_idx%beg = sys_size + 1
    feature_idx%end = sys_size + num_feature_equations
    sys_size = feature_idx%end
end if
```

This scales to any number of equations and new features. This is what some parts of the code already do, though not this concisely. We just keep increasing `sys_size` and assigning indices accordingly. Nothing is stopping you from creating variables after chemistry ones, just increment `sys_size` accordingly.

src/simulation/m_global_parameters.fpp
henryleberre: Sure, I don't mind these being false by default. Though, it could be confusing to new users that don't read the docs.

We usually don't have derived types that define "simulation algorithm parameters" but I think it's a good idea to modularize them like this. I might rename these in the future. `chem_params%advection` is readable and explicit in a case file but `chemistry%advection` would be even better. It's just that the `chemistry` input is a logical kill-switch right now.

src/common/include/macros.fpp
henryleberre: Sure, it can clean up so much code. Take this example adapted from `m_checker` (there are literally hundreds of lines like this):
```fortran
if (acoustic(j)%npulse >= 5 .and. acoustic(j)%dipole) then
    call s_mpi_abort('acoustic('//trim(jStr)//')%dipole is not '// &
                     'supported for support >= 5 '// &
                     '(non-planar supports). Exiting ...')
end if
```

This can be simplified to:
```
@:PROHIBIT(acoustic(j)%npulse >= 5 .and. acoustic(j)%dipole, 'for acoustic source '//trim(jStr))
```

This is what would get printed (I actually ran this):
```
m_checker.fpp:186: Prohibited condition: acoustic(j)%npulse >= 5 .and. acoustic(j)%dipole. for acoustic source 1
```

I think it's much better than our current solution. Less code and less duplication. Of course you can still make this example better but I hope this proves my point.

src/common/include/macros.fpp
henryleberre: Same as my answer for `@:PROHIBIT`. Assert is just the opposite of prohibit. Depending on the condition, it can be clearer to use one or the other.

src/simulation/m_mpi_proxy.fpp
henryleberre: If I ran the 2-rank test cases on Frontier and Wingtip with rdma_mpi ON, would that be sufficient for you?

src/simulation/m_global_parameters.fpp
sbryngelson: I guess it's of no consequence so long as one always checks if `chemistry` is `.true.` before invoking any of the `chem_params`. I guess you can check this in `m_checker` that the `chem_params` must all be `.false.` if `chemistry` is `.false.`. I'm not sure the best way to do that (how to check all derived type quantities at once).

src/simulation/m_global_parameters.fpp
sbryngelson: You are right. The only caution is that one should always invoke `feature_idx%beg` and `feature_idx%end` and never `sys_size-1` (or something like this). We used to do the latter on occasion but fixed it at some point. I don't think the code does that anywhere anymore.

src/simulation/m_mpi_proxy.fpp
sbryngelson: I'm not so much worried that this PR is incorrect, as that it would be useful to add `rdma_mpi == .true.` as a 2-rank test case on the phoenix/gpu and frontier runners (presuming one can use `rdma_mpi` on Phoenix GPU). Otherwise we have no check for all these chunks of code.

src/common/include/macros.fpp
sbryngelson: Yes very good. Can you open an issue for this to track this substitution as a code improvement? (presumably after merging this PR).

.github/workflows/frontier/build.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
 . ./mfc.sh load -c f -m g
-./mfc.sh build -j 8 --gpu
+./mfc.sh build -j 8 --gpu --sys-hdf5
```
sbryngelson: I wish this could somehow be included in `-c frontier` since it's such an obvious system specific invocation

src/common/include/macros.fpp
sbryngelson: same comment as for assert

src/simulation/m_mpi_proxy.fpp
sbryngelson: it actually wouldn't necessarily need to be 2-ranks either. it could be 1-rank with periodic BCs and mpi=true (I think).

.github/workflows/frontier/build.sh
sbryngelson: ... and I am going to forget about it very often

src/common/include/macros.fpp
henryleberre: #591 🚀

src/simulation/m_global_parameters.fpp
henryleberre: I see this as an implementation detail of chemistry that will be resolved in the full PR. Let me know if you disagree.

.github/workflows/frontier/build.sh
henryleberre: I agree. One option is to have `--sys-hdf5`, `--sys-fftw`, and `--sys-silo` be switches that are saved (remembered) like `--mpi` and `--gpu`. I don't think I need to fix it in this PR though.

src/simulation/m_mpi_proxy.fpp
henryleberre: This would indeed be nice but I don't think this PR should be the one to fix it.

henryleberre:
> I think some documentation may need to be updated. For example, the variables here: https://mflowcode.github.io/documentation/md_visualization.html

Since this PR adds the infrastructure for chemistry and does not claim to have working chemistry, I actively think we should refrain from documenting the existence of chemistry so as not to confuse users.

sbryngelson:
Phoenix is down, so I might have to wait to merge this. It looks like Frontier is still not working and it is interesting the Phoenix GPU test suite that finished before the servers went down showed a marked increase in performance.

sbryngelson:
@henryleberre I suppose this is up next, once conflicts are resolved? I merged #599

sbryngelson:
This is the first almost-mergable PR since #619. It's failing the CI cleanliness check, which is helpful, but in the diff it does find things like this (I was just perusing the "everything diff":
```
<   766 |         real(kind(0d0)) :: start, finish
---
>   261 |         real(kind(0d0)) :: start, finish
2781c2410
<   766 |         real(kind(0d0)) :: start, finish
---
>   261 |         real(kind(0d0)) :: start, finish
2820,2823d2448
```

We should probably change the workflow so that strip line numbers or some such before doing the diff. This is a real warning in the sense that those variables are probably not used, but it doesn't make sense to fail the check just because the user moved the subroutine down a few lines or something. @okBrian .

henryleberre:
@sbryngelson Do I need to pass the Cleanness CI to merge?

sbryngelson:
I'd like to see what's involved in passing this test (barring the silly diff warnings, like the line change I commented on above). if no one can actually pass the test, then it needs to be adjusted somehow or removed.

henryleberre:
@sbryngelson I'm sure I can remove the new warnings but a few are them are spurious since the testing is done with the `chemistry` parameter set to False. So everything that's behind `chemistry = T` causes warnings for being unused. Also, even if I got the new warnings to go away, the CI would still have to be updated to ignore the line number where the warnings were produced on.

sbryngelson:
ok then we can ignore it

PR #571, opened 2024-08-13
Add support for analytic definition of color function

wilfonba:
## Description

This PR adds support for analytic definition of the color function in the case file.

### Type of change

### Scope

## How Has This Been Tested?

Created an initial condition with an analytical color function


wilfonba:
This may not actually work as intended. I'll send a message when I verify it works.

wilfonba:
I can confirm this functions as expected. The problems I was seeing are an IO issue on a different branch where I implemented this same feature to accomplish something I needed.

PR #574, opened 2024-08-17
Fix runtime info dumping output

### LINKED ISSUE #504: Some `runtime_info` dumps to default fortran output file
Running a case I get a file called `fort.1`, which is a default fortran output file when an output name is not indicated. In this case it includes some runtime info:

```console
shb-m1pro-3: examples/1D_impact $ cat fort.1
--------------------------------------------------------------------------------

ICFL Max:  0.210141

Run-time: 0s
================================================================================
```

arjunj05:
runtime_info.txt

## Description

Issue #504 shows a default fortran file being outputted when running 1D_impact case. I renamed the output file from fortran.1 to runtime_info.txt, a more appropriate name.

Fixes #504

### Type of change

### Scope

## How Has This Been Tested?

- [x] Test A:  Running ./mfc.sh run examples/1D_impact/case.py shows the file runtime_info.txt file instead of fortran.1

**Test Configuration**:

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ examples/*/restart_data*
 examples/*/*.out
 examples/*/binary
 examples/*/fort.1
+examples/*/runtime_info.txt
 examples/*/*.sh
 examples/*/*.err
 examples/*/viz/
--- a/src/simulation/m_data_output.fpp
+++ b/src/simulation/m_data_output.fpp
@@ -1639,9 +1639,9 @@ contains
         real(kind(0d0)) :: run_time !< Run-time of the simulation
 
         character(len=*), parameter :: file_name = 'runtime_info.txt'
-        
+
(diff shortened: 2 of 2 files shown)


sbryngelson:
Shouldn't the file name be `run_time.inf`: https://github.com/MFlowCode/MFC/blob/4c4fb74324cabdafe6ea6bcaf993cf9a9c1842d7/src/simulation/m_data_output.fpp#L114 ?

Something doesn't seem right here. The subroutine you modified already assumes a file is open at I/O file id `(1)`, which should be `run_time.inf` (first argument to the `write` statement here: https://github.com/MFlowCode/MFC/blob/4c4fb74324cabdafe6ea6bcaf993cf9a9c1842d7/src/simulation/m_data_output.fpp#L1642).

Where does the `fort.1` file come from, and what does it contain?

PR #576, opened 2024-08-18
Fix adding footer output to runtimeinfo Fixes #574

arjunj05:
## Description

Orignally, the subroutine s_close_run_time_info, which intended to write a footer to run_time.inf file, was writing output to fort.1 file. This is because the run_time.inf was only being opened under unit 1 in the subroutine s_open_run_time_info. However, the subroutine s_save_performance_metrics(t_step, time_avg, time_final, io_time_avg, & io_time_final, proc_time, io_proc_time, file_exists, start, finish, nt) , which is called in between s_open_run_time_info and s_close_run_time_info, opened then closed unit 1. As a result, s_close_run_time_info was attempting to write to an unopened file, and was thus outputting to fort.1

Solution:
Reopened run_time.inf file under unit in s_close_run_time_info subroutine

Fixes #574 

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [x] Test A After running "./mfc.sh run examples/1D_impact/case.py" no fort.1 file is outputted and run_time.inf has footer.
- [ ] Test B

**Test Configuration**:

## Checklist

### DIFF
--- a/src/simulation/m_data_output.fpp
+++ b/src/simulation/m_data_output.fpp
@@ -1638,6 +1638,26 @@ contains
 
         real(kind(0d0)) :: run_time !< Run-time of the simulation
 
+        character(LEN=name_len) :: file_name = 'run_time.inf' !<
+            !! Name of the run-time information file
+
+        character(LEN=path_len + name_len) :: file_path !<
+            !! Relative path to a file in the case directory
+
+        logical :: file_exist !<
+            !! Logical used to check existence of run-time information file
+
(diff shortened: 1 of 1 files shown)


sbryngelson:
Can this not just be changed to a 3 line-of-code fix by opening the file with a different identifier (other than 1?). or, even easier, opening the performance metrics file in `s_open_run_time_info` with an identifier other than 1.

sbryngelson (approved):

Issue #457, opened 2024-06-07
remove magic numbers


sbryngelson:
The code doesn't have too many magic numbers https://en.wikipedia.org/wiki/Magic_number_(programming), but it has some, like this:

https://github.com/MFlowCode/MFC/blob/f347b30f9c0194c0ab3492b7772fa6edbc8d0860/src/simulation/m_qbmm.fpp#L1115C36-L1116

remove them (can set them as constant parameters in the module or subroutine).

sbryngelson:
I'd be curious to see this strategy continued: https://github.com/MFlowCode/MFC/pull/580/files#diff-3cbade161f91b38a72f4c35e4ab1a5fd66b24ab50d700e4e56eae364e5bd84ed

PR #580, opened 2024-08-20
remove magic numbers

LegitBaller0243:
## Description
Remove magic numbers issue: I updated the lines highlighted from #457, as well as two others (one was for creating a variable for standard atmospheric and the other was when the simulation checks if the grid is 3D) throughout the code. 

### Type of change

- [ ] Code cleaning

### Scope

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

## Checklist


src/pre_process/m_patches.fpp
```diff
@@ -29,6 +29,7 @@ module m_patches
     ! ==========================================================================
 
     implicit none
+    integer, parameter :: grid_3D = 3
```
sbryngelson, line 32: grid_geometry = 3 means 3D_cylindrical, not 3D (general)

sbryngelson (changes requested):
I find it hard to believe there were only a few magic numbers in the whole codebase

sbryngelson:
A bunch of magic numbers here: https://github.com/LegitBaller0243/MFC/blob/e719c1631b75296d7b97c869d56546e65fae212c/src/simulation/m_rhs.fpp#L1901-L1946

LegitBaller0243:
> A bunch of magic numbers here: https://github.com/LegitBaller0243/MFC/blob/e719c1631b75296d7b97c869d56546e65fae212c/src/simulation/m_rhs.fpp#L1901-L1946

Thank you for this and sounds good. I started with a couple, but needed a second opinion just to make sure I was on the right track. Will go back over code now.

henryleberre:
@sbryngelson I would argue that all the following are also _magic numbers_:
- The boundary condition types.
- The model_eqns.
- The regular patch types.
- The acoustic supports.
- The immersed boundary patch types.

I feel like they should be named (using enums) and we just use the name in the code instead of the integer value. They are documented in https://mflowcode.github.io/documentation/md_case.html so it should be straightforward.

henryleberre:
> > @sbryngelson I would argue that all the following are also _magic numbers_:
> > 
> > * The boundary condition types.
> > * The model_eqns.
> > * The regular patch types.
> > * The acoustic supports.
> > * The riemann solver types.
> > * The immersed boundary patch types.
> > * Probably many more
> >   …
> > 
> > I feel like they should be named (using enums) and we just use the name in the code instead of the integer value. They are documented in [mflowcode.github.io/documentation/md_case.html](https://mflowcode.github.io/documentation/md_case.html) so it should be straightforward.
> 
> I agree. I've thought about this before but never had the gumption to do something about it (breaking change of sorts).
> 
> What do you mean by
> 
> > I feel like they should be named (using enums)
> 
> .. Do you mean in the input files? If so, why is an enum necessary/useful here?

We could also name them in the input files but there isn't a very clean way of doing this. I was suggesting we could give them names in the Fortran code using a pattern a little like this:

m_patches.fpp:
```
module m_patches
    ...
    integer, parameter :: kLineSegmentPatch = 1
```

m_initial_condition.fpp:
```
if (patch_icpp(i)%geometry == kLineSegmentPatch) then
    call s_line_segment(i, patch_id_fp, q_prim_vf)
```

This is a lot more useful in places where the code is less readable like in the case of model_eqns. It also helps us keep better track of what integer maps to what type.

The alternative is to use [Fortran Enumerations](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-1/enumerations-and-enumerators.html).

sbryngelson:
Yes this seems like a good idea.

LegitBaller0243:
> > > @sbryngelson I would argue that all the following are also _magic numbers_:
> > > 
> > > * The boundary condition types.
> > > * The model_eqns.
> > > * The regular patch types.
> > > * The acoustic supports.
> > > * The riemann solver types.
> > > * The immersed boundary patch types.
> > > * Probably many more
> > >   …
> > > 
> > > I feel like they should be named (using enums) and we just use the name in the code instead of the integer value. They are documented in [mflowcode.github.io/documentation/md_case.html](https://mflowcode.github.io/documentation/md_case.html) so it should be straightforward.
> > 
> > 
> > I agree. I've thought about this before but never had the gumption to do something about it (breaking change of sorts).
> > What do you mean by
> > > I feel like they should be named (using enums)
> > 
> > 
> > .. Do you mean in the input files? If so, why is an enum necessary/useful here?
> 
> We could also name them in the input files but there isn't a very clean way of doing this. I was suggesting we could give them names in the Fortran code using a pattern a little like this:
> 
> m_patches.fpp:
> 
> ```
> module m_patches
>     ...
>     integer, parameter :: kLineSegmentPatch = 1
> ```
> 
> m_initial_condition.fpp:
> 
> ```
> if (patch_icpp(i)%geometry == kLineSegmentPatch) then
>     call s_line_segment(i, patch_id_fp, q_prim_vf)
> ```
> 
> This is a lot more useful in places where the code is less readable like in the case of model_eqns. It also helps us keep better track of what integer maps to what type.
> 
> The alternative is to use [Fortran Enumerations](https://www.intel.com/content/www/us/en/docs/fortran-compiler/developer-guide-reference/2024-1/enumerations-and-enumerators.html).


When I test the kLineSegmentPatch line, I keep getting this error: Error: Symbol 'klinesegmentpatch' at (1) has no IMPLICIT type. I followed the compiler instructions when setting up, and modules have use statements. What might be the reason causing this?

sbryngelson:
I don't know i don't even see the code

src/pre_process/m_patches.fpp
```diff
@@ -57,6 +58,30 @@ module m_patches
     real(kind(0d0)) :: x_centroid, y_centroid, z_centroid
     real(kind(0d0)) :: length_x, length_y, length_z
 
+    enum, bind(c)
+        enumerator :: kLineSegmentPatch = 1
```
sbryngelson, line 62: I'm unsure how this is different or helpful when compared to just declaring these as integer constant parameters in `common/m_constants.fpp`. @henryleberre had something in mind but honestly I'm not familiar with Fortran enumerations.

sbryngelson:
closing due to inactivity

PR #569, opened 2024-08-07
Two-stage compilation to allow cross-module inlining on NVHPC

AiredaleDev:
## Description

Please include a summary of the changes and the related issue(s) if they exist.
Please also include relevant motivation and context.

Fixes #328 

### Type of change

Please delete options that are not relevant.

- [ ] New feature (non-breaking change which adds functionality)

### Scope

- [ ] This PR comprises a set of related changes with a common goal

If you cannot check the above box, please split your PR into multiple PRs that each have a common goal.

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce. 
Please also list any relevant details for your test configuration

- [ ] Load NVHPC Modules on your cluster of choice
- [ ] Build MFC with GPU support
- [ ] Observe that performance remains unchanged!

**Test Configuration**:

* What computers and compilers did you use to test this:
* Rogue's Gallery -- Uwing (GH200)
* Phoenix -- 1xV100 node

In the process of implementing this, I discovered some quirks that may be worth addressing in the future:
* For some reason, CMake will re-link the "lib" targets on each run of the test suite.
* Building with `--unified` causes the compiler to remark `Error 2`.
* Blindly rebasing this with `master` causes the compiler to remark `Error 2`.

## Checklist

- [ ] I ran `./mfc.sh format` before committing my code
- [ ] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled
- [ ] This PR does not introduce any repeated code (it follows the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
- [ ] I cannot think of a way to condense this code and reduce any introduced additional line count

### If your code changes any code source files (anything in `src/simulation`)

To make sure the code is performing as expected on GPU devices, I have:
- [ ] Checked that the code compiles using NVHPC compilers

PR #581, opened 2024-08-20
Two-stage IPO for NVHPC

### LINKED ISSUE #328: Link-time optimization and inlining
A possible fix for the performance degradation seen on NVHPC when calling subroutines across modules: https://forums.developer.nvidia.com/t/nvhpc-23-11-fortran-does-it-inline-public-subroutines-across-modules/281047/2

> Though for cross-file inlining you can try the two-pass method. First create an inline extract library but compiling all files with “-Mextract=lib:libname” replacing “libname” with what you’d like to call it. Then compile with “-Minline=lib:libname” to use the extract library. Inlining is performed prior to the device code generation.

henryleberre:
See #569, fixes #328. I cleaned up the code and the git history from that PR.


sbryngelson:
Would it be possible to move most or all macros in https://github.com/MFlowCode/MFC/blob/master/src/simulation/include/inline_riemann.fpp to subroutines, now?

henryleberre:
@sbryngelson I tried doing the others yesterday but got some issues on Frontier. For now, I think we should just merge this PR and remove the other macros with the other frontier hacks.

sbryngelson:
> @sbryngelson I tried doing the others yesterday but got some issues on Frontier. For now, I think we should just merge this PR and remove the other macros with the other frontier hacks.

Yes this is fine with me. Just waiting on it to be 'working'. That is, passing CI with good benchmark results.

sbryngelson (approved):

PR #589, opened 2024-08-24
Benchmarks report grind time

henryleberre:
## Description

Benchmarks now report grind time for `simulation`, not just execution time. Varying IO times on GT Phoenix make it difficult to judge whether a PR makes the code slower or faster based on execution time alone.


sbryngelson (approved):

sbryngelson:
Awesome. Looks good.

PR #594, opened 2024-08-29
Code Cleanness CI

### LINKED ISSUE #467: CI Check for code cleanness
Can add a CI check for the _introduction_ of code uncleanness by diffing an MFC build on a PR and on master, both with an appended `-Wall` flag to the `GNU` compile options in `CMakeLists.txt`, piping the output as `./mfc.sh build > out.txt 2>&1`, and printing and counting the warnings via `grep '\-W' out.txt`. Then deleting common lines.

okBrian:
## Description

Code Cleanness CI that has specific parts for unused variables, values, dummy arguments, maybe-uninitialized

Fixes #467 

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- Github Actions on the repository

**Test Configuration**:

* Github Actions

## Checklist


henryleberre (approved):
:rocket:

sbryngelson:
Thanks, @okBrian! I will look at this more closely soon. If @henryleberre thinks the PR looks good, then it's probably good. 

Since you've already been doing a lot of CI work, I think it might be useful to look at some of these code "uncleanness" situations in the Fortran source code and squash some of those warnings that the CI shows. There are many, so I don't recommend doing all of them at once.

sbryngelson (changes requested):
This should have a summary step at the end that greps the grep for "Warning" to count how many warnings there are so we can keep track, since there's no way to squash all of them. and presumably how many warnings there were in the current master branch with a diff between the PR and master. and perhaps also something to show the new warnings so we can see what was added that threw a warning.

okBrian:
The PR should be working the difference is large I think because master doesn't have all warnings enabled on --debug build, so when I build with pr its adding a lot of warnings.

sbryngelson:
it looks like you need to rebase..

sbryngelson:
> The PR should be working the difference is large I think because master doesn't have all warnings enabled on --debug build, so when I build with pr its adding a lot of warnings.

You would need this PR to build `master` with the warning flags for comparison purposes. 

I suppose the problem is that you don't have the option to add at the moment those because they are not available in the current master branch `CMakeLists.txt`? If so, you can first make a small PR that enables them if a certain flag is used, which we can merge right away, then we can check how this one is working.

sbryngelson:
We might need an additional flag other than `--debug`. In many cases I can imagine that users invoking `--debug` do not want to see all of these warnings! (though I'm open to other views @henryleberre @okBrian ?)

okBrian:
Honestly I am just not sure where else we would put warn all. I feel like making another flag to just show all warnings may be a bit awkward and I feel like grouping it under debug makes sense because you want to be warned of all potential things that could cause the error in your code. Even though you do get a lot of warnings you can grep through it to find a specific warning.

sbryngelson:
> Honestly I am just not sure where else we would put warn all. I feel like making another flag to just show all warnings may be a bit awkward and I feel like grouping it under debug makes sense because you want to be warned of all potential things that could cause the error in your code. Even though you do get a lot of warnings you can grep through it to find a specific warning.

Fair enough. Other opinions, @wilfonba @henryleberre etc.?

sbryngelson:
Update: I just ran `--debug` from @okBrian's fork and think it's OK to leave these warnings in `--debug`

sbryngelson:
I think you need to pull from master @okBrian 

From CI:
```console
Difference is how many warnings were added or removed from master to pr, negative numbers are better since you are removing warnings
Unused Variable Count: 379, Difference: 377
Unused Dummy Argument: 185, Difference: 185
Unused Value: [24](https://github.com/MFlowCode/MFC/actions/runs/10707730293/job/29688381847?pr=594#step:11:25), Difference: 24
Maybe Uninitialized: 8, Difference: 7
Everything: 1485, Difference: 619
```

henryleberre:
> > Honestly I am just not sure where else we would put warn all. I feel like making another flag to just show all warnings may be a bit awkward and I feel like grouping it under debug makes sense because you want to be warned of all potential things that could cause the error in your code. Even though you do get a lot of warnings you can grep through it to find a specific warning.
> 
> Fair enough. Other opinions, @wilfonba @henryleberre etc.?

I agree with @okBrian. I am also a little skeptical about the utility of this workflow but I am open to trying it. The approach I think makes more sense is to enable all the warnings that can be important (unused variable, use before declaration, etc..) in both Release/Debug and have the CI fail if there are any warnings. We would then require that there be no such warnings before merging a PR. Think of it like linting for Python code.

sbryngelson:
> > > Honestly I am just not sure where else we would put warn all. I feel like making another flag to just show all warnings may be a bit awkward and I feel like grouping it under debug makes sense because you want to be warned of all potential things that could cause the error in your code. Even though you do get a lot of warnings you can grep through it to find a specific warning.
> > 
> > 
> > Fair enough. Other opinions, @wilfonba @henryleberre etc.?
> 
> I agree with @okBrian. I am also a little skeptical about the utility of this workflow but I am open to trying it. The approach I think makes more sense is to enable all the warnings that can be important (unused variable, use before declaration, etc..) in both Release/Debug and have the CI fail if there are any warnings. We would then require that there be no such warnings before merging a PR. Think of it like linting for Python code.

It is an interesting idea, but a problem could be that while we have a way of disabling linting checks locally in some Python code regions, which seems necessary for the toolchain, we don't have this for Fortran (and probably wouldn't want it?). 

If you look at the output of @okBrian's PR CI 'cleanness' run, you'll see many warnings. Most are legitimate and should be cleaned up. But I worry that our use of metaprogramming and `ifdef`s results in warnings in some cases (e.g., GPU compilation, or only for `simulation` or `pre_process`) that cannot be 'cleaned.' I would need to look closer to verify this is wholly true (there is no way to remove the warning), but keep it in mind.

sbryngelson:
This looks "right" now. @okBrian can you confirm this to be working as you expect?

@henryleberre @okBrian how should we converge on failure/pass of the CI run? My current opinion is that 'cleanliness' CI should fail (non-zero stop code) if the diff in the # of warnings between pr and master is > 0. If this is the case, we can still merge, but we would at least be aware of it. Asserting the # warnings = 0 seems impossible to maintain (though not a bad target).

sbryngelson:
oh and @okBrian you still have 300 files changed in this PR. can you rebase (or whatever is needed) to sync your changes (backup your work if you do!)? alternatively you can open a new PR that has only the needed changes, unclear to me which is more work.

okBrian:
I am going to remake this PR since it'll probably be easiest way to fix the pr.

PR #596, opened 2024-08-31
CCE 18 on Frontier & Benchmarking fixes among others

henryleberre:
Attempting to upstream the bits from #544 that do not hinder performance.


sbryngelson:
Ugh, waiting on Phoenix GPU nodes. I will look more closely at that this afternoon and see what's up.

sbryngelson (approved):

PR #599, opened 2024-08-31
Remove build hacks that are no longer needed

henryleberre:
- Removed CCE 16 workgroup hacks. 
- ~~Removed Silo patch for HDF5 on Frontier.~~
- Removed AppleClang workarounds on macOS, including building MPI from scratch.
- Only build HDF5, FFTW, and SILO packages if they are not installed/compatible.
- Removed Docker (nobody uses it, and maintaining it is cumbersome).
- Switched to using the proper Silo patch for an error on GNU 13 and above. See my comment here: https://github.com/LLNL/Silo/issues/375#issuecomment-2323029131.
- Removed installation instructions for Arch Linux. They installed their operating system, I'm sure they can install gcc.
- ~~Removed our FindHDF5 file. We can use the built-in.~~


README.md
sbryngelson, line 1: Can't you remove the `export [...]` line from the readme?

henryleberre: Yes I'm aware the docs still have issues. I'm figuring out what is needed and what isn't partially thanks to CI. It looks like `export FC=gfortran-14` might be necessary on macOS. It isn't on my machine but it fails without it in CI.

sbryngelson: 👍

sbryngelson:
[TODO] Check FOM is roughly the same w/o workaround for workgroup size on CCE18.

henryleberre:
> > * Removed installation instructions for Arch Linux. They installed their operating system, I'm sure they can install gcc.
> 
> 🤣 🤣 🤣 You might still leave them a list of packages they need. MFC's dependencies are not obvious other than from our installation instructions.

The assumption is that they can look at those needed on Debian as reference. The only one that isn't obvious is `gcc-fortran` but solving that is one error and a google search.

Issue #206, opened 2023-08-12
Run a test case again if it fails


sbryngelson:
Right now, the test suite aborts if a test fails (there is a 'relentless mode' where it powers through everything, but I'm not talking about this, and it requires a flag). In some cases, a test fails spuriously on CI, and we need to re-run the CI. Notably, this happens on MacOS runners, but I presume it happens in other cases. 

It would be a nice feature (I think) if we attempt to re-run a case if it fails, perhaps immediately after it fails, with a message saying `'Case XXXXXX failed, attempting one more time'`, then attempts the case again, before exit 1'ing. I do not know if this will fix the GitHub runner issue, but it seems useful and easy.

PR #605, opened 2024-09-03
Relentless fix

### LINKED ISSUE #597: Potential improvement to `./mfc.sh test`
It seems like `./mfc.sh test` should be relentless by default (via the existing `-r` flag) and report the number of failures, passes, and skips at the end of testing. This is how most software libraries do tests. This would involve changing the default `test` behavior and a new exit code for failure that counts test failures in `-r` mode.

It might also be useful to set the default to 0 retries (the default is 3) and only use 3 retries when in CI mode (this would require a new flag).

aricer123:
## Description

Fixes #597  

#New feature (non-breaking change which adds functionality)

Made relentless by default
Added ci flag 
Now prints metrics of test after testing is over(successes, failures, skips)

- [X] This PR comprises a set of related changes with a common goal

**Test Configuration**:

Tested on my personal computer


toolchain/mfc/args.py
```diff
@@ -181,3 +182,4 @@ def add_common_arguments(p, mask = None):
             args[e] = os.path.abspath(args[e])
 
     return args
+git clone https://github.com/aricer123/MFC.git
```
sbryngelson: ?

toolchain/mfc/args.py
```diff
@@ -81,10 +81,11 @@ def add_common_arguments(p, mask = None):
...
     test.add_argument("-t", "--to",           default=test_cases[-1].get_uuid(), type=str, help="Last test UUID to run.")
     test.add_argument("-o", "--only",         nargs="+", type=str, default=[], metavar="L", help="Only run tests with specified properties.")
-    test.add_argument("-r", "--relentless",   action="store_true", default=False, help="Run all tests, even if multiple fail.")
+    test.add_argument("-r", "--relentless",   action="store_true", default=True, help="Run all tests, even if multiple fail.")
     test.add_argument("-a", "--test-all",     action="store_true", default=False, help="Run the Post Process Tests too.")
     test.add_argument("-%", "--percent",      type=int, default=100, help="Percentage of tests to run.")
-    test.add_argument("-m", "--max-attempts", type=int, default=3, help="Maximum number of attempts to run a test.")
+    test.add_argument("-i", "--ci-mode",      action="store_true", default=False, help="Toggles CI mode.")
+    test.add_argument("-m", "--max-attempts", type=int, default=0, help="Maximum number of attempts to run a test.")
```
sbryngelson: default should be 1 perhaps?

toolchain/mfc/test/test.py
```diff
@@ -217,27 +213,31 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
...
                 cons.print(f"[bold yellow] Attempt {nAttempts}: Failed test {case.get_uuid()}. Retrying...[/bold yellow]")
                 continue
 
             nFAIL += 1
-
             cons.print(f"[bold red]Failed test {case} after {nAttempts} attempt(s).[/bold red]")
 
             if ARG("relentless"):
                 cons.print(f"{exc}")
+                nSKIP += 1
```
sbryngelson: this is unclear to me. it appears you are skipping every test case in `-r` mode? also at the moment it is unclear to me if we have any cases that are "skipped." a case I can think of is if a package is unavailable so then we skip the test for the case, but for now we have not had this (so may be worth removing?)

aricer123:
Hello Professor Brygelson, 
I noticed it failed a lot of cases. Looks like while i was creating a new branch and preparing to submit the PR, I had accidentally copied one of the commands into the file args.py This was after I finished testing so I missed it. I just updated this, could you check it again now?

toolchain/mfc/test/test.py
sbryngelson: Or I suppose this is saying that if we are in relentless mode, all failed cases are actually skipped cases? 

If so, this is not what I intended. If we have nfail > 0, we report this # at the end, then exit with a non-0 exit code.

toolchain/mfc/test/test.py
```diff
@@ -18,6 +18,8 @@
 
 
 nFAIL = 0
+nPASS = 0
+nSKIP = 0
```
sbryngelson, line 22: Even if it's unclear how `nSKIP` will be used (I don't see any test cases right now that we will skip, but I could imagine some), it seems like we should keep this 'feature'

sbryngelson (changes requested):

toolchain/mfc/args.py
aricer123: https://github.com/MFlowCode/MFC/pull/605#issuecomment-2326927176
Resolved in the most recent commit

sbryngelson:
You have unresolved conversations @aricer123

toolchain/mfc/test/test.py
aricer123: Addressed in most recent commit.

toolchain/mfc/args.py
aricer123: Addressed

toolchain/mfc/args.py
```diff
@@ -81,10 +81,11 @@ def add_common_arguments(p, mask = None):
     test.add_argument("-f", "--from",         default=test_cases[0].get_uuid(), type=str, help="First test UUID to run.")
     test.add_argument("-t", "--to",           default=test_cases[-1].get_uuid(), type=str, help="Last test UUID to run.")
     test.add_argument("-o", "--only",         nargs="+", type=str, default=[], metavar="L", help="Only run tests with specified properties.")
-    test.add_argument("-r", "--relentless",   action="store_true", default=False, help="Run all tests, even if multiple fail.")
+    test.add_argument("-r", "--relentless",   action="store_true", default=True, help="Run all tests, even if multiple fail.")
```
henryleberre: This is just like removing `--relentless`. That is, there is no way of turning it off through the CLI. We have two options:
- Remove the option entirely.
- Rename it to `--no-relentless` (or some better name) and change action to `store_false` and set default back to `False`.

I think the first option is best. Failing quickly can be useful to CI but I think it's cleaner not to have it right now.

toolchain/mfc/args.py
```diff
@@ -81,10 +81,11 @@ def add_common_arguments(p, mask = None):
...
     test.add_argument("-a", "--test-all",     action="store_true", default=False, help="Run the Post Process Tests too.")
     test.add_argument("-%", "--percent",      type=int, default=100, help="Percentage of tests to run.")
-    test.add_argument("-m", "--max-attempts", type=int, default=3, help="Maximum number of attempts to run a test.")
+    test.add_argument("-i", "--ci-mode",      action="store_true", default=False, help="Toggles CI mode.")
```
henryleberre: Why have `--ci-mode` when you can just pass `--max-attempts 3` ?

toolchain/mfc/test/test.py
```diff
@@ -113,25 +111,19 @@ def test():
...
-    if nFAIL == 0:
-        cons.print("Tested Simulation [bold green]✓[/bold green]")
-    else:
-        raise MFCException(f"Testing: Encountered [bold red]{nFAIL}[/bold red] failure(s).")
-
     cons.unindent()
+    cons.print(f"\nTest Summary: [bold green]{nPASS}[/bold green] passed, [bold red]{nFAIL}[/bold red] failed, [bold yellow]{nSKIP}[/bold yellow] skipped.")
+   
+    if nFAIL > 0:
+        exit(nFAIL)
```
henryleberre: You can just remove the if-else entirely and just `exit(nFail)`.

toolchain/mfc/test/test.py
```diff
@@ -215,25 +208,27 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
...
-    # pylint: disable=global-statement
-    global nFAIL
+    global nFAIL, nPASS, nSKIP
 
     nAttempts = 0
-
+    maxAttempts = ARG("max_attempts")
+    
+    if ARG("ci_mode"):
+        maxAttempts = 3
```
henryleberre: See previous comment about `--ci-mode`.

henryleberre (changes requested):

toolchain/mfc/test/test.py
aricer123: I added it because in the issue it said "It might also be useful to set the default to 0 retries (the default is 3) and only use 3 retries when in CI mode (this would require a new flag)."
Should I keep this or remove it? I have removed it for now in the lastest commit.

henryleberre: Right, but as this currently is, there is a semantic gap between `--ci-mode` and `--max-retries`. It isn't obvious that these options would be coupled. For example, what should the behavior be if you pass both `--ci-mode` and `--max-retries 10` ? The best is just to default `--max-retries` to 0 and let the CI runners explicitly pass `--max-retries` to 3 if need be.

Actually, I believe this feature is no longer necessary. It was introduced to hide spurious failures in the macOS runner but I'm pretty sure the issue has gone away (or will in #599). You can keep for now if you don't want to do testing to be sure the original problem has gone away.

henryleberre: Well, we can collectively blame @sbryngelson for this :)

henryleberre: In engineering, always question the requirements - and authority. I also just realized I kept writing `--max-retries` instead of `--max-attempts`.. my bad!

aricer123: I removed it anyway. Everything should be up to date now.

toolchain/mfc/args.py
sbryngelson: good point, somehow this didn't even cross my mind. but I still agree with you. users can ctrl+c if they see failure and want to quit, as well.

toolchain/mfc/args.py
sbryngelson: good catch, again. I would recommend just having max-attempts and getting rid of ci-mode, which seems a bit silly.

toolchain/mfc/test/test.py
sbryngelson: @henryleberre having 3 retries has actually saved us in CI _many_ times. It also comes up on Frontier (or used to), for mysterious reasons. @aricer123 I recommend keeping `--max-attempt=3` in CI.

henryleberre (approved):

sbryngelson:
@aricer123 you need to update the CI runner workflows to use `--max-retries=3`.

henryleberre:
@aricer123 You can run `./mfc.sh lint` locally to help you make the Lint Toolchain workflow pass.

henryleberre:
> @aricer123 you need to update the CI runner workflows to use `--max-retries=3`.

He probably doesn't.. As I mentioned in more detail in one of above the discussions, this was created to hide spurious MPI failures in the macOS CI. These probably got fixed some time ago (library upgrades or GH updates) or will probably be irrelevant after #599. If the test suite passes without it, we are probably fine.

sbryngelson:
> > @aricer123 you need to update the CI runner workflows to use `--max-retries=3`.
> 
> He probably doesn't.. As I mentioned in more detail in one of above the discussions, this was created to hide spurious MPI failures in the macOS CI. These probably got fixed some time ago (library upgrades or GH updates) or will probably be irrelevant after #599. If the test suite passes without it, we are probably fine.

My comment from a closed convo.: @henryleberre having 3 retries has actually saved us in CI _many_ times. It also comes up on Frontier (or used to), for mysterious reasons. @aricer123 I recommend keeping `--max-attempt=3` in CI.

henryleberre:
> > > @aricer123 you need to update the CI runner workflows to use `--max-retries=3`.
> > 
> > 
> > He probably doesn't.. As I mentioned in more detail in one of above the discussions, this was created to hide spurious MPI failures in the macOS CI. These probably got fixed some time ago (library upgrades or GH updates) or will probably be irrelevant after #599. If the test suite passes without it, we are probably fine.
> 
> My comment from a closed convo.: @henryleberre having 3 retries has actually saved us in CI _many_ times. It also comes up on Frontier (or used to, mostly on the 2-rank case), for mysterious reasons. @aricer123 I recommend keeping `--max-attempt=3` in CI.
> 
> Update: I suppose I'm open to removing it _later_, but for now I would like to retain "default" behavior in CI?
> 
> Update 2: Also, why do you think your PR #599 is relevant, do you think that's why it was failing? I suppose it's possible but really I have no idea what was going on.

I stand corrected.
I thought #599 might be relevant because it removes some of the MPI hacks. Admittedly, we no longer

henryleberre:
We> > > @aricer123 you need to update the CI runner workflows to use `--max-retries=3`.
> > 
> > 
> > He probably doesn't.. As I mentioned in more detail in one of above the discussions, this was created to hide spurious MPI failures in the macOS CI. These probably got fixed some time ago (library upgrades or GH updates) or will probably be irrelevant after #599. If the test suite passes without it, we are probably fine.
> 
> My comment from a closed convo.: @henryleberre having 3 retries has actually saved us in CI _many_ times. It also comes up on Frontier (or used to, mostly on the 2-rank case), for mysterious reasons. @aricer123 I recommend keeping `--max-attempt=3` in CI.
> 
> Update: I suppose I'm open to removing it _later_, but for now I would like to retain "default" behavior in CI?
> 
> Update 2: Also, why do you think your PR #599 is relevant, do you think that's why it was failing? I suppose it's possible but really I have no idea what was going on.

I stand corrected. We used to build OpenMPI from scratch on macOS (which was very hacky). Then we switched to `mpich` which hasn't been working great for everyone. Since #599 will make use of regular OpenMPI (from HomeBrew) it is likely we won't see these failures on mac again.

sbryngelson:
> I stand corrected. We used to build OpenMPI from scratch on macOS (which was very hacky). Then we switched to `mpich` which hasn't been working great for everyone. Since #599 will make use of regular OpenMPI (from HomeBrew) it is likely we won't see these failures on mac again.

I remember this differently but certainly have no way of proving it. I suppose I'd like to see this PR keep the max-retries feature just so that I know this specific PR did not in any way "break" our CI. Then we can try removing it in a separate PR and seeing if that trips up future CI runs.

sbryngelson (changes requested):
add `--max-atempts=3` to relevant CI workflows

henryleberre:
> > I stand corrected. We used to build OpenMPI from scratch on macOS (which was very hacky). Then we switched to `mpich` which hasn't been working great for everyone. Since #599 will make use of regular OpenMPI (from HomeBrew) it is likely we won't see these failures on mac again.
> 
> I remember this differently but certainly have no way of proving it. I suppose I'd like to see this PR keep the max-retries feature just so that I know this specific PR did not in any way "break" our CI. Then we can try removing it in a separate PR and seeing if that trips up future CI runs.

I might be misremembering. I tracked down the original issue to #206 and you make no reference to macOS. I distinctly remember, however, that the issue where we had to rerun workflows upon spurious was particularly recurrent on macOS. I we are both right, then, in some limited capacity.

henryleberre:
@sbryngelson Can I pre-order the first volume of "A history of MFC: Conflicting and erroneous accounts" ?

sbryngelson:
> > > I stand corrected. We used to build OpenMPI from scratch on macOS (which was very hacky). Then we switched to `mpich` which hasn't been working great for everyone. Since #599 will make use of regular OpenMPI (from HomeBrew) it is likely we won't see these failures on mac again.
> > 
> > 
> > I remember this differently but certainly have no way of proving it. I suppose I'd like to see this PR keep the max-retries feature just so that I know this specific PR did not in any way "break" our CI. Then we can try removing it in a separate PR and seeing if that trips up future CI runs.
> 
> I might be misremembering. I tracked down the original issue to #206 and you make no reference to macOS. I distinctly remember, however, that the issue where we had to rerun workflows upon spurious failures was particularly recurrent on macOS. We are both right, then, in some limited capacity.

It says 

> Notably, this happens on MacOS runners, but I presume it happens in other cases.

Which I did remember. I guess I just want to make sure the MacOS runner (and Frontier runners) are fine before taking out the flag.

sbryngelson:
> @sbryngelson Can I pre-order the first volume of "A history of MFC: Conflicting and erroneous accounts" ?

We are really on a roll with your last gaff 😆

henryleberre:
I hereby retract my correction claiming you were partially right.

sbryngelson:
> I hereby retract my correction claiming you were partially right.

one for the ages

aricer123:
So the default behavior of max-retries should be 3?
Should I bring back the CI flag?

sbryngelson:
> So the default behavior of max-retries should be 3? Should I bring back the CI flag?

You can just use the --max-attempts=3 flag. I do not see how the CI flag is useful unless it combines many/multiple flags for brevity (which it does not).

henryleberre:
> So the default behavior of max-retries should be 3? Should I bring back the CI flag?

You just have to modify these lines:
- https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/test.yml#L97
- https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/test.yml#L112
- https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/frontier/test.sh#L6
- https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/frontier/test.sh#L6

to add `--max-attempts 3` to the command-line invocation of `./mfc.sh test`.

sbryngelson:
> > So the default behavior of max-retries should be 3? Should I bring back the CI flag?
> 
> You just have to modify these lines:
> 
> * https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/test.yml#L97
> * https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/test.yml#L112
> * https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/frontier/test.sh#L6
> * https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/.github/workflows/frontier/test.sh#L6
> * I think that's all of them...
> 
> to add `--max-attempts 3` to the command-line invocation of `./mfc.sh test`.

Phoenix test?

henryleberre:
@sbryngelson Good catch, just made the correction.

aricer123:
Thanks!
Not fully familiar with the project structure yet so it's hard to find what to change and where.
Am updating right now.

aricer123:
all changes completed

sbryngelson:
it's failing CI...

aricer123:
its failing only this Lint / Lint Toolchain (pull_request) 
@henryleberre what should I do

sbryngelson:
Please fix one issue per PR when possible (I see you are making commits regarding RDMA).

sbryngelson:
it looks like Phoenix is undergoing some maintenance, will merge once that is back up.

sbryngelson (dismissed):

sbryngelson:
Something is not correct. It reports 0 passes, 0 fails, and 0 skips no matter what:
* (GPU Phoenix) https://github.com/MFlowCode/MFC/actions/runs/10694192035/job/29690089123?pr=605#step:6:13775
* (CPU Phoenix) https://github.com/MFlowCode/MFC/actions/runs/10694192035/job/29690088560?pr=605#step:6:5045
* (A github runner) https://github.com/MFlowCode/MFC/actions/runs/10694192035/job/29690090256?pr=605#step:8:372

henryleberre:
@sbryngelson I looked into the Benchmark one and it failed (I think) because my latest PR. They share the same build folder and that created issues. So, if I were you, I would `rm -r` the build folders on Phoenix.

aricer123:
I think it should report passes now

sbryngelson:
@aricer123 how did you confirm that this fails CI if a test fails?

aricer123:
I've been testing so far by running the full test suite on my mac.
I tested failure by introducing a bug and seeing what the output would be if it failed.

henryleberre:
I see the Benchmarking CI ran to completion this time. Nice. This actually gives me a clue to solve the issue in my current PR.

sbryngelson:
> I see the Benchmarking CI ran to completion this time. Nice. This actually gives me a clue to solve the issue in my current PR.

It has failed spuriously on a few different PRs today. I think it was in part due to a problem with Phoenix that was resolved, but am not sure.

henryleberre:
> > I see the Benchmarking CI ran to completion this time. Nice. This actually gives me a clue to solve the issue in my current PR.
> 
> It has failed spuriously on a few different PRs today. I think it was in part due to a problem with Phoenix that was resolved, but am not sure.

The spurious nature is likely due to it requiring to be assigned the same runner that handled that one PR of mine. If those failures today were happening while installing HDF5, it was most likely my fault (although this cross-PR contamination should not exist in the first place...).

sbryngelson:
> > They weren't happening while installing HDF5, some failed randomly and some didn't even give error messages (just job failure). I logged into Phoenix and saw that all the RHEL9 nodes were down for maintenance, and got an email an hour later saying that a problem was resolved. So I don't really have clear evidence that there is runner cross-pollination happening or that, if there is, it's causing a problem.
> 
> Interesting. I can confirm there is evidence of cross-pollination (or at least sharing of the build directory) because there were two `build/benchmarks/<benchmark slug>` directories in this run that I saw failed. We should only create one benchmark slug per benchmark (one in each clone of master and the PR branch)

This could perhaps happen if a run fails, then we re-attempt it? This seems like one logical conclusion, though I still don't see why it doesn't do a fresh clone upon starting the re-attempt. Each invocation of CI clone should clone a clean MFC.git, and there's no way for a runner to put data into a different runner's `_work` directory (that I could possibly imagine).

aricer123:
Are there any additional tests you want me to run or tasks I could to do to help resolve this issue?

sbryngelson (approved):

PR #606, opened 2024-09-04
Add rdma

### LINKED ISSUE #598: Add `rdma_mpi` to test suite
Add `rdma_mpi = 'T'` to test suite. This only gets turned on in OpenACC (GPU) mode, which only exists on the Phoenix GPU runner and the Frontier runner, both of which support it.

Note: There needs to be a way to skip this test if the compiler does not support it (in non-CI cases).

aricer123:
## Description

Fixes #598 

###
Adds flag for rdma for testing and passes to ./mfc sh run
Compiles on macbook and runs tests but needs to be tested on gpu with rdma, may have errors

### DIFF
--- a/.github/workflows/frontier/test.sh
+++ b/.github/workflows/frontier/test.sh
@@ -3,4 +3,4 @@
 gpus=`rocm-smi --showid | awk '{print $1}' | grep -Eo '[0-9]+' | uniq | tr '\n' ' '`
 ngpus=`echo "$gpus" | tr -d '[:space:]' | wc -c`
 
-./mfc.sh test -j $ngpus --sys-hdf5 --sys-fftw -- -c frontier
+./mfc.sh test -j $ngpus --sys-hdf5 --sys-fftw -- -c frontier -- --rdmi
--- a/.github/workflows/phoenix/test.sh
+++ b/.github/workflows/phoenix/test.sh
@@ -16,5 +16,5 @@ if [ "$job_device" == "gpu" ]; then
     n_test_threads=`expr $gpu_count \* 2`
 fi
 
-./mfc.sh test -a -j $n_test_threads $device_opts -- -c phoenix
+./mfc.sh test -a -j $n_test_threads $device_opts -- -c phoenix -- --rdmi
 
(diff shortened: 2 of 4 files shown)


sbryngelson (changes requested):
> Adds flag for rdma for testing and passes to ./mfc sh run

This PR indeed does this but does not do anything differently within the `./mfc.sh run` environment. It should ensure the `RDMA_MPI = 'T'` in the generated `case.py` file.

henryleberre:
The changes in the `toolchain` folder can be undone. See https://github.com/MFlowCode/MFC/issues/598#issuecomment-2322997011. 

To clarify, when you `./mfc.sh test -- -c phoenix -- --rdma` this is what happens when we follow the execution of a single test:
- We generate a `case.py` file in the `tests/<case slug>/` folder.
- We execute `./mfc.sh run tests/<case slug>/case.py -c phoenix -- --rdma`. There are more arguments but you don't need to worry about these.
- When `./mfc.sh run` executes, it basically runs the `tests/<case slug>/case.py` file and passes the `--rdma` flag.

So there is no need to have the `--rdma` file registered with `./mfc.sh test` or `./mfc.sh run`. The case file that is autogenerated should be the one that parses this argument.

aricer123:
> * We generate a `case.py` file in the `tests/<case slug>/` folder.

Where in the project/program is the case.py file generated? From my understanding each case.py contains the configurations for its respective test/example and is manually written.

I understand the rest of how the test and run features work as well as how the arguments including the case file are passed.

henryleberre:
> > * We generate a `case.py` file in the `tests/<case slug>/` folder.
> 
> Where in the project/program is the case.py file generated? From my understanding each case.py contains the configurations for its respective test/example and is manually written.
> 
> I understand the rest of how the test and run features work as well as how the arguments including the case file are passed.

This is where we declare the parser of the case.py files used for testing: https://github.com/MFlowCode/MFC/blob/0c345eac0079a873f254f2fe49cf6f838ed96cad/toolchain/mfc/test/case.py#L182. You need to add the `--rdma` flag and make it set `rdma_mpi={'T' if rdma else 'F'}` in the case dictionary printed by the case file.

aricer123:
got it thanks, i missed that when i was going through the files

PR #611, opened 2024-09-05
Add --rdma to the test suite

### LINKED ISSUE #598: Add `rdma_mpi` to test suite

aricer123:
## Description

Fixes #598 
Parses arguments and turns on rdma mode if set for frontier and phoenix


sbryngelson:
this doesn't seem to actually close https://github.com/MFlowCode/MFC/issues/598 , it just adds a feature that is necessary for it

sbryngelson:
closing due to inactivity

PR #619, opened 2024-09-09
Add Cleanness CI #2 

### LINKED ISSUE #467: CI Check for code cleanness

okBrian:
## Description

Add Cleanness CI 2nd PR
Fails if pr # warning > master # warnings

Fixes #467

### Type of change
### Scope
## How Has This Been Tested?

[x] Make a commit that includes a new unused variable and check if the CI fails
[x] Make sure once the commit is reverted that it shows -1 and that doesn't fail the CI

**Test Configuration**:

Github Actions on Ubuntu and GNU compilers

## Checklist


sbryngelson:
I just realized that if we have
```
Unused Variable Count: 379, Difference: 0
```
then if `Difference: 5` it's going to be rather impossible for the user to find what ones they contributed! Can you use `diff` or some such to print the additional ones (or removed ones).

sbryngelson:
Phoenix CI won't run because the computer is down... but that doesn't matter for this PR.

okBrian:
So.... I have tried to diff these for the past couple hours. I have been trying a couple different variations on my own branch and none of them have worked for how I think we want them to ie
If Master and PR are different and PR has unique lines then it should output the unique lines of PR.

sbryngelson:
This seems rather trivial, though admittedly I haven't tried it myself. Can you give an example of what you've tried? It seems like `diff file1.txt file2.txt` should be sufficient. You could also ignore the condition that you should only show the diff of one of the files (e.g., PR).

okBrian:
I've been using some variations of grep, diff, and even some with awk
First I would write each grep to a file like this
``
grep -F 'Wunused-variable' master.txt -B 4 >> mUnused.
grep -F 'Wunused-variable' pr.txt -B 4 >> prUnused.
```
Then some examples of commands I ran include these with some variation
```
grep -vxFf mUnused.txt prUnused.txt -B 4
awk 'NR==FNR{a[$0]=1;next}!a[$0]' mUnused.txt prUnused.txt
diff mUnused.txt prUnused
```

The main issue is that many of warnings will have paths like [stuff]/pr/[more stuff] and [stuff]/master/[more stuff] and the diff will pick that up as a difference which is obviously not right.

sbryngelson:
You can use `sed` to remove any instances of `/pr/` or `/master/` from the files altogether. that should fix it.

sbryngelson:
@okBrian what is the status of this?

okBrian:
Assuming I am reading the diff's correctly the warnings showing up are the exact same but they are on different line numbers??? Can you take a look at the most recent commit as I am just showing the diff for both and I am just confused why both are showing up even when they look the exact same.

sbryngelson:
Probably because you are not synced with master.

okBrian:
It works!!!

sbryngelson (approved):

sbryngelson:
Is this ready to merge?

okBrian:
Yes this is ready to merge.

PR #626, opened 2024-09-13
Add modular precision

### LINKED ISSUE #42: Switching to modern and modular precision declaration

aricer123:
## Description

Draft for adding modern and modular precision, replaces instances of kind(0d0) and MPI_DOUBLE_PRECISION with a constant declared in the common/ directory

Fixes #42


sbryngelson:
What about all the `_d0` or `_0d0` statements?

aricer123:
> What about all the `_d0` or `_0d0` statements?

working on those next
Why is it failing on the GPUs? Is this happening with other PRs too?

sbryngelson:
I'm not sure, to be honest. I will check back tomorrow.

sbryngelson:
Did you read the log in the failing CI run?

https://github.com/MFlowCode/MFC/actions/runs/10854071036/job/30132336050?pr=626#:~:text=364-,%5B%2050%25%5D%20Building%20Fortran%20object%20CMakeFiles/syscheck_lib.dir/fypp/syscheck/syscheck.fpp,actions%2Drunner%2D3/_work/MFC/MFC/src/syscheck/syscheck.fpp%3A%2056),-367

sbryngelson:
@wilfonba per your request

src/common/m_precision_select.f90
```diff
@@ -0,0 +1,17 @@
+!>
+!! @file m_precision_select.f90
+!! @brief Contains module m_precision_select
+
+!> @brief This file contains the definition of floating point used in MFC
+module m_precision_select
+    use mpi
```
henryleberre: @aricer123 if you wrap the `use mpi` like this in an `ifdef`,
```fortran
#ifdef MFC_MPI
    use mpi                    !< Message passing interface (MPI) module
#endif
```
it should pass the `--no-mpi` workflow. MFC supports being built with and without MPI, the library used to scale to multiple cores/nodes.

aricer123: yea i figured because it failed on the test with no mpi

aricer123: there was another issue with syscheck module i fixed too will commit another draft soon

src/common/m_precision_select.f90
```diff
@@ -0,0 +1,17 @@
...
+
+    implicit none
+
+    integer, parameter :: single_precision = selected_real_kind(6, 37)
+    integer, parameter :: double_precision = selected_real_kind(15, 307)
+
+    integer, parameter :: wp = double_precision
+    integer, parameter :: mpi_p = MPI_DOUBLE_PRECISION
```
henryleberre: My bad @aricer123, I forgot to mention something else. The parts where you use symbols shipped with MPI, like `MPI_DOUBLE_PRECISION`, will need to have "dummy" values for the case where MFC is not built with MPI. Since the variables won't be used when MPI is off, the actual value you give these variables should not matter. So maybe something like this:
```fortran
#ifdef MFC_MPI
integer, parameter :: mpi_p = MPI_DOUBLE_PRECISION
#else
integer, parameter :: mpi_p = dflt_int
#endif
```

if you `use m_constants`. Note: You only need to provide alternate values if `mpi_p` is used by compiled code when MFC is built with `--no-mpi`. That is, if `mpi_p` is only ever referenced behind `#ifdef MFC_MPI`s, it doesn't have to exist when MFC is built with `--no-mpi`.

By the way, it's nice to use `dflt_int` (=-100) because it's a recognizable value.

src/common/m_precision_select.f90
henryleberre: Great!

src/common/m_precision_select.f90
sbryngelson: @aricer123, you can find fixes for many of these things on the previous PR that worked on this, https://github.com/MFlowCode/MFC/pull/119

aricer123: I fixed these issues and added modular precision for other instances of d0 which i had missed. Submitted another draft. Let me know if theres anything I missed.

aricer123:
Looks like its only failing formatting.
Any tips to pass @henryleberre? I haven't seen this one before.

sbryngelson:
@aricer123 have you read the code documentation on the website?

aricer123:
> @aricer123 have you read the code documentation on the website?

Yes I have taken a look at it but I still can't figure out why this would fail the formatting test. I don't recall changing anything else in the file which is currently failing but I will take a closer look and let you know.

sbryngelson:
`./mfc.sh format`

aricer123:
done

sbryngelson:
@arciyer123 this PR is ready to merge?

henryleberre:
@sbryngelson Looks like my latest PR is the source of a lot of conflicts.

sbryngelson:
> @sbryngelson Looks like my latest PR is the source of a lot of conflicts.

I know, that's ok. It was a rhetorical question.

aricer123:
Formatting is done still need to resolve the conflicts

aricer123:
nevermind, having issues with the merge since it's making me do it manually
looks like somehow i screwed up my working branch
Do you guys have any tips to do this in a cleaner way?

PR #630, opened 2024-09-17
Small correction to CFL time-stepping

wilfonba:
## Description

Fix a small logic error with CFL-based dt

### Type of change

### Scope


sbryngelson (approved):

sbryngelson:
Will merge on passing CI

wilfonba:
The failed test is a Phoenix issue I think. I tried running something earlier today and got an "orte dameon has disconnected" or something along those lines.

sbryngelson:
> The failed test is a Phoenix issue I think. I tried running something earlier today and got an "orte dameon has disconnected" or something along those lines.

Max will happy chew your ear off about these weird OpenMPI warnings that compilers sometimes spit out (many involving `orte`). You may be talking about an error though, which is more unusual. Some warnings are not meaningful to us. In any case, I re-ran the runner, and it will probably pass. If not, I/we can see if they changed the compilers or something. Eventually, the RHEL9 nodes will become all the nodes, and some modules could change -- could be related to that though unlikely.

PR #639, opened 2024-10-01
Increase granularity of halo-exchange timing info

max-Hawkins:
## Description

Previously, the NVTX ranges measuring the so-called 'MPI' time included the time to unpack and pack the contiguous buffers actually exchanged during the MPI_SENDRECV operation. While this may make sense, to avoid confusion and always be able to get proper communication time, I renamed the 'RHS-MPI' NVTX range to 'RHS-MPI+BufPack' and added an NVTX range only around the MPI_SENDRECV operation called 'RHS-MPI_SENDRECV.'

### Type of change

- [x ] New feature (non-breaking change which adds functionality)

## How Has This Been Tested?
I ran an example case under nsys with and without this change. The reported timing from the new RHS-MPI_SENDRECV NVTX range was within 5% error of the MPI trace time reporting for this example.

See below for screenshots from the NSYS reports. In this example, the MPI_SENDRECV time is ~1.4% of the total 'MPI' time.

This shows the NSYS MPI trace timing info. Note the highlighted line's 'total time'
Screenshot 2024-09-30 at 5 47 37 PM
This is the NVTX range timing information. Note that the RHS-MPI_SENDRECV range total time is similar to the new NVTX range result:
Screenshot 2024-09-30 at 5 50 07 PM
Screenshot 2024-09-30 at 5 50 35 PM

**Test Configuration**:
4 V100 nodes on Phoenix running the 2D shockbubble case for 700 timesteps.

## Checklist

- [ x] I ran `./mfc.sh format` before committing my code
- [ x] This PR does not introduce any repeated code (it follows the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
- [ x] I cannot think of a way to condense this code and reduce any introduced additional line count

- [x ] Checked that the code compiles using NVHPC compilers
- [ x] Ran the code on either V100, A100, or H100 GPUs and ensured the new feature performed as expected (the GPU results match the CPU results)
- [ x] Enclosed the new feature via `nvtx` ranges so that they can be identified in profiles
- [ x] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this PR

### DIFF
--- a/src/simulation/m_mpi_proxy.fpp
+++ b/src/simulation/m_mpi_proxy.fpp
@@ -27,6 +27,8 @@ module m_mpi_proxy
 
     use m_mpi_common
 
+    use m_nvtx
+
     use ieee_arithmetic
     ! ==========================================================================
 
@@ -1077,10 +1079,12 @@ contains
                     !$acc update host(q_cons_buff_send, ib_buff_send)
                 #:endif
 
+                call nvtxStartRange("RHS-MPI-SENDRECV")
                 call MPI_SENDRECV( &
                     p_send, buffer_count, MPI_DOUBLE_PRECISION, dst_proc, send_tag, &
                     p_recv, buffer_count, MPI_DOUBLE_PRECISION, src_proc, recv_tag, &
                     MPI_COMM_WORLD, MPI_STATUS_IGNORE, ierr)
+                call nvtxEndRange
 
                 #:if rdma_mpi
                     !$acc end host_data
--- a/src/simulation/m_rhs.fpp
+++ b/src/simulation/m_rhs.fpp
@@ -807,7 +807,7 @@ contains
             ix, iy, iz)
         call nvtxEndRange
 
-        call nvtxStartRange("RHS-MPI")
+        call nvtxStartRange("RHS-MPI+BufPack")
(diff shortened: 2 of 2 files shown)


sbryngelson:
Is it possible to adjust the NVTX range naming to make the hierarchy of the ranges more obvious? For example, `TimeStep — RHS — Communication — MPI/SendRecv` or something similar, with this style used for all ranges. Right now, it's hard to discern which call nests other calls. I do realize that other parts of Nsys GUI make this more obvious.

max-Hawkins:
Just updated things. Here's the output viewed in NSYS (table and graph views):

Screenshot 2024-10-17 at 11 06 23 AM

sbryngelson:
@max-Hawkins would you mind updating/finish this for merge?

max-Hawkins:
@henryleberre Ready for your evaluation.
Screenshot 2024-11-11 at 7 33 54 PM

sbryngelson (approved):

sbryngelson:
needs `./mfc.sh format -j 4`

PR #651, opened 2024-10-16
Fix Delta Again

wilfonba:
## Description

Fixes a bug in the Delta template file that made it incompatible with case file command line options introduced in #338 when this file was first created. I'm guessing this hadn't been caught since using command line arguments in case files isn't particularly common.

### Type of change

### Scope

## How Has This Been Tested?

Successfully submitted job with case file command line options on Delta.


sbryngelson:
Fair enough though I think you tagged the wrong PR?

wilfonba:
Corrected

henryleberre:
That's right @wilfonba. The `--` actually used to forward arguments to the `mpirun`/`srun`/... command and not the case file.. The template files did away with this terrible idea.

PR #653, opened 2024-10-16
Working 0-1D Chemistry (among other things)

henryleberre:
## Description

- Adds three new working example cases `nD_perfect_reactor`, `1D_reactive_shocktube`, and `1D_inert_shocktube`.
- Fixes HLL+Advection, Pressure/Energy calculations, and more (for chemistry).
- Removes the horrible `misc/viz.py` script and instead adds a reusable library in the toolchain for this. See the small, individual, `viz.py` files for each example case I added.
- I put some code I often used to remove case parameters for dimensions above 1 or 2 in one file. Cases can just import `mfc.case_utils`.
- For all of this to work, I created a `pyproject.toml` for the toolchain so that it can be installed in the venv. You can also `pip3 install -e toolchain/`.
- When `./mfc.sh run` passes its internal state to the cases it reads, through CLI arguments, it now does so using the `--mfc` argument (instead of using positional arguments). This improves the usability of the feature because you no longer have to pass your case a fake first argument when running it using `python3`.
- Now that pyrometheus/pyrometheus#87 and pyrometheus/pyrometheus#91 have been merged, we point directly to pyrometheus's first release on PyPi.
- Refactors the usage of the `i[x,y,z]` variables everywhere that made the code confusing to read.
- While computing the RHS, we no longer perform the unnecessary conversion of the buffer region's conservatives variables to primitives ones. That is already covered by populating the buffer region from the interior points. This greatly improves the performance of the chemistry code since the Newton solve for temperature would fail in this region due to an invalid conservative state.
- Augments #410 by correctly (and automatically) handling test cases that need a non-standard build of MFC (for features like Chemistry and Analytically defined patches). 
- Makes validating case files a lot faster using `fastjsonschema`. Needed for loading many case files from files (e.g. testing using example cases).
- Defers loading/generating cases to after filtering. Improves performance, especially when tests include examples.
- Adds test cases for chemistry:
```
    UUID     Trace
 ─────────────────────────────────────────────────────────────────────────────────────
  5DCF300C   1D -> Chemistry -> Perfect Reactor
  1550B67E   2D -> Chemistry -> Perfect Reactor
  E8372F50   3D -> Chemistry -> Perfect Reactor
  4D7D85B4   1D -> Chemistry -> Inert Shocktube -> Riemann Solver 1 -> Gamma Method 1
  2BDE2018   1D -> Chemistry -> Inert Shocktube -> Riemann Solver 1 -> Gamma Method 2
  6B3AE48F   1D -> Chemistry -> Inert Shocktube -> Riemann Solver 2 -> Gamma Method 1
  F8ADA51B   1D -> Chemistry -> Inert Shocktube -> Riemann Solver 2 -> Gamma Method 2
```
- Fixed the documentation for setting `BOOST_INCLUDE` on macOS.
- HLLC & two ways of calculating $\gamma$.

result
result
result

### Type of change

### Scope

## How Has This Been Tested?

The three example cases recreate simulations from papers referenced in their respective case.py files and READMEs.


src/simulation/m_riemann_solvers.fpp
sbryngelson, line 1: please add to HLLC (we use HLLC for most simulations and it has better dissipation properties). otherwise, this is looking very good!

henryleberre: @IamImbact98 is working on it!

sbryngelson: perfect

sbryngelson (dismissed):
Can you add goldenfiles/tests since we now have examples? Even if they are only for 0D or 1D. this would also help ensure the user has installed boost/pyro and other dependencies correctly.

henryleberre:
> Can you add goldenfiles/tests since we now have examples? Even if they are only for 0D or 1D. this would also help ensure the user has installed boost/pyro and other dependencies correctly.

Yes, will do. There is one caveat, however. `./mfc.sh test` would build two versions of MFC, the regular, and one using chemistry + h2o2.yml (the reaction mechanism). Are you fine with this? The user need not worry about this happening, they will just MFC "twice". Are you fine with this?

sbryngelson:
> > Can you add goldenfiles/tests since we now have examples? Even if they are only for 0D or 1D. this would also help ensure the user has installed boost/pyro and other dependencies correctly.
> 
> Yes, will do. There is one caveat, however. `./mfc.sh test` would build two versions of MFC, the regular, and one using chemistry + h2o2.yml (the reaction mechanism). Are you fine with this? The user need not worry about this happening, they will just see MFC build "twice". Are you fine with this?

Perhaps... what is the reason? For example, as a user it would be convenient to be able to run a chemistry case, and then switch to a 'normal' case without rebuilding (or vice-versa). Is the mechanism only available at compile time for some reason?

henryleberre:
> > > Can you add goldenfiles/tests since we now have examples? Even if they are only for 0D or 1D. this would also help ensure the user has installed boost/pyro and other dependencies correctly.
> > 
> > 
> > Yes, will do. There is one caveat, however. `./mfc.sh test` would build two versions of MFC, the regular, and one using chemistry + h2o2.yml (the reaction mechanism). Are you fine with this? The user need not worry about this happening, they will just see MFC build "twice". Are you fine with this?
> 
> Perhaps... what is the reason? For example, as a user it would be convenient to be able to run a chemistry case, and then switch to a 'normal' case without rebuilding (or vice-versa). Is the mechanism only available at compile time for some reason?

This is transparent to the user. Chemistry is a compile-time feature because it needs to invoke Pyrometheus to generate code for a specific mechanism file. The build "slug" system is in place to prevent rebuilds for the exact reason you described. The slug is (partially) a function of the mechanism. When you switch back to running a "regular" MFC case, it will re-use a previous build.

sbryngelson:
> > > > Can you add goldenfiles/tests since we now have examples? Even if they are only for 0D or 1D. this would also help ensure the user has installed boost/pyro and other dependencies correctly.
> > > 
> > > 
> > > Yes, will do. There is one caveat, however. `./mfc.sh test` would build two versions of MFC, the regular, and one using chemistry + h2o2.yml (the reaction mechanism). Are you fine with this? The user need not worry about this happening, they will just see MFC build "twice". Are you fine with this?
> > 
> > 
> > Perhaps... what is the reason? For example, as a user it would be convenient to be able to run a chemistry case, and then switch to a 'normal' case without rebuilding (or vice-versa). Is the mechanism only available at compile time for some reason?
> 
> This is transparent to the user. Chemistry is a compile-time feature because it needs to invoke Pyrometheus to generate code for a specific mechanism file. The build "slug" system is in place to prevent rebuilds for the exact reason you described. The slug is (partially) a function of the mechanism. When you switch back to running a "regular" MFC case, it will re-use a previous build.
> 
> ```python
>     def get_slug(self, case: input.MFCInputFile) -> str:
>         if self.isDependency:
>             return self.name
> 
>         m = hashlib.sha256()
>         m.update(self.name.encode())
>         m.update(CFG().make_slug().encode())
>         m.update(case.get_fpp(self, False).encode())
> 
>         if case.params.get('chemistry', 'F') == 'T':
>             m.update(case.get_cantera_solution().name.encode())
> 
>         return m.hexdigest()[:10]
> ```

Yes it seems there is no way to avoid this, so long as Pyro is generating our code on the fly for each chemistry case. So yes I am OK with this, as we should probably have a 'store' of mechanism files that we test with located somewhere in VC

henryleberre:
> Yes it seems there is no way to avoid this, so long as Pyro is generating our code on the fly for each chemistry case. So yes I am OK with this, as we should probably have a 'store' of mechanism files that we test with located somewhere in VC

I agree. A bunch of standard mechanisms are shipped with Cantera (see https://github.com/Cantera/cantera/tree/main/data). This is why you can run these example cases without downloading a mechanism file.

sbryngelson:
Benchmarking is failing with
```console

      .=++*:          -+*+=.        | sbryngelson3@atl1-1-01-003-34-0.pace.gatech.edu [Linux]
     :+   -*-        ==   =* .      | -------------------------------------------------------
  :*##-.....:*+   .#%+++=--+=:::.   | --jobs 24
  -=-++-======#=--**+++==+*++=::-:. | --mpi --no-gpu --no-debug --no-gcov --no-unified
 .:++=----------====+*= ==..:%..... | --targets pre_process, simulation, and post_process


 Benchmarking pre_process, simulation, and post_process (build/benchmarks/e180):
  
   1/4: 5eq_rk3_weno3_hllc @ benchmarks/5eq_rk3_weno3_hllc/case.py
    
     > Log:     build/benchmarks/e180/5eq_rk3_weno3_hllc.out
     > Summary: build/benchmarks/e180/5eq_rk3_weno3_hllc.yaml
     $ ./mfc.sh run /storage/scratch1/6/sbryngelson3/runners/actions-runner-4/_work/MFC/MFC/pr/benchmarks/5eq_rk3_weno3_hllc/case.py --case-optimization --targets pre_process simulation post_process --output-summary /storage/scratch1/6/sbryngelson3/runners/actions-runner-4/_work/MFC/MFC/pr/build/benchmarks/e180/5eq_rk3_weno3_hllc.yaml -c phoenix --gpu -g 0 1 -n 2 -- 1

 

Error: Failed to load YAML from "/storage/scratch1/6/sbryngelson3/runners/actions-runner-4/_work/MFC/MFC/pr/build/benchmarks/e180/5eq_rk3_weno3_hllc.yaml": [Errno 2] No such file or directory: '/storage/scratch1/6/sbryngelson3/runners/actions-runner-4/_work/MFC/MFC/pr/build/benchmarks/e180/5eq_rk3_weno3_hllc.yaml'

Terminated

mfc: ERROR > main.py finished with a 143 exit code.
```

henryleberre:
@sbryngelson Thanks for letting me know. Just fixed it.

sbryngelson:
@henryleberre mind pulling from upstream for 'cleanliness'? I made an update there yesterday and want to see how it worked out.

sbryngelson:
@henryleberre pull my 'fix cleanliness' PR so it's up to date with master if possible (that's the thing I would like to test).

henryleberre:
> @henryleberre pull my 'fix cleanliness' PR so it's up to date with master if possible (that's the thing I would like to test).

Yes I did, that was the force-push you saw (a rebase on upstream). This is the workflow step that is causing it to fail:
https://github.com/henryleberre/ChemMFC/blob/23df5d6551e7a82217b3259272efe7075d1cf30a/.github/workflows/cleanliness.yml#L99-L103

sbryngelson (commented):
🚀🙌🏻

sbryngelson (commented):
Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?

sbryngelson:
> Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?

Also @henryleberre: It's somewhat unclear what all the different variables really mean (and they don't seem to have comments). I think even after refactor we have 2 or 3 versions of this variable. Can we put comments to say what those are/why they exist?

henryleberre:
> > Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?
> 
> Also @henryleberre: It's somewhat unclear what all the different variables really mean (and they don't seem to have comments). I think even after refactor we have 2 or 3 versions of this variable. Can we put comments to say what those are/why they exist?

I did "document" these like this:
```
    ! Cell Indices for the interior points (O-m, O-n, 0-p).
    type(int_bounds_info) :: idwint(1:3)
    !$acc declare create(idwint)

    ! Cell Indices for the entire domain, including the buffer region.
    type(int_bounds_info) :: idwbuff(1:3)
    !$acc declare create(idwbuff)
```

I could maybe have specified these were the proc's local domain. They are only defined in one place now so it's very straightforward to understand. The other indices are module-specific and I'm not sure I have enough context to explain them better than the code does.

henryleberre:
> Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?

I'm not sure how this is true. Why are the 2-rank test cases not enough?

sbryngelson:
> > Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?
> 
> I'm not sure how this is true. Why are the 2-rank test cases not enough?

2-rank may be enough, but I'm not sure if our current single test case in the suite is enough.

sbryngelson:
> > > Testing this is a bit harder because it only appears meaningfully in multi-rank cases. Once you're done, can you test for correctness on a longer 4-rank case, for example, to be sure there isn't a subtle mistake?
> > 
> > 
> > Also @henryleberre: It's somewhat unclear what all the different variables really mean (and they don't seem to have comments). I think even after refactor we have 2 or 3 versions of this variable. Can we put comments to say what those are/why they exist?
> 
> I did "document" these like this:
> 
> ```
>     ! Cell Indices for the interior points (O-m, O-n, 0-p).
>     type(int_bounds_info) :: idwint(1:3)
>     !$acc declare create(idwint)
> 
>     ! Cell Indices for the entire domain, including the buffer region.
>     type(int_bounds_info) :: idwbuff(1:3)
>     !$acc declare create(idwbuff)
> ```
> 
> I could maybe have specified these were the proc's local domain. They are only defined in one place now so it's very straightforward to understand. The other indices are module-specific and I'm not sure I have enough context to explain them better than the code does.

Ah I did not see this part. I mostly saw the module indices. I think those are still confusing. Why do we need any such variables beyond `idwint` and `idwbuff`? I can imagine perhaps a couple edge cases but that's about it (like when we do direction reshaping).

src/simulation/m_riemann_solvers.fpp
henryleberre: It's... DONE!

sbryngelson:
The debug cases are failing on this case (perhaps among others, they only run a fraction of the cases in CI):

```
 m_mpi_proxy.fpp:1289: @:ALLOCATE_GLOBAL(ib_buff_send(0:-1 + gp_layers * (m + 2*gp_layers + 1)* (n + 2*gp_layers + 1)* (p + 2*gp_layers + 1)/ (min(m, n, p) + 2*gp_layers + 1)))
 m_mpi_proxy.fpp:1301: @:ALLOCATE_GLOBAL(ib_buff_recv(0:ubound(ib_buff_send, 1)))
 m_ibm.fpp:106: @:ALLOCATE_GLOBAL(ghost_points(num_gps))
 m_ibm.fpp:107: @:ALLOCATE_GLOBAL(inner_points(num_inner_gps))
At line 567 of file /Users/runner/work/MFC/MFC/src/simulation/m_ibm.fpp
Fortran runtime error: Index '1' of dimension 1 of array 'ghost_points' above upper bound of -99999918
Error termination. Backtrace:
Could not print backtrace: executable file is not an executable
Could not print backtrace: executable file is not an executable
Could not print backtrace: executable file is not an executable
Could not print backtrace: executable file is not an executable
#0  0x10345ad2f
#1  0x10345b8d7
#2  0x10345bc8f
#3  0x1022f8bdf
#4  0x10230ad67
#5  0x102555d83
#6  0x1026ef79b
#7  0x1026efdfb
mfc: ERROR > :( /Users/runner/work/MFC/MFC/build/install/7e1418ef14/bin/simulation failed with exit code 2.
```

Case `725A3D56`, `F13DF273`, and `0F6C4340` from the fraction that ran in CI here
https://github.com/MFlowCode/MFC/actions/runs/11660348744/job/32478811683?pr=653#step:7:5812

CMakeLists.txt
```diff
@@ -133,6 +133,8 @@ if (CMAKE_Fortran_COMPILER_ID STREQUAL "GNU")
             -fimplicit-none
             #-ffpe-trap=invalid,zero,denormal,overflow
             -fsignaling-nans
+            -finit-real=snan
+            -finit-integer=-99999999
```
sbryngelson, line 137: good idea, though you uncovered an issue that should be fixed it appears

src/common/m_derived_types.fpp
```diff
@@ -311,9 +311,10 @@ module m_derived_types
     type chemistry_parameters
         character(LEN=name_len) :: cantera_file !< Path to Cantera file
 
-        logical :: advection
         logical :: diffusion
         logical :: reactions
+
+        integer :: gamma_method
```
sbryngelson, line 320: can we add a comment here that indicates what `gamma_method` ultimately means?

src/common/m_mpi_common.fpp
```diff
@@ -254,6 +254,15 @@ contains
                             MPI_COMM_WORLD, ierr)
         end if
 
+#else
+
+        icfl_max_glb = icfl_max_loc
+
+        if (any(Re_size > 0)) then
```
sbryngelson, line 261: do we have an `if(viscous)` logical flag?

src/common/m_variables_conversion.fpp
```diff
@@ -1158,13 +1133,13 @@ contains
                         end do
 
                         call get_mixture_molecular_weight(Ys, mix_mol_weight)
-                        T = q_prim_vf(E_idx)%sf(j, k, l)*mix_mol_weight/(gas_constant*rho)
+                        T = q_prim_vf(tempxb)%sf(j, k, l)
```
sbryngelson: tempxb -> tempx_beg (following other nomenclature in the code)?

src/post_process/m_global_parameters.fpp
```diff
@@ -691,6 +698,19 @@ contains
             buff_size = buff_size + fd_number
         end if
 
+        ! Configuring Coordinate Direction Indexes =========================
+        idwint(1)%beg = 0; idwint(2)%beg = 0; idwint(3)%beg = 0
```
sbryngelson, line 699: comment on what `idw` means?

src/simulation/m_rhs.fpp
```diff
@@ -831,24 +801,16 @@ contains
                                                  qR_prim, &
                                                  q_prim_qp, &
                                                  dq_prim_dx_qp, dq_prim_dy_qp, dq_prim_dz_qp, &
-                                                 ix, iy, iz)
+                                                 idwbuff(1), idwbuff(2), idwbuff(3))
```
sbryngelson, line 804: one day we will collapse these into `indices -> indices%x -> indices%x%beg`... just not yet

src/simulation/m_riemann_solvers.fpp
```diff
@@ -469,11 +483,59 @@ contains
...
 
-                            H_L = (E_L + pres_L)/rho_L
-                            H_R = (E_R + pres_R)/rho_R
+                                T_L = pres_L/rho_L/R_gas_L
+                                T_R = pres_R/rho_R/R_gas_R
+
+                                call get_species_specific_heats_r(T_L, Cp_iL)
+                                call get_species_specific_heats_r(T_R, Cp_iR)
+
+                                if (chem_params%gamma_method == 1) then
```
sbryngelson, line 508: comment above this line on what the difference between these two is

src/common/m_variables_conversion.fpp
```diff
@@ -1434,47 +1409,57 @@ contains
         real(kind(0d0)), intent(in) :: H
         real(kind(0d0)), dimension(num_fluids), intent(in) :: adv
         real(kind(0d0)), intent(in) :: vel_sum
+        real(kind(0d0)), intent(in) :: c_avggg
```
sbryngelson: name seems weird

src/simulation/include/inline_riemann.fpp
```diff
@@ -26,10 +28,39 @@
     gamma_avg = (sqrt(rho_L)*gamma_L + sqrt(rho_R)*gamma_R)/ &
                 (sqrt(rho_L) + sqrt(rho_R))
 
-    rho_avg = sqrt(rho_L*rho_R)
     vel_avg_rms = (sqrt(rho_L)*vel_L(1) + sqrt(rho_R)*vel_R(1))**2d0/ &
                   (sqrt(rho_L) + sqrt(rho_R))**2d0
 
+    #:if chemistry
```
sbryngelson, line 34: is there a reason we use fypp macros here instead if just `if` statements? it does make doing static code analysis harder, since the generated code diverges depending on options.

src/simulation/m_riemann_solvers.fpp
```diff
@@ -2107,26 +2180,77 @@ contains
...
+                                    E_L = gamma_L*pres_L + pi_inf_L + 5d-1*rho_L*vel_L_rms + qv_L
+
+                                    E_R = gamma_R*pres_R + pi_inf_R + 5d-1*rho_R*vel_R_rms + qv_R
+
+                                    H_L = (E_L + pres_L)/rho_L
+                                    H_R = (E_R + pres_R)/rho_R
+                                #:endif
 
                                 @:compute_average_state()
+                                !print *, c_avggg
```
sbryngelson: remove?

sbryngelson (changes requested):
this isn't a complete review but covers at least 90% of the diff

src/simulation/m_rhs.fpp
sbryngelson: perhaps create a tracking GH issue for this as just something that seems like a 'good idea' (if you agree it seems like a good idea). I'm not assigning it to you.

CMakeLists.txt
henryleberre: Figured it out, `num_gps` for IBM was never explicitly initialized to 0.

src/common/m_variables_conversion.fpp
henryleberre: The nomenclature in the code is actually:
- `E_idx` for Energy/pressure. This is because there is always only one such equation. Fixed.
- `<var>xb` and `<var>xe` for all other variables.

So I'm changing the use for temperature to `T_idx` to follow the current nomenclature, based on a sample size of 1.

src/simulation/include/inline_riemann.fpp
henryleberre: We do this because of a _"limitation with Pyrometheus"_. Indeed, we cannot generate a thermochem file for chemistry if there is no chemistry. So a bunch of symbols would be missing if we didn't use `#:if chemistry`. I see no reason why we couldn't generate a thermochem file for a random mechanism if chemistry is off. This could be a GH issue @sbryngelson.

src/post_process/m_global_parameters.fpp
henryleberre: I updated the comments to include what the variable stands for instead  of just its meaning:
```fortran
! Cell Indices for the (local) interior points (O-m, O-n, 0-p).
! Stands for "InDices With BUFFer".
type(int_bounds_info) :: idwint(1:3)

! Cell Indices for the entire (local) domain. In simulation, this includes
! the buffer region. idwbuff and idwint are the same otherwise.
! Stands for "InDices With BUFFer".
type(int_bounds_info) :: idwbuff(1:3)
```

src/common/m_mpi_common.fpp
henryleberre: Not as far as I know.. though we should. I see 82 instances of `Re_size > 0` so even if we did, it isn't used much.

```console
MFC:[chemfc-up] $ grep 'Re_size > 0' -R src/ | wc -l
      82
```

Maybe create a GH issue for this.

sbryngelson: yes please create a GH issue for someone to add this :D

src/simulation/include/inline_riemann.fpp
sbryngelson: I see. It does seem worth creating a `null` reaction mechanism that makes the symbols exist but otherwise does no compute and minimal (or no?) allocations. Can be a GH issue.

src/simulation/m_riemann_solvers.fpp
henryleberre: Done.

src/common/m_variables_conversion.fpp
henryleberre: Fixed.

src/common/m_derived_types.fpp
henryleberre: Done.

src/simulation/include/inline_riemann.fpp
henryleberre: We can actually use any reaction mechanism since it will never be used by MFC

src/common/m_mpi_common.fpp
henryleberre: https://github.com/MFlowCode/MFC/issues/677

src/simulation/m_rhs.fpp
henryleberre: I agree this could be improved when CCE is fixed. It's everywhere so we'll be reminded of it. I haven't thought about what the best solution is.

sbryngelson: can you open a GH issue for it (perhaps tag it as some future issue to worry about when Cray is more stable) then resolve this conversation?

src/simulation/include/inline_riemann.fpp
sbryngelson: I suggest adding that to the PR unless it's a lot of work. Otherwise that's my last comment

src/simulation/m_ibm.fpp
```diff
@@ -106,6 +106,8 @@ contains
         @:ALLOCATE_GLOBAL(ghost_points(num_gps))
         @:ALLOCATE_GLOBAL(inner_points(num_inner_gps))
 
+        print *, "num_gps was ", num_gps
```
sbryngelson, line 109: remove

src/simulation/include/inline_riemann.fpp
sbryngelson: closed in lieu of this issue https://github.com/MFlowCode/MFC/issues/678

src/simulation/m_rhs.fpp
sbryngelson: ref in #679

sbryngelson (approved):

PR #664, opened 2024-10-28
Fixed Examples, Added Examples CI.

### LINKED ISSUE #474: Add `examples/` into CI

okBrian:
## Description

Reduced Example Test cases and added to CI
Fixed examples 
Fixed bugs breaking some of the examples

Currently skips these examples:
["2D_ibm_cfl_dt", "1D_sodHypo", "2D_viscous", "2D_laplace_pressure_jump", "2D_bubbly_steady_shock", "2D_advection", "2D_hardcodied_ic", "2D_ibm_multiphase"]

Fixes #474 

### To-do

- [ ] Add GH issue listing skipped example cases in CI, what it means for them to be currently "broken", and what you remember about why they are broken (NaN, tolerance, etc.). Include some possible paths to success (grid size? time step size? etc.)

### Type of change

### Scope

## How Has This Been Tested?

- [x] Github CI
- [x] Locally on a Ubuntu VM with intel compilers, and my compiler with GCC 14.2

**Test Configuration**:

## Checklist


sbryngelson:
This seems good to me. What options did you use to generate the goldenfiles? If `debug`, can you open a GitHub issue on MFC to force `debug` mode when generating new goldenfiles?

Also requesting review from @henryleberre particularly on the `toolchain/` in case there's something that should be adjusted there.

sbryngelson (approved):
looks good, @henryleberre may have a few comments. i put a comment on the PR for some minor details.

okBrian:
> This seems good to me. What options did you use to generate the goldenfiles? If `debug`, can you open a GitHub issue on MFC to force `debug` mode when generating new goldenfiles?

They are generated with default settings ie (MPI, no-debug, intel: false), so no need for anything special.

toolchain/mfc/test/cases.py
```diff
@@ -711,7 +714,44 @@ def foreach_dimension():
...
+            if path == "scaling":
+                continue
+
+            # # List of currently broken examples -> currently attempting to fix!
+            brokenCases = ["2D_ibm_cfl_dt", "1D_sodHypo", "2D_viscous", "2D_laplace_pressure_jump", "2D_bubbly_steady_shock", "2D_advection", "2D_hardcodied_ic", "2D_ibm_multiphase"]
+            if path in brokenCases:
+                continue
+            name = f"{path.split('_')[0]} -> Example -> {'_'.join(path.split('_')[1:])}"
+            path = os.path.join(common.MFC_EXAMPLE_DIRPATH, path, "case.py")
+            if not os.path.isfile(path):
```
henryleberre, line 815: What case does this avoid? (I might have added this..) An example directory without a case.py feels like it shouldn't exist.

toolchain/mfc/test/cases.py
```diff
@@ -711,7 +714,44 @@ def foreach_dimension():
...
+
+                caseSize = case['m'] * max(case['n'], 1) * max(case['p'], 1)
+                if caseSize > 625:
+                    if case['n'] == 0 and case['p'] == 0:
+                        case['m'] = 625
+                    elif case['p'] == 0:
+                        case['m'] = 25
+                        case['n'] = 25
+                    # m, n, p < 25 causes errors
+                    elif caseSize > 15625:
```
henryleberre: you wouldn't need the comment if you had `elif caseSize > 25*25*25`.

toolchain/mfc/test/test.py
```diff
@@ -92,13 +92,18 @@ def test():
...
+    codes = list(REQUIRED_TARGETS) + [PRE_PROCESS, SIMULATION] + ([POST_PROCESS] if ARG('test_all') else [])
 
+    built_slugs = set()
     for case in cases:
-        if case.rebuild:
-            build(codes, MFCInputFile(os.path.basename(case.get_dirpath()), case.get_dirpath(), case.params))
+        case.delete_output()
+        case.create_directory()
+
+        for code in codes:
```
henryleberre: I wrote another version of this logic in my latest PR which is cleaner but I'll fix the conflicts when I merge it.

toolchain/mfc/test/case.py
```diff
@@ -209,17 +210,22 @@ def create_directory(self):
...
+            return 1e-3
+
         if self.params.get("hypoelasticity", 'F') == 'T':
             return 1e-7
 
-        if any(self.params.get(key, 'F') == 'T' for key in ['relax', 'ib', 'qbmm', 'bubbles']):
-            return 1e-10
-
-        if self.params.get("low_Mach", 'F') == 1 or self.params.get("low_Mach", 'F') == 2:
+        if any(self.params.get(key, 'F') == 'T' for key in ['relax', 'ib', 'qbmm', 'bubbles']) or self.params.get("low_Mach", 'F') == 1 or self.params.get("low_Mach", 'F') == 2:
```
henryleberre: Nitpick but the default value of `low_Mach` probably shouldn't be a boolean. If you just check equality, specifying a default value to `.get` is redundant.

tests/0501B3DA/golden-metadata.txt
```diff
@@ -1,19 +1,19 @@
-This file was created on 2024-09-04 20:38:28.260934.
+This file was created on 2024-10-25 13:13:54.004925.
 
 mfc.sh:
 
-    Invocation: test --generate -o cfl_adap_dt=T
+    Invocation: test -j 8 --generate
```
henryleberre: I see that you regenerated a lot of existing golden files with commands like:
```
./mfc.sh test -j 8 --generate
./mfc.sh test -j 8 --generate -o 3D
```

Are all of these warranted? Given that the only fix in `src` is for analytical patches used in model patches, I don't see how this would be needed.

henryleberre (changes requested):
All my comments are nitpicks except the one about regenerating too many golden files. I'll fix the nitpicks about my own code in this PR in #653 since they are fixed there (mostly).

tests/0501B3DA/golden-metadata.txt
sbryngelson: I'm also confused why we are re-generating goldenfiles.

okBrian: I regenerated the base test suite from when I was trying to fix the tol issues with -O0 since the errors were kind of unpredictable, so I needed to regenerate them to ensure they weren't going off of the -O0 numbers which were giving me issues.
It also probably be better to re-generate them anyways to make sure all examples and base test suite are up to date with the current code in MFC.

sbryngelson: Well we know that they are 'up to date' because they currently pass all CI, right?

henryleberre: Sure, it was useful for debugging, but are they required? As a rule of thumb, it's better not to regenerate golden files when we don't absolutely have to.

toolchain/mfc/test/cases.py
okBrian: I didn't write this portion but I removed it since an example directory without a case.py shouldn't exist.

toolchain/mfc/test/cases.py
okBrian: Done!

toolchain/mfc/test/case.py
okBrian: Done!

tests/0501B3DA/golden-metadata.txt
okBrian: Done!

okBrian:
Considering 175 Base Test cases are breaking on Phoenix CPU after I ran it, I am going to regenerate the base test suite.

sbryngelson:
Why did they fail? If you didn't change how those tests were being run, the tests should pass as they do on the normal master branch, right? Or did you change the tolerances of the tests or something else?

okBrian:
Oh yeaaa, why am I regenerating them. it seems to be an issue with mpirun, so I will look into it.

sbryngelson:
Can be merged once CI passes + merge conflicts resolved.

sbryngelson:
I opened a companion issue to fix the broken CI cases: https://github.com/MFlowCode/MFC/issues/751 @okBrian

Issue #687, opened 2024-11-05
MFC_CUDA_CC flag crashes compilation on Phoenix


max-Hawkins:
Building GPU-enabled MFC on non-compute nodes builds the GPU kernels for many compute capabilities. This causes the build to take very long (>~ 30 minutes). Using `MFC_CUDA_CC` is one solution to specify a subset of compute capabilities to compile kernels for. This works as expected on Delta but causes the builds to crash on Phoenix with the following error:

```
[ 25%] Preprocessing (Fypp) syscheck.fpp
[ 50%] Building Fortran object CMakeFiles/syscheck_lib.dir/fypp/syscheck/syscheck.fpp.f90.o
nvfortran-Error-A CUDA toolkit matching the current driver version (0) or a supported older version (11.8) was not installed with this HPC SDK.
gmake[3]: *** [CMakeFiles/syscheck_lib.dir/build.make:80: CMakeFiles/syscheck_lib.dir/fypp/syscheck/syscheck.fpp.f90.o] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:111: CMakeFiles/syscheck_lib.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:92: CMakeFiles/syscheck.dir/rule] Error 2
gmake: *** [Makefile:170: syscheck] Error 2
```

Full error:
```
[mhawkins60@login-phoenix-rh9-2 MFC]$ ./mfc.sh build --gpu -j 8
mfc: OK > (venv) Entered the Python 3.10.10 virtual environment (>= 3.9).

      .=++*:          -+*+=.        | mhawkins60@login-phoenix-rh9-2.pace.gatech.edu [Linux]
     :+   -*-        ==   =* .      | ------------------------------------------------------
   :*+      ==      ++    .+-       | 
  :*##-.....:*+   .#%+++=--+=:::.   | --jobs 8
  -=-++-======#=--**+++==+*++=::-:. | --mpi --gpu --no-debug --no-gcov --no-unified
 .:++=----------====+*= ==..:%..... | --targets pre_process, simulation, and post_process
  .:-=++++===--==+=-+=   +.  :=     | 
  +#=::::::::=%=. -+:    =+   *:    | ----------------------------------------------------------
 .*=-=*=..    :=+*+:      -...--    | $ ./mfc.sh (build, run, test, clean, count, packer) --help

 Build | syscheck, pre_process, simulation, and post_process | Generic Build

[...]

PR #697, opened 2024-11-06
Specify CUDA version for login node builds

max-Hawkins:
## Description
Currently, specifying the MFC_CUDA_CC flag can cause builds to fail on login nodes where the system can't determine the CUDA version to build for. This adds a compiler flag to specify the CUDA version explicitly. To ensure that rearranging the build process doesn't break, I force CMAKE to find CUDA when the MFC_CUDA_CC flag is set.

Fixes #(687)

Fixes https://github.com/MFlowCode/MFC/issues/687

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

Load modules for specific system
`rm -rf build/staging`
`export MFC_CUDA_CC=XX` where XX is a compute capability
Build GPU MFC on Phoenix and Delta

See above

## Checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -196,10 +196,8 @@ elseif ((CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC") OR (CMAKE_Fortran_COMPILER_
     endif()
 
     if (DEFINED ENV{MFC_CUDA_CC})
-    	# Need to find CUDA to get the version to specify for login node builds
-     	find_package(CUDA)
         string(REGEX MATCHALL "[0-9]+" MFC_CUDA_CC $ENV{MFC_CUDA_CC})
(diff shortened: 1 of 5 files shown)


max-Hawkins:
Mat Colgrove's advice: https://forums.developer.nvidia.com/t/cannot-compile-on-login-node/282759

sbryngelson:
@max-Hawkins let me know the status of this. Will merge when it is ready (it passed Phoenix  tests), just need to confirm it works fine on Delta - perhaps you already checked this.

max-Hawkins:
Tested on both Delta and Phoenix after removing the entire build dir. All we had to do was set the NVHPC_CUDA_HOME environment variable. It just took me longer than it should have to realize this.

sbryngelson (approved):

PR #702, opened 2024-11-07
Fix RDMA MPI

max-Hawkins:
## Description

Previously when trying to run with 'rdma_mpi' equal to true, I would get illegal buffer pointer values in the halo communication routines. After a small change that makes sense to me, CUDA-aware MPI now seems to work on both Delta and Phoenix. 

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

I built and ran 2 cases that differed only by the 'rdma_mpi' status on both Phoenix and Delta. On Delta, it was 3D_TGV and on Phoenix it was 2D_riemman test. h5diff on the resulting hd5f files of the last step showed no difference on Phoenix. Additionally, I took nsys reports on phoenix to show the lack of H2D/D2H copies.

First without RDMA_MPI, the code moves the GPU data to the CPU before MPI exchange:
Screenshot 2024-11-07 at 12 29 07 PM

With RDMA_MPI enabled, there are no data copies. This led to a marginal improvement in the RHS-MPI time for this case:
Screenshot 2024-11-07 at 12 29 04 PM

### DIFF
--- a/src/simulation/m_mpi_proxy.fpp
+++ b/src/simulation/m_mpi_proxy.fpp
@@ -872,8 +872,6 @@ contains
 
         integer :: pack_offsets(1:3), unpack_offsets(1:3)
         integer :: pack_offset, unpack_offset
-        real(kind(0d0)), pointer :: p_send, p_recv
-        integer, pointer, dimension(:) :: p_i_send, p_i_recv
 
 #ifdef MFC_MPI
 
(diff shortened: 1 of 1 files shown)


abbotts:
Unfortunately the bounce pointers are there because CCE 18.0.0 has a bug that will crash the compiler using module arrays directly in the host_data directive. The bounce pointers should be valid too, so if they're failing on a different compiler then we probably have dueling compiler bugs we'll have to handle in a more complex way.

max-Hawkins:
> Unfortunately the bounce pointers are there because CCE 18.0.0 has a bug that will crash the compiler using module arrays directly in the host_data directive. The bounce pointers should be valid too, so if they're failing on a different compiler then we probably have dueling compiler bugs we'll have to handle in a more complex way.

Ah dang. That's unfortunate. I hoped there wasn't some Cray compiler reason for the way things were. Henry isn't in person at the hackathon or I'm sure he would've told me this. I'll just keep this change for my personal work then. Thanks!

PR #703, opened 2024-11-07
Ignore syscheck_lib

max-Hawkins:
## Description

I'm tired of seeing untracked files. This just ignores the 'syscheck_lib' folder.


sbryngelson:
I've never seen this. I'm not sure how this happened in your case. Let me know if others have seen it and we can reopen.

PR #704, opened 2024-11-07
Fix seg fault when fd_order is on

haochey:
## Description

The buffer_size is not initialized for all the variable bound constant (e.g. `idwbuff(2)%beg:idwbuff(2)%end` from [PR# 653](https://github.com/MFlowCode/MFC/commit/e2ceaf555330d70fb6faf407b980a487d4aa9d73#diff-fe8ed9906ee74f3d3f0dbc702d75863fc90286a345cc7067acc6e6cfe80b6d52L762)) when fd_order is on. This will cause seg fault issue on variables like `flux_gsrc_n`. Maybe @henryleberre can confirm this.

### Type of change

### Scope

## How Has This Been Tested?

Turn on `fd_order` and the seg fault issue is resolved.

**Test Configuration**:

* What computers and compilers did you use to test this: MacOS CPU/GCC, Phoenix CPU/GCC, and Delta A100/NVHPC

## Checklist

### DIFF
--- a/src/simulation/m_global_parameters.fpp
+++ b/src/simulation/m_global_parameters.fpp
@@ -1057,6 +1057,11 @@ contains
             buff_size = weno_polyn + 2
         end if
 
+        if (probe_wrt) then
+            fd_number = max(1, fd_order/2)
+            buff_size = buff_size + fd_number
+        end if
+
         ! Configuring Coordinate Direction Indexes =========================
         idwint(1)%beg = 0; idwint(2)%beg = 0; idwint(3)%beg = 0
         idwint(1)%end = m; idwint(2)%end = n; idwint(3)%end = p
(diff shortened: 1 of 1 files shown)


src/simulation/m_global_parameters.fpp
```diff
@@ -1068,6 +1068,11 @@ contains
             buff_size = weno_polyn + 2
         end if
 
+        if (probe_wrt) then
+            fd_number = max(1, fd_order/2)
+            buff_size = buff_size + fd_number
```
sbryngelson: Something here doesn't seem right. You are changing the buffer size (halo region) for the whole simulation based on whether or not you are using a probe. These two should be independent of one another.

haochey: That's right, only the fd_number needs to be updated for using a probe. MFC 4.9.6 has the same issue (buff_size depends on whether or not using a probe) because the halo region was alternated several times in the `s_compute_rhs()`. Fixed.

sbryngelson (approved):

sbryngelson (approved):

sbryngelson:
I think I understand better now after looking at this more properly.

PR #705, opened 2024-11-07
Fix 2D_shockdroplet example case

wilfonba:
## Description

This case file failed as the resolution was increased because the rectangular patches did not cover the entire stretched domain in the y direction.

Fixes #(issue) [optional]

### Type of change

### Scope


sbryngelson (approved):

VaLiuqqq:
if it is 14*D, then the y_centroid is 7*D not 6*D.

sbryngelson:
> if it is 14_D, then the y_centroid is 7_D not 6*D.

@wilfonba

wilfonba:
> > if it is 14_D, then the y_centroid is 7_D not 6*D.
> 
> @wilfonba

The actual value doesn't matter as long as the patch covers the entire grid in the y direction.

PR #706, opened 2024-11-08
Add Broadband Acoustic Src

haochey:
## Description
Broadband source creates acoustic wave with similar amplitude across a wide range of frequency. It will be useful for simulations for noise reduction.

### Type of change

### Scope

## How Has This Been Tested?

Using a probe to record the pressure time history of the source.

**Test Configuration**: 
import json, math

# Configuring case dictionary
print(json.dumps({
    # Logistics ================================================================
    'run_time_info'                : 'T',
    # ==========================================================================

    # Computational Domain Parameters ==========================================
    'x_domain%beg'                 : 0,
    'x_domain%end'                 : 0.3,
    'y_domain%beg'                 : 0,
    'y_domain%end'                 : 0.1,
    'm'                            : 299,
    'n'                            : 99,
    'p'                            : 0,
    'dt'                           : 5e-7,
    't_step_start'                 : 0,
    't_step_stop'                  : 50000,
    't_step_save'                  : 500,
    # ==========================================================================

    # Simulation Algorithm Parameters ==========================================
    'num_patches'                  : 1,
    'model_eqns'                   : 2,
    'alt_soundspeed'               : 'F',
    'num_fluids'                   : 1,
    'mpp_lim'                      : 'F',
    'mixture_err'                  : 'F',
    'time_stepper'                 : 3,
    'weno_order'                   : 5,
    'weno_eps'                     : 1.E-16,
    'teno'                         : 'T',
    'teno_CT'                      : 1E-8,
    'null_weights'                 : 'F',
    'mp_weno'                      : 'F',
    'riemann_solver'               : 2,
    'wave_speeds'                  : 1,
    'avg_state'                    : 2,
    'bc_x%beg'                     : -6,
    'bc_x%end'                     : -6,
    'bc_y%beg'                     : -6,
    'bc_y%end'                     : -6,
    # ==========================================================================

    # Formatted Database Files Structure Parameters ============================
    'format'                       : 1,
    'precision'                    : 2,
    'prim_vars_wrt'                :'T',
    'parallel_io'                  :'T',
    'probe_wrt'                    :'T',
    'fd_order'                     : 2,
    'num_probes'                   : 1,
    'probe(1)%x'                   : 0.13,
    'probe(1)%y'                   : 0.05,

    # ==========================================================================

    # Patch 1 Liquid ===========================================================
    'patch_icpp(1)%geometry'       : 3,
    'patch_icpp(1)%x_centroid'     : 0.15,
    'patch_icpp(1)%y_centroid'     : 0.05,
    'patch_icpp(1)%length_x'       : 0.3,
    'patch_icpp(1)%length_y'       : 0.1,
    'patch_icpp(1)%vel(1)'         : 0.0,
    'patch_icpp(1)%vel(2)'         : 0.0,
    'patch_icpp(1)%pres'           : 1E+05,
    'patch_icpp(1)%alpha_rho(1)'   : 1.19,
    'patch_icpp(1)%alpha(1)'       : 1.0,
    # ==========================================================================

    # Acoustic source ==========================================================
    'acoustic_source'              : 'T',
    'num_source'                   : 1,
    'acoustic(1)%support'          : 2,
    'acoustic(1)%loc(1)'           : 0.1,
    'acoustic(1)%loc(2)'           : 0.05,
    'acoustic(1)%dir'              : math.pi * 0,
    'acoustic(1)%length'           : 0.1,
    'acoustic(1)%pulse'            : 4,
    'acoustic(1)%npulse'           : 1000000,
    'acoustic(1)%mag'              : 10.,
    # 'acoustic(1)%gauss_sigma_time' : 2E-8,
    # 'acoustic(1)%delay'            : 1E-7,
    # ==========================================================================

    # Fluids Physical Parameters ===============================================
    'fluid_pp(1)%gamma'            : 1.E+00/(4.4E+00-1.E+00),
    'fluid_pp(1)%pi_inf'           : 0,
    # ==========================================================================
}))

Screen Shot 2024-11-07 at 11 40 27 PM

* What computers and compilers did you use to test this: MacOS/GCC, Phoenix-CPU/GCC, Phoenix-GPU/NVHPC

## Checklist


src/simulation/m_acoustic_src.fpp
```diff
@@ -173,6 +173,8 @@ contains
         real(kind(0d0)) :: frequency_local, gauss_sigma_time_local
         real(kind(0d0)) :: mass_src_diff, mom_src_diff
         real(kind(0d0)) :: source_temporal
+        real(kind(0d0)), dimension(1:100) :: f_BB, period_BB, sl_BB, bwid_BB, ffre_BB, phi_rn
```
sbryngelson: where does 100 come from? this is a magic number

src/simulation/m_mpi_proxy.fpp
```diff
@@ -2333,6 +2333,13 @@ contains
 
     end subroutine s_mpi_sendrecv_capilary_variables_buffers
 
+    subroutine s_mpi_send_random_number(phi_rn)
```
sbryngelson: what is the purpose of this?

src/simulation/m_acoustic_src.fpp
sbryngelson: i don't know what any of these variables do

sbryngelson (changes requested):
please address comments and add full documentation :D

src/simulation/m_mpi_proxy.fpp
haochey: This ensures the random phase shift is the same for all the ranks as the acoustic source support can span multiple ranks.

src/simulation/m_mpi_proxy.fpp
```diff
@@ -2333,6 +2333,13 @@ contains
...
+        real(kind(0d0)), dimension(1:100) :: phi_rn
```
sbryngelson: this dimension `1:100` is concerning and it should probably be `dimension(num_broadband_freq)` where `num_broadband_freq` is defined in `m_constants`.

src/simulation/m_mpi_proxy.fpp
sbryngelson: Does this need a separate subroutine, or can you just add this to an existing MPI broadcast subroutine (wrapped in an `if(broadband_source)` statement)?

src/simulation/m_acoustic_src.fpp
```diff
@@ -208,6 +210,20 @@ contains
 
             num_points = source_spatials_num_points(ai) ! Use scalar to force firstprivate to prevent GPU bug
 
+            call random_number(phi_rn(1:100))
+            call s_mpi_send_random_number(phi_rn)
```
sbryngelson: i think i see why you have an mpi routine for this now. you create a new random number every time step, but need all ranks to have the same random number. is that right? if so, then yes it will need its own subroutine like you've already created.

src/simulation/m_acoustic_src.fpp
```diff
@@ -208,6 +210,20 @@ contains
...
+            sum_BB = 0d0
+
+            !$acc loop
+            do k = 1, 100
+                f_BB(k) = 500d0 + k*100d0 ! Discrete frequency specturm center
```
sbryngelson: many many magic numbers here 
https://en.wikipedia.org/wiki/Magic_number_(programming)

haochey: Thanks for pointing me to this useful resources! I will properly treat all the magical constant here with their physical meanings.

src/simulation/m_acoustic_src.fpp
haochey: Exactly, I will need to broadcast it every time step.

src/simulation/m_acoustic_src.fpp
```diff
@@ -171,6 +173,18 @@ contains
 
             num_points = source_spatials_num_points(ai) ! Use scalar to force firstprivate to prevent GPU bug
 
+            call random_number(phi_rn(1:100))
```
sbryngelson: there is still a magic number here

src/simulation/m_acoustic_src.fpp
```diff
@@ -171,6 +173,18 @@ contains
...
+            call s_mpi_send_random_number(phi_rn)
+            sum_BB = 0d0
+
+            !$acc loop
```
sbryngelson: Confused about this. should it be parallel loop? sequential loop? this is ambiguous

src/simulation/m_acoustic_src.fpp
```diff
@@ -171,6 +173,18 @@ contains
...
+            do k = 1, num_broadband_freq
+                period_BB(k) = 1d0/(broadband_freq_lowest + k*broadband_bandwidth) ! Acoustic period of the wave at each discrete frequency
+                sl_BB(k) = broadband_spectral_level_constant*mag(ai) + k*mag(ai)/broadband_spectral_level_growth_rate ! Spectral level at each frequency
```
sbryngelson: put comments on new lines

src/simulation/m_acoustic_src.fpp
```diff
@@ -171,6 +173,18 @@ contains
...
+                ffre_BB(k) = dsqrt((2d0*sl_BB(k)*broadband_bandwidth))*cos((sim_time)*2.d0*pi/period_BB(k) + 2d0*pi*phi_rn(k)) ! Source term corresponding to each frequencies
+                sum_BB = sum_BB + ffre_BB(k) ! Total source term for the broadband wave
```
sbryngelson: comments on new lines

sbryngelson:
Needs test suite cases in 1D and 2D at least

src/simulation/m_acoustic_src.fpp
```diff
@@ -136,6 +136,8 @@ contains
         real(kind(0d0)) :: frequency_local, gauss_sigma_time_local
         real(kind(0d0)) :: mass_src_diff, mom_src_diff
         real(kind(0d0)) :: source_temporal
+        real(kind(0d0)), dimension(1:num_broadband_freq) :: period_BB, sl_BB, ffre_BB, phi_rn ! broadband source variables
```
sbryngelson: this is not doxygen format (see line 140/142 for doxygen docstrings)

sbryngelson: add more details as well!

src/simulation/m_mpi_proxy.fpp
sbryngelson: shoudln't this be `intent(inout)` and have a named dimension size (non magic number)

src/simulation/m_acoustic_src.fpp
```diff
@@ -171,6 +176,23 @@ contains
 
             num_points = source_spatials_num_points(ai) ! Use scalar to force firstprivate to prevent GPU bug
 
+            call random_number(phi_rn(1:num_broadband_freq))
+            ! Ensure all the ranks have the same random phase shift
+            call s_mpi_send_random_number(phi_rn)
+            sum_BB = 0d0
+
+            !$acc loop seq
```
sbryngelson: the loop below this one is parallel, I think this one should also be parallel?

haochey: I thought we may have to use sequential loop because the variable `sum_BB` depends on the last iteration. 

Would this be better `!$acc loop reduction(+:sum_BB)` for this loop just like line 215 here `!$acc loop reduction(+:myRho,B_tait,small_gamma)`?

src/simulation/m_mpi_proxy.fpp
haochey: Thanks...I am easily caught by magic numbers :/.

src/simulation/m_acoustic_src.fpp
sbryngelson: that loop on line 215 should really be a sequential one. yes you can use loop reductions indeed

src/common/m_constants.fpp
```diff
@@ -26,4 +26,11 @@ module m_constants
     real(kind(0d0)), parameter :: acoustic_spatial_support_width = 2.5d0 !< Spatial support width of acoustic source, used in s_source_spatial
     real(kind(0d0)), parameter :: dflt_vcfl_dt = 100d0 !< value of vcfl_dt when viscosity is off for computing adaptive timestep size
 
+    !< Broadband acoustic source constant (reference: Tam et al. JSV 2005)
+    integer, parameter :: num_broadband_freq = 100 !< The number of sine wave frequencies in broadband acoustic source.
+    real(kind(0d0)), parameter :: broadband_freq_lowest = 500d0 !< The lower sine wave frequency bound in broadband acoustic source.
+    real(kind(0d0)), parameter :: broadband_bandwidth = 100d0 !< The bandwidth of the discretized sine wave frequencies.
```
sbryngelson: it seems like the user should be able to set of a couple of these parameters themself, like the bandwidth itself, right? or is that something that you just set and would never really want to change?

sbryngelson (changes requested):
just a couple more questions/changes

src/common/m_constants.fpp
haochey: It's something user can set as for their own frequency range of interest. I can make all the constant here as input parameters.

sbryngelson: Does it seem like the frequency range would be something the user would want to set? It seems this way to me. If so, please make it a user input instead of a hard-coded constant. Thanks!

haochey: I've added the requested input parameters and refactored the code.

sbryngelson (approved):

PR #730, opened 2024-11-13
IBM+STL

haochey:
## Description
This PR enables complex geometry (in STL format) in immersed boundary method. Also, analytical levelset for cuboid is added.

For 2D STL files, boundary vertices are grouped to determine the levelset and levelset normals. For 3D STL files, all the vertices are boundary vertices so the procedure to group boundary vertices are skipped.

Here is the procedure implemented:
1. Check and group boundary vertices.
    - 2D case: use an edge manifoldness algorithm to recursively check if an edge is shared by more than one triangle facets. 
                       If so the end vertices of the edge are considered inner vertices.
    - 3D case: skipped
2. Check if interpolation of the STL vertices is required.
    - 2D case: interpolation is required if the length of any boundary edge is longer than the cell width.
    - 3D case: interpolation is required if the length of any boundary edge is longer than the cell width or the area of any 
                       triangle facet is larger than the face area of the cell.
3. Interpolate the STL vertices
    - 2D case: add constant multiples of vertices along the boundary edges with even distribution.
    - 3D case: add constant multiples of vertices along the boundary edges with even distribution and triangle facets with 
                       random distribution based on barycentric coordinates.
4. If the STL file is interpolated, use the updated boundary vertex group and perform the following; otherwise, directly 
    perform the following.
    - 2D case: use shortest pair algorithm to find the distance (levelset) between each cell center and the boundary of the 
                       model and the normals (levelset_norm) associated with the boundary edge.
    - 3D case: use shortest pair algorithm to find the distance (levelset) between each cell center and the boundary of the 
                       model and the normals (levelset_norm) associated with the triangle facets. 
5. Correct the signs of the levelset and levelset_norm.
    - 2D case: negative if it's inside the immersed boundary, flip the normals if it's outside the immersed boundary.
    - 3D case: same as the 2D case.

Test suite adds two tests: 2D->IBM->STL and 3D->IBM->STL 

### Type of change

### Scope

## How Has This Been Tested?
Compared the 2D IBM (rectangles and circles) and 3D (cubes and spheres) between analytical method and STL+IBM method.

**Test Configuration**:

2D Comparison (Rectangles)
https://github.com/user-attachments/assets/6948d404-96ff-4219-8464-a768d1b3b530

3D Comparison (Cuboid)
https://github.com/user-attachments/assets/e852a124-3ce7-4e84-8e2c-a6c9f55fdae9

2D New Geo
https://github.com/user-attachments/assets/1e405a1a-9aff-40d2-8d99-dc2ea3a12df8

2D Subsonic flow over MFC Character
https://github.com/user-attachments/assets/f56b8df9-1035-43bf-ba87-b7fb03338aca

3D New Geo
https://github.com/user-attachments/assets/635ad04c-dcd1-470b-8311-d114f0593a66

Mach  1.5 over a Mexican pyramid
https://github.com/user-attachments/assets/88d7731d-e660-47aa-9c29-b866f34ea21d
X-Y section view
https://github.com/user-attachments/assets/719557a6-d45b-4638-9957-1f3bba1f1aa7

* What computers and compilers did you use to test this: MacOS/gcc, Phoenix-cpu/gcc, Phoenix-gpu/NVHPC

## Checklist


src/pre_process/m_compute_levelset.fpp
```diff
@@ -315,23 +316,25 @@ contains
                         if (side_dists(2) == 0) then
                             levelset_norm%sf(i, j, 0, ib_patch_id, 1) = 0d0
                         else
-                            levelset_norm%sf(i, j, 0, ib_patch_id, 1) = side_dists(2)/ &
+                            levelset_norm%sf(i, j, 0, ib_patch_id, 1) = -side_dists(2)/ &
```
wilfonba, line 324: This subroutine has several minor changes, and I can't tell what their purpose is. I see that the simulation results for a 2D rectangle are correct, so I'm guessing these are a result of adding the cuboid patch?

src/pre_process/m_model.fpp
```diff
@@ -549,4 +554,697 @@ contains
...
+                boundary_v(store_index, 1, :) = temp_boundary_v(i, 1, :)
+                boundary_v(store_index, 2, :) = temp_boundary_v(i, 2, :)
+            end if
+        end do
+
+        !  do i = 1, boundary_edge_count
+        !     print*, 'edge:',i, 'start', boundary_v(i, 1, 1:2)
+        !     print*, 'edge:',i, 'end', boundary_v(i, 2, 1:2)
+        !     print*, '=============================================='
+        ! end do
```
wilfonba: Remove?

src/pre_process/m_model.fpp
```diff
@@ -549,4 +554,697 @@ contains
...
+            ! Add vertices inside the triangle
+            x1 = model%trs(i)%v(1, 1)
+            y1 = model%trs(i)%v(1, 2)
+            z1 = model%trs(i)%v(1, 3)
+            x2 = model%trs(i)%v(2, 1)
+            y2 = model%trs(i)%v(2, 2)
+            z2 = model%trs(i)%v(2, 3)
+            x3 = model%trs(i)%v(3, 1)
+            y3 = model%trs(i)%v(3, 2)
+            z3 = model%trs(i)%v(3, 3)
```
wilfonba: There's an existing derived type, `t_triangle`, that would these assignments conform to.

src/pre_process/m_model.fpp
```diff
@@ -549,4 +554,697 @@ contains
...
+        do i = 1, num_triangles
+            ! Loop through the 3 edges of each triangle
+            do j = 1, 3
+                ! Get the coordinates of the two vertices of the current edge
+                x1 = model%trs(i)%v(j, 1)
+                y1 = model%trs(i)%v(j, 2)
+                z1 = model%trs(i)%v(j, 3)
+                x2 = model%trs(i)%v(mod(j, 3) + 1, 1)  ! Next vertex in the triangle (cyclic)
+                y2 = model%trs(i)%v(mod(j, 3) + 1, 2)
+                z2 = model%trs(i)%v(mod(j, 3) + 1, 3)
```
wilfonba: Same comment as before about the `t_triangle` type.

src/pre_process/m_model.fpp
```diff
@@ -549,4 +554,697 @@ contains
...
+            x1 = model%trs(i)%v(1, 1)
+            y1 = model%trs(i)%v(1, 2)
+            z1 = model%trs(i)%v(1, 3)
+            x2 = model%trs(i)%v(2, 1)
+            y2 = model%trs(i)%v(2, 2)
+            z2 = model%trs(i)%v(2, 3)
+            x3 = model%trs(i)%v(3, 1)
+            y3 = model%trs(i)%v(3, 2)
+            z3 = model%trs(i)%v(3, 3)
+            tri_area = f_tri_area(x1, y1, z1, x2, y2, z2, x3, y3, z3)
```
wilfonba: Same comment about `t_triangle` as above. Maybe the area could be added to that derived type too?

wilfonba (commented):
I think there should be STL IBM cases added to the test suite, although I'm not sure how this would be done given the current test suite framework. As we've seen in your updates, even minor changes to any step in the process of generating the levelset can lead to significant differences in results. I'd also be interested in seeing a comparison for "exact" (I'm not sure it's actually exact) vs. STL level set with an airfoil, at least in 2D. Maybe something wild like a NACA 9510 to add some weird curvature. I'm not sure I'd expect the wakes to be the same, but some sort of integral quantity like lift or drag could be compared. I'll let @sbryngelson give his opinion on if this is worthwhile though.

sbryngelson:
> I think there should be STL IBM cases added to the test suite, although I'm not sure how this would be done given the current test suite framework. As we've seen in your updates, even minor changes to any step in the process of generating the levelset can lead to significant differences in results. I'd also be interested in seeing a comparison for "exact" (I'm not sure it's actually exact) vs. STL level set with an airfoil, at least in 2D. Maybe something wild like a NACA 9510 to add some weird curvature. I'm not sure I'd expect the wakes to be the same, but some sort of integral quantity like lift or drag could be compared. I'll let @sbryngelson give his opinion on if this is worthwhile though.

I agree, there needs to be something in the test suite for this.

src/pre_process/m_model.fpp
sbryngelson: many scalar quantities could just be a single array...

src/common/m_constants.fpp
```diff
@@ -28,4 +28,13 @@ module m_constants
     real(kind(0d0)), parameter :: broadband_spectral_level_constant = 20d0 !< The constant to scale the spectral level at the lower frequency bound
     real(kind(0d0)), parameter :: broadband_spectral_level_growth_rate = 10d0 !< The spectral level constant to correct the magnitude at each frqeuency to ensure the source is overall broadband
 
+    ! IBM+STL interpolation constants
+    integer, parameter :: Ifactor_2D = 100 !< Number of interpolated vertices per edge for 2D models
+    integer, parameter :: Ifactor_3D = 10 !< Number of interpolated vertices per edge for 3D models
+    integer, parameter :: Ifactor_bary_3D = 30 !< Number of interpolated vertices per barycentric facet for 3D models
+    real(kind(0d0)), parameter :: threshold_vector_zero = 1d-10 !< Threshold to treat the component of a vector to be zero
```
sbryngelson, line 37: are any of these parameters specific to the grid size? i suspect yes

examples/3D_IBM_STL_ellipsoid/case.py
```diff
@@ -0,0 +1,108 @@
...
+    'patch_ib(1)%model%translate(3)'           : 0 - 0.01,
+    'patch_ib(1)%model%spc'                    : 100,
+    'patch_ib(1)%model%threshold'              : 0.01,
+    'patch_ib(1)%slip'                         : 'F',
+    # ==========================================================================
+
+    # Fluids Physical Parameters ===============================================
+    'fluid_pp(1)%gamma'            : 1.E+00/(gam_a-1.E+00),
+    'fluid_pp(1)%pi_inf'           : 0,
+    'fluid_pp(1)%Re(1)'            : 7535533.2,
```
sbryngelson: this example won't work on master (no viscous = 't')

src/pre_process/m_compute_levelset.fpp
```diff
@@ -344,6 +347,120 @@ contains
 
     end subroutine s_compute_rectangle_levelset
 
+    subroutine s_compute_cuboid_levelset(levelset, levelset_norm, ib_patch_id)
```
sbryngelson: can we change all the `s_compute_XXXXXX` routines here to `s_XXXXXX`. I suppose the `compute` part is implied.

src/pre_process/m_compute_levelset.fpp
```diff
@@ -344,6 +347,120 @@ contains
...
+
+        type(levelset_field), intent(INOUT) :: levelset
+        type(levelset_norm_field), intent(INOUT) :: levelset_norm
+
+        integer :: ib_patch_id
```
sbryngelson: give this an `intent`

src/pre_process/m_compute_levelset.fpp
```diff
@@ -344,6 +347,120 @@ contains
...
+                                levelset_norm%sf(i, j, k, ib_patch_id, 3) = side_dists(5)/ &
+                                                                            abs(side_dists(5))
+                            end if
+
+                        else if (min_dist == abs(side_dists(6))) then
+                            levelset%sf(i, j, k, ib_patch_id) = side_dists(6)
+                            if (side_dists(6) == 0) then
+                                levelset_norm%sf(i, j, k, ib_patch_id, 3) = 0d0
+                            else
+                                levelset_norm%sf(i, j, k, ib_patch_id, 3) = -side_dists(6)/ &
```
sbryngelson, line 455: all of these levelsets need their own test in the test suite.

src/pre_process/m_global_parameters.fpp
```diff
@@ -445,6 +445,14 @@ contains
             patch_ib(i)%m = dflt_real
             patch_ib(i)%p = dflt_real
             patch_ib(i)%slip = .false.
+
+            ! Proper default values for translating STL models
+            patch_ib(i)%model%scale(:) = 1d0
```
sbryngelson: i don't think these defaults make sense. what happened to `dflt_real` and `dflt_int`? i guess we also now need `dflt_str`

src/pre_process/m_patches.fpp
```diff
@@ -1969,12 +1969,26 @@ contains
     !! @param patch_id is the patch identifier
     !! @param patch_id_fp Array to track patch ids
     !! @param q_prim_vf Primitive variables
-    subroutine s_model(patch_id, patch_id_fp, q_prim_vf)
+    !! @param ib True if this patch is an immersed boundary
+    !! @param STL_levelset STL levelset
+    !! @param STL_levelset_norm STL levelset normals
+    subroutine s_model(patch_id, patch_id_fp, q_prim_vf, ib, STL_levelset, STL_levelset_norm)
```
sbryngelson, line 1946: this is not a very descriptive subroutine name

src/pre_process/m_patches.fpp
```diff
@@ -2044,26 +2110,113 @@ contains
...
+                        ! print*, i, j, k, 'levelset', STL_levelset%sf(i, j, k, patch_id), patch_id_fp(i, j, k)
+                        ! print*, i, j, k, STL_levelset_norm%vf(i, j, k, patch_id, 1:3)
+                    end if
 
-                    ! Note: Should probably use *eta* to compute primitive variables
-                    ! if defining them analytically.
-                    @:analytical()
+                    if (j == 41) then
+                        ! print*, i, j, 0, STL_levelset%sf(i, j, 0, patch_id), patch_id_fp(i, j, k)
+                        ! print*, i, j, normals
```
sbryngelson: remove

src/pre_process/m_patches.fpp
```diff
@@ -2044,26 +2110,113 @@ contains
...
+                    if (ib) then
+                        eta = f_model_is_inside(model, point, (/dx, dy, dz/), patch_ib(patch_id)%model%spc)
+                    else
+                        eta = f_model_is_inside(model, point, (/dx, dy, dz/), patch_icpp(patch_id)%model%spc)
+                    end if
 
-                    if (patch_icpp(patch_id)%smoothen) then
-                        if (eta > patch_icpp(patch_id)%model%threshold) then
-                            eta = 1d0
+                    if (.not. ib) then
```
sbryngelson: why don't you switch this so `if(ib)` is first, then else?

src/pre_process/m_initial_condition.fpp
```diff
@@ -181,7 +181,7 @@ contains
 
                     ! 3D STL patch
                 elseif (patch_icpp(i)%geometry == 21) then
-                    call s_model(i, patch_id_fp, q_prim_vf)
+                    call s_model(i, patch_id_fp, q_prim_vf, .false.)
```
sbryngelson: i don't think we want to be calling functions with a 'magic' value like `.false.`. instead, you should define or use an existing boolean for this argument.

src/pre_process/m_initial_condition.fpp
```diff
@@ -206,6 +206,10 @@ contains
                 elseif (patch_ib(i)%geometry == 11) then
                     call s_3D_airfoil(i, ib_markers%sf, q_prim_vf, .true.)
                     call s_compute_3D_airfoil_levelset(levelset, levelset_norm, i)
+
+                    ! STL+IBM patch
+                elseif (patch_ib(i)%geometry == 12) then
+                    call s_model(i, ib_markers%sf, q_prim_vf, .true., levelset, levelset_norm)
```
sbryngelson: see above comment about putting a raw boolean in the middle of the argument list.

src/pre_process/m_initial_condition.fpp
```diff
@@ -206,6 +206,10 @@ contains
                 elseif (patch_ib(i)%geometry == 11) then
                     call s_3D_airfoil(i, ib_markers%sf, q_prim_vf, .true.)
```
sbryngelson: see above comment about putting a raw boolean in the middle of the argument list.

src/pre_process/m_initial_condition.fpp
```diff
@@ -262,7 +266,7 @@ contains
 
                     ! STL patch
                 elseif (patch_icpp(i)%geometry == 21) then
-                    call s_model(i, patch_id_fp, q_prim_vf)
+                    call s_model(i, patch_id_fp, q_prim_vf, .false.)
```
sbryngelson: see above comment about putting a raw boolean in the middle of the argument list.

src/pre_process/m_initial_condition.fpp
```diff
@@ -284,6 +288,9 @@ contains
                 elseif (patch_ib(i)%geometry == 4) then
                     call s_airfoil(i, ib_markers%sf, q_prim_vf, .true.)
```
sbryngelson: see above comment about putting a raw boolean in the middle of the argument list.

src/pre_process/m_initial_condition.fpp
```diff
@@ -284,6 +288,9 @@ contains
...
                     call s_compute_airfoil_levelset(levelset, levelset_norm, i)
+                    ! STL+IBM patch
+                elseif (patch_ib(i)%geometry == 5) then
+                    call s_model(i, ib_markers%sf, q_prim_vf, .true., levelset, levelset_norm)
```
sbryngelson: see above comment about putting a raw boolean in the middle of the argument list.

src/pre_process/m_initial_condition.fpp
sbryngelson: weird subroutine name. s_airfoil_XXX what does this subroutine do?

src/pre_process/m_initial_condition.fpp
```diff
@@ -351,6 +358,9 @@ contains
         deallocate (patch_id_fp)
         deallocate (ib_markers%sf)
 
+        ! deallocate (STL_levelset)
```
sbryngelson: remove?

src/pre_process/m_patches.fpp
```diff
@@ -1990,19 +2004,71 @@ contains
...
+        ! Interpolate the STL model along the edges (2D) and on triangle facets (3D)
+        if (interpolate) then
+            if (proc_rank == 0) then
+                print *, ""
+                print *, 'Interpolating STL vertices...'
+            end if
+
+            if (p > 0) then
+                call f_interpolate_3D(model, (/dx, dy, dz/), interpolated_boundary_v, total_vertices)
+            else
```
sbryngelson, line 2036: what if it's 1D?

src/pre_process/m_patches.fpp
```diff
@@ -1990,19 +2004,71 @@ contains
...
+        if (proc_rank == 0 .and. p == 0) then
+            print *, 'Number of 2D model edges:', 3*model%ntrs
+            print *, ""
+            print *, 'Number of 2D model boundary edges:', boundary_edge_count
+        end if
+
+        ! Interpolate the STL model along the edges (2D) and on triangle facets (3D)
+        if (interpolate) then
+            if (proc_rank == 0) then
+                print *, ""
```
sbryngelson: remove?

src/pre_process/m_patches.fpp
```diff
@@ -1990,19 +2004,71 @@ contains
...
+        if (p > 0) then
+            call f_check_interpolation_3D(model, (/dx, dy, dz/), interpolate)
+        else
+            call f_check_interpolation_2D(boundary_v, boundary_vertex_count, (/dx, dy, dz/), interpolate)
+        end if
+
+        ! Show the number of edges and boundary edges in 2D STL models
+        if (proc_rank == 0 .and. p == 0) then
+            print *, 'Number of 2D model edges:', 3*model%ntrs
+            print *, ""
```
sbryngelson: remove?

src/pre_process/m_initial_condition.fpp
haochey: It's an analytical airfoil patch. I believe @anandrdbz created this patch. Can I make a separate PR specifically for refactoring something like this later?

sbryngelson: please just fix it now, it should only change a couple of lines?

src/pre_process/m_compute_levelset.fpp
haochey: That was a bug and @anshgupta1234 can confirm this.

sbryngelson: it seems like @wilfonba is saying there were several changes in this subroutine. Were they all bugs?

haochey: > it seems like @wilfonba is saying there were several changes in this subroutine. Were they all bugs?

Yes

sbryngelson:
Please add more examples and flow visualizations to a `README.md` in the `example/<your example>/` directory, as is done for many other examples.

src/pre_process/m_check_ib_patches.fpp
```diff
@@ -227,6 +259,28 @@ contains
 
     end subroutine s_check_cylinder_ib_patch_geometry
 
+    !>  This subroutine verifies that the geometric parameters of
+        !!      the model patch have consistently been inputted by
+        !!      the user.
+        !!  @param patch_id Patch identifier
+    subroutine s_check_model_ib_patch_geometry(patch_id)
```
anshgupta1234, line 266: Maybe instead of the routines that say s_xxx_model_xxx, we should replace the terminology 'model' with 'stl'? os this would be `s_check_stl_ib_patch_geometry`

src/pre_process/m_compute_levelset.fpp
anshgupta1234: Yep, this subroutine had some bugs we had found earlier, which is why the analytical & stl patches weren't returning the same results. These changes fix it

src/pre_process/m_check_ib_patches.fpp
sbryngelson: if all cases of `model` are STL files then yes

src/common/m_constants.fpp
haochey: This is actually a multiple factor so this should be a constant. The final number of interpolated vertices depends on the ratio between the edge and the cell width or triangle area and the cell face area (e.g. in m_model.fpp line: 899. `num_segments = Ifactor_3D*ceiling(edge_length/cell_width)`).

I think my docstring is causing confusion so I will update it.

src/pre_process/m_check_ib_patches.fpp
henryleberre: We support both STL and OBJ so I had picked `model`

sbryngelson: if we _actually_ support OBJ we should have some examples with it (it would be great if we did!)

haochey:
I've made all the requested changes. It'll be great if it can get merged by the 1st MFC5.0 draft.

sbryngelson:
> Thanks for all the reviews. I've made the requested changes. It'll be great if it can get merged by the 1st MFC5.0 draft.

Great! I think I can merge, will have a final look today (or very soon). Thanks @haochey

sbryngelson (approved):

Issue #665, opened 2024-10-31
Support stretched grids for 7th-order reconstructions


sbryngelson:
Support stretched grids for 7th-order reconstructions: WENO/TENO7.

Some relevant papers for the smoothness indicators: 
* [arxiv.org/abs/2405.09367](https://arxiv.org/abs/2405.09367) 
* [wias-berlin.de/people/john/BETREUUNG/bachelor_rupp.pdf](https://www.wias-berlin.de/people/john/BETREUUNG/bachelor_rupp.pdf)

PR #743, opened 2024-11-19
Weno7 nonuniform

ChrisZYJ:
## Description

#665 MFC now supports stretched grids for 7th order WENO (-JS, -M, -Z). 
(Stretched grids for TENO7 are not ready yet)

#### Implementation Details

The WENO7 golden files were regenerated, which was necessary due to machine errors introduced by the large number of operations. The consistency of the new non-uniform formulas with the uniform formulas in uniform grid cases is demonstrated in the 1D Shu-Osher Uniform Grid plot. The old test suites also passed with increased tolerance.

The code contains ridiculously long equations, which have already reached their simplest form. I made a controversial decision to keep the lines long, as it improves code aesthetics (avoiding having pages of equations displayed while scrolling through) and because the equations are not human-readable anyway. Please let me know if alternative methods are preferred.

#### Equation Derivation

None of the equations from published papers are applicable. They are either approximations, don't solve the problems we need, or lack explicit forms. MFC also uses a very specific form that requires the coefficients to be for cell value differences rather than for the cell values themselves (this reduces computational costs). I had to derive the equations myself.

Non-uniform TENO7 requires different ideal weight formulas, which I need to derive. They aren't too complex, but I will complete them in another PR in the coming days.

### Type of change

### Scope

## How Has This Been Tested?

- [x] 1D Shu-Osher Uniform Grid
weno7_uniform_grid

- [x] 1D Shu-Osher Non-Uniform Grid
weno7_nonuniform_grid

- [x] Old Test Suites (with increased tolerance)

## Checklist


sbryngelson:
This is well done, I can already tell. I will do code review soon.

src/simulation/m_weno.fpp
```diff
@@ -447,20 +452,166 @@ contains
...
-                        d_cbL_${XYZ}$ (0, :) = 4d0/35d0
-                        d_cbL_${XYZ}$ (1, :) = 18d0/35d0
-                        d_cbL_${XYZ}$ (2, :) = 12d0/35d0
-                        d_cbL_${XYZ}$ (3, :) = 1d0/35d0
 
-                        d_cbR_${XYZ}$ (0, :) = 1d0/35d0
-                        d_cbR_${XYZ}$ (1, :) = 12d0/35d0
-                        d_cbR_${XYZ}$ (2, :) = 18d0/35d0
-                        d_cbR_${XYZ}$ (3, :) = 4d0/35d0
+                        do i = is%beg - 1 + weno_polyn, is%end - 1 - weno_polyn
```
sbryngelson, line 458: can you put a ref. as to where these come from?

sbryngelson (commented):
Are stretched grids included in the test suite? It seems perhaps not. Would be good to have. Codecov also seems to thing many lines of code are not tested.

sbryngelson:
> > Are stretched grids included in the test suite? It seems perhaps not. Would be good to have. Codecov also seems to thing many lines of code are not tested.
> 
> I think all added lines are covered; removing any line will cause the test suites to fail.
> 
> My guess is that Codecov has trouble correctly identifying lines in m_weno.fpp. In the past, Codecov has misidentified many covered lines as uncovered. You can see it [here](https://app.codecov.io/gh/MFlowCode/MFC/blob/master/src%2Fsimulation%2Fm_weno.fpp#L260).
> 
> There is only one set of equations for both non-uniform and uniform grids (same as WENO3/5). The non-uniform coefficients simplify to the uniform grid coefficients when the grids are uniform. Therefore, the existing test suites cover all the new equations.

🫡 you are correct on all counts. Thanks for your continued high-quality PRs!

sbryngelson (approved):

PR #749, opened 2024-11-28
Lagrangian subgrid bubble model

dgvacarevelo:
## Description

The Euler-Lagrange model for sub-grid bubble dynamics is based on the volume-averaged equations
of motion to describe the dynamics of a mixture of dispersed bubbles and a compressible liquid.

### Type of change

### Scope

## How Has This Been Tested?

- [x] Test A: Bubble screen problem
A cuboidal bubble cloud interacts with a sinusoidal planar wave. This case is included as an example case (examples/3D_lag_bubbles_bubblescreen). The pressure at the center of the cloud is measured and compared, as follows.

image

- [x] Test B: Singel bubble collapse
A single bubble of 8 micrometers encounters a sinusoidal planar wave. This case is also included as an example case (examples/3D_lag_bubbles_shbubcollapse). The radius evolution of the bubble is compared with the experiment reported by (Ohl et. al, 1999) (https://doi.org/10.1098/rsta.1999.0327).

image

**Test Configuration**:

1. Delta: I tested both problems with CPUs (gcc/11.4.0, openmpi/4.1.6) and GPUs (nvhpc/24.1, openmpi/4.1.5+cuda, cuda/12.3.0, cmake/3.27.9). I used the NVIDIA A100 to run my GPU cases.
2. Turing (WPI HPC): Here, I used the following CPU compilers: gcc/12.1.0, openmpi/4.1.3, python/3.11.6

## Checklist

- [x] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled. _Here I only had access to NVIDIA hardware with NVHPC compilers._


sbryngelson:
Thanks! This is failing some tests, like Frontier w/ CCE18 and AMD GPUs (MI250X). I'm not sure if you have access to this machine to debug yourself @dgvacarevelo

dgvacarevelo:
@sbryngelson No, I don't have access to Frontier. I implemented the GPU acceleration using DELTA through ACCESS resources.

sbryngelson:
Requesting code review from @anshgupta1234 @wilfonba @haochey @ChrisZYJ

src/simulation/m_acoustic_src.fpp
```diff
@@ -159,6 +159,7 @@ contains
         integer, parameter :: mass_label = 1, mom_label = 2
 
         sim_time = t_step*dt
+        if (lag_bubbles) sim_time = time_tmp
```
wilfonba: I'm not sure I understand what the purpose of this assignment is. Could you elaborate a little?

src/simulation/m_global_parameters.fpp
```diff
@@ -670,12 +710,37 @@ contains
             integral(i)%ymax = dflt_real
         end do
 
-        ! GRCBC flags
-        #:for dir in {'x', 'y', 'z'}
-            bc_${dir}$%grcbc_in = .false.
-            bc_${dir}$%grcbc_out = .false.
-            bc_${dir}$%grcbc_vel_out = .false.
-        #:endfor
```
wilfonba: Were these lines added back somewhere else? I'm not seeing them anywhere

src/simulation/m_kernel_functions.fpp
```diff
@@ -0,0 +1,501 @@
+!>
+!! @file m_kernel_functions.f90
+!! @brief Contains module m_kernel_functions
+
+#:include 'macros.fpp'
+
+!> @brief This module contains kernel functions used to map the effect of the lagrangian bubbles
+!!        in the Eulerian framework.
+module m_kernel_functions
```
wilfonba: It might be worthwhile to add something to this name to indicate it's for Lagrangian bubbles. Maybe `m_lag_bub_kernel_functions` or something along those lines?

src/simulation/m_start_up.fpp
```diff
@@ -1397,6 +1428,9 @@ contains
     end subroutine s_save_data
 
     subroutine s_initialize_modules
+
+        integer :: i,j,k,l
```
wilfonba: Are these used anywhere?

wilfonba (commented):
Really nice and well put together PR @dgvacarevelo! I made a few minor comments, but overall every thing looks good.

src/simulation/m_global_parameters.fpp
```diff
@@ -670,12 +710,37 @@ contains
...
+        ! Lagrangian subgrid bubble model
+        lag_bubbles = .false.
+        lag_write_bubbles = .false.
+        lag_write_bubble_stats = .false.
+        lag_bubble_model = dflt_int
```
sbryngelson: i think these should be in a derived type instead of `lag_XYZ`

src/simulation/m_global_parameters.fpp
```diff
@@ -982,6 +1047,9 @@ contains
         if (qbmm .and. .not. polytropic) then
             allocate (MPI_IO_DATA%view(1:sys_size + 2*nb*4))
             allocate (MPI_IO_DATA%var(1:sys_size + 2*nb*4))
+        elseif (lag_bubbles) then
```
sbryngelson: perhaps we should change the code logic/branching to bubbles -> "Euler" or "Lagrange" ->  (if euler [set options], else if lagrange [set options]). ultimately i'd like to see the 'bubbles' keyword be the first class term for sub-grid bubbles, and then we differentiate down from there.

src/simulation/m_kernel_functions.fpp
```diff
@@ -0,0 +1,501 @@
...
+            end if
+
+            !Update void fraction field
+            addFun1 = strength_vol/Vol
+            !$acc atomic update
+            updatedvar%vf(1)%sf(cell(1), cell(2), cell(3)) = updatedvar%vf(1)%sf(cell(1), cell(2), cell(3)) + addFun1
+
+            !Update time derivative of void fraction
+            addFun2 = strength_vel/Vol
+            !$acc atomic update
```
sbryngelson: calling attention to this line. might need careful attention for CCE+AMD GPU.

src/simulation/m_kernel_functions.fpp
```diff
@@ -0,0 +1,501 @@
...
+            if (num_dims == 2) then
+                Vol = dx(cell(1))*dy(cell(2))*lag_charwidth
+                if (cyl_coord) Vol = dx(cell(1))*dy(cell(2))*y_cc(cell(2))*2d0*pi
+            else
+                Vol = dx(cell(1))*dy(cell(2))*dz(cell(3))
+            end if
+
+            !Update void fraction field
+            addFun1 = strength_vol/Vol
+            !$acc atomic update
```
sbryngelson: calling attention to this line. might need careful attention for CCE+AMD GPU.

src/simulation/m_start_up.fpp
```diff
@@ -165,7 +167,15 @@ contains
             k_x, k_y, k_z, w_x, w_y, w_z, p_x, p_y, p_z, &
             g_x, g_y, g_z, n_start, t_save, t_stop, &
             cfl_adap_dt, cfl_const_dt, cfl_target, &
-            viscous, surface_tension
+            viscous, surface_tension, &
+            lag_bubbles, lag_write_bubbles, lag_write_bubble_stats, &
+            lag_bubble_model, lag_cluster_type, lag_heatTransfer_model,&
```
sbryngelson: yeah all these definitely need to be wrapped in a derived type

src/simulation/m_start_up.fpp
```diff
@@ -1259,17 +1270,29 @@ contains
...
-        ! Total-variation-diminishing (TVD) Runge-Kutta (RK) time-steppers
-        if (time_stepper == 1) then
-            call s_1st_order_tvd_rk(t_step, time_avg)
-        elseif (time_stepper == 2) then
-            call s_2nd_order_tvd_rk(t_step, time_avg)
-        elseif (time_stepper == 3 .and. (.not. adap_dt)) then
-            call s_3rd_order_tvd_rk(t_step, time_avg)
-        elseif (time_stepper == 3 .and. adap_dt) then
-            call s_strang_splitting(t_step, time_avg)
+        if(lag_bubbles) then
```
sbryngelson: is this option really 'forced' to a user using E-L bubbles? no clean and easy way to leave the user with the fall back option of a regular time stepper? if we can give them all options, that would be very handy (and we can also abstract out the lag_bubble part of the time stepper routine here into its own subroutine).

src/simulation/m_time_steppers.fpp
```diff
@@ -258,6 +268,20 @@ contains
             @:ACC_SETUP_SFs(rhs_vf(i))
         end do
 
+        ! Allocating the cell-average RHS time-stages for adaptive RKCK stepper
+        if (lag_bubbles) then
+            @:ALLOCATE(rhs_ts_adapt(1:6))
```
sbryngelson: curious if this 1:6 notation works for this data type on Frontier. i think maybe not, not 100% sure though.

src/simulation/m_time_steppers.fpp
```diff
@@ -258,6 +268,20 @@ contains
...
+            do i = 1, 6
```
sbryngelson: `6` is a magic number here. please fix 😄

src/simulation/m_time_steppers.fpp
```diff
@@ -258,6 +268,20 @@ contains
...
+                @:ALLOCATE(rhs_ts_adapt(i)%vf(1:sys_size))
+            end do
+            do i = 1, 6
```
sbryngelson: magic number 6

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
     end subroutine s_time_step_cycling
+
+    !> (Adaptive) 4th/5th order Runge—Kutta–Cash–Karp (RKCK) time-stepping algorithm (Cash J. and Karp A., 1990)
+        !!      Method for initial value problems with rapidly varying RHS. A maximum error between the 4th and 5th
+        !!      order Runge-Kutta-Cash-Karp solutions for the same time step size is calculated. If the error is
+        !!      smaller than a tolerance, then the algorithm employs the 5th order solution, while if not, both
+        !!      eulerian/lagrangian variables are re-calculated with a smaller time step size.
+        !! @param t_step Current time-step
+        !! @param hdid Advanced time increment (adaptive time stepping)
+    subroutine s_4th_5th_order_rkck(t_step, hdid, time_prev, time_avg)
```
sbryngelson: same comment about time stepper options for EL.

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        real(kind(0.d0)) :: newtime, errmax_glb, dttarget, qtime
+        real(kind(0.d0)) :: RKh, RKh_glb, htemp, SAFETY = 0.9d0, PGROW = -0.2d0, &
+                            PSHRNK = -0.25d0, ERRCON = 1.89d-4
+        integer :: i, j, k, l, q, RKstep
+
+        real(kind(0.0d0)) :: maxDV_rhs, minDV_rhs
+        real(kind(0.0d0)), dimension(sys_size) :: maxQcons, minQcons
+        integer, dimension(3) :: cellDV_max
+
+        RKh = min(dt, dt_max)
```
sbryngelson: these are not in a GPU routine but are doing (perhaps?) computation that should live on the device (GPU). might need an OpenACC directive? perhaps even `!$acc kernels`?

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+
+        if (cfl_dt) then
+            ! Truncation Error
+#ifdef DEBUG
+            if (proc_rank == 0) print *, 'Computing truncation error (4th/5th RKCK)'
+#endif
+            call s_calculate_rkck_truncation_error()
+        end if
+
+502     if (lag_largestep) then ! Encountered negative radius, so reduce dt and restart time step
```
sbryngelson: tag `502`? remove?

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+            if (cfl_dt) then
+                if (RKh > 1.0d-14) then
```
sbryngelson: 1e-14 magic number. move to m_constants

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        if (cfl_dt) then ! Checking truncation error
+            lag_errmax = min(lag_errmax, 1.0d0)
+            if (num_procs > 1) then
+                call s_mpi_allreduce_max(lag_errmax, errmax_glb)
+                lag_errmax = errmax_glb
+            end if
+            lag_errmax = lag_errmax/lag_rkck_tolerance ! Scale relative to user required tolerance.
+
+            if ((lag_errmax > 1.0d0)) then   ! Truncation error too large, reduce dt and restart time step
+                htemp = SAFETY*RKh*((floor(lag_errmax*1.0d05)/1.0d05)**PSHRNK)
```
sbryngelson: magic numbers here? move to `m_constants` where and however best possible

sbryngelson (changes requested):
Given the power, complexity, and utility of this feature and its code, this is a very very good PR. but, it still needs some attention. We will try to take care of some of the Frontier/GPU stuff on our side, but for the rest do have a look at on your own @dgvacarevelo). [it's also a couples of orders of magnitude higher quality than the original PR for this feature, you're a quick study.]

src/simulation/m_lag_bubbles.fpp
sbryngelson, line 1: we already have a module called `m_bubbles`. i think (like a previous comment), we need to differentiate at a higher level now that we have EE an EL capability. perhaps `m_bubbles` can just handle a bit of high-level logic but after that most of the code currently in `m_bubbles` should be put in a new modules called `m_bubles_EE` or `m_bubbles_euler` (or some such). Similarly, this module might be called `m_bubbles_lagrange`.

src/simulation/m_lag_bubbles.fpp
```diff
@@ -0,0 +1,1746 @@
...
+            save_count = t_step_start
+            qtime = t_step_start*dt
+        end if
+
+        if (save_count == 0) then
+            if (proc_rank == 0) print *, 'Reading lag_bubbles input file at save_count: ', save_count
+            inquire (file='input/lag_bubbles.dat', exist=file_exist)
+            if (file_exist) then
+                open (unit=85, file='input/lag_bubbles.dat', form='formatted')
+101             read (85, *, end=102) (inputparticle(i), i=1, 8)
```
sbryngelson: identifier numbers here? i'd like to to stay away with this since we use it no where else in the code and I'm not sure what it might break (now or in the future). can we do without these?

src/simulation/m_lag_bubbles.fpp
```diff
@@ -0,0 +1,1746 @@
...
+            end do
+        end do
+
+    end subroutine s_add_sources
+
+    !>  This subroutine computes the Keller-Miksis equation
+        !! @param nparticles Particle identifier
+        !! @param step Current time-stage in the 4th/5th order Runge—Kutta–Cash–Karp time-stepping algorithm
+        !! @param q_prim_vf Primitive variables
+    subroutine s_compute_KM(nparticles, step, q_prim_vf)
```
sbryngelson: I would have to imagine that some of this code duplicates what is in `m_bubbles` for computing Keller--Miksis. is there a way to consolidate the bubble models for the radial dynamics down to one routine?

src/simulation/m_acoustic_src.fpp
dgvacarevelo: Thank you @wilfonba for your comments. The s_4th_5th_order_rkck time stepper can indeed be used with an adaptive functionality, meaning that the time step (dt) may vary. As a result, the physical time will not necessarily be equal to t_step*dt.

src/simulation/m_global_parameters.fpp
dgvacarevelo: My apologies for the oversight. Those lines were unintentionally removed, and I am now adding them back.

src/simulation/m_kernel_functions.fpp
dgvacarevelo: I am updating the name from m_kernel_functions to m_bubbles_EL_kernel_functions, and I have also renamed the lagrangian bubbles module from m_lag_bubbles to m_bubbles_EL, in line with one of Spencer's suggestions.

src/simulation/m_start_up.fpp
dgvacarevelo: These variables were used for debugging purposes, and I am now removing them.

src/simulation/m_acoustic_src.fpp
sbryngelson: Do we have a test case in the test suite for adaptive time stepping with EL bubbles? Is the adaptive time stepping different from the adaptive time stepping already in the code?

dgvacarevelo: Yes, I added two test cases that use the adaptive RKCK stepper. The first one uses a one-way EL coupling approach and the other uses a two-way EL coupling approach. 

Addressing your second question: The key difference lies in the method used to adjust the time step (dt). The existing implementation calculates dt using the CFL number from the Eulerian framework (please correct me if I'm wrong). On the other hand, the adaptive RKCK stepper, calculates dt based on the truncation error of the Lagrange variables between the 4th and 5th order RKCK time stepper solutions.

sbryngelson: Yes this sounds right to me. So do you have an example in the test suite that uses the 'regular' RK3 (possible adaptive one-way) time stepper (and another that uses RKCK)?

src/simulation/m_kernel_functions.fpp
sbryngelson: `m_bubbles_EL_kernel_functions` is a bit of a mouthful, perhaps `m_bubbles_EL_kernels` is better. thanks!

src/simulation/m_acoustic_src.fpp
dgvacarevelo: Currently, the test cases are only utilizing the RKCK algorithm. Two of the cases employ the adaptive RKCK stepper, while the other two use the RKCK stepper with a constant time step. I am currently working on porting the Lagrange bubbles subroutines into the standard RK1, RK2, and RK3 steppers, which will address one of your comments.

sbryngelson: perfect, I appreciate it! i think having as much flexibility will be handy

src/simulation/m_lag_bubbles.fpp
```diff
@@ -0,0 +1,1746 @@
...
+                (lag_bub_motion_pos(k, 1, 2) /= lag_bub_motion_pos(k, 1, 2))) then   ! finite bubble location
+
+                print *, 'Large time step. Radius from:', radiusOld, ' to :', lag_bub_interface_rad(k, 2), &
+                    '; and velocity from', velOld, ' to :', lag_bub_interface_vel(k, 2)
+                lag_largestep = .true.
+                if (dt < 2.d-15) then
+                    print *, 'WARNING large step', dt, '. Removing bubble:', lag_bub_id(k, 1)
+                    call s_remove_lag_bubble(k)
+                end if
+                goto 711
```
anandrdbz: Can't do this on GPUs as the outer loop (k) isn't sequential (goto and continue)

src/simulation/m_lag_bubbles.fpp
```diff
@@ -0,0 +1,1746 @@
...
+
+        integer :: i, j, k, l, q, ierr
+        logical :: change, indomain
+        integer, dimension(3) :: cell
+        real(kind(0.d0)) radiusOld, velOld
+        logical :: aux
+
+        call s_transfer_data_to_tmp()
+
+        !$acc parallel loop gang vector default(present) reduction(IOR: lag_largestep) private(k) copyin(RKstep)
```
anandrdbz: the reduction operation implies lag_largestep is a private variable, but seems like it's a global scalar. Might be better off using a new local variable to the function that's implicitly private

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+
+502     if (lag_largestep) then ! Encountered negative radius, so reduce dt and restart time step
+            if (cfl_dt) then
+                if (RKh > 1.0d-14) then
+                    RKh = RKh/2.0d0
+                    if (proc_rank == 0) print *, '>>>>> WARNING: Reducing dt and restarting time step, now dt: ', RKh
+                    lag_largestep = .false.
+                    dt = RKh
+                    !$acc update device(lag_largestep, dt)
+                    goto 501
```
sbryngelson: no `goto`

sbryngelson (changes requested):
remove all `goto` statements

src/simulation/m_global_parameters.fpp
dgvacarevelo: I have added a derived type to wrap up all the user parameters needed for the Lagrange bubbles.

src/simulation/m_start_up.fpp
dgvacarevelo: I have defined time_stepper == 4 to activate the RKCK time stepper, and added the Lagrange subroutines into the regular TVD RK steppers. With this, I have tested these changes using the bubble screen problem using different time-steppers.
Comparison-PressureEvolution

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: I have removed all goto and continue statements. I need a small clarification for this specific comment, when you say identifier numbers, you mean the '101' and '102' numbers right? or do you refer to the file unit identifier number (85)?

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: Thanks @anandrdbz. I have removed the goto and continue statements.

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: Thanks for the information, I am changing this too.

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: Here, are you giving me free pass to go into your m_bubbles module to make all these modifications (I don't want to be disrespectful with your work)? If so, I agree with what you are proposing. I would also add the bubble models in the high-level module since they can be used in the EE and EL capabilities.

sbryngelson: i'm suggesting you rename `m_bubbles -> m_bubbles_EE` and then create a new module called `m_bubbles` for the high-level logic that is common to both EE and EL models or points the code to one model or the other. If you have to make some changes in the `m_bubbles_EE` that is OK of course.

src/simulation/m_lag_bubbles.fpp
sbryngelson: Right, I mean `101/102` here. Please just follow the usual MFC syntax here.

src/simulation/m_time_steppers.fpp
sbryngelson: to be fixed still

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        !! @param hdid Advanced time increment (adaptive time stepping)
+    subroutine s_4th_5th_order_rkck(t_step, hdid, time_prev, time_avg)
+
+        integer, intent(in) :: t_step
+        real(kind(0d0)), intent(inout) :: hdid
+        real(kind(0d0)), intent(in) :: time_prev
+        real(kind(0d0)), intent(inout) :: time_avg
+
+        real(kind(0.d0)) :: newtime, errmax_glb, dttarget, qtime
+        real(kind(0.d0)) :: RKh, RKh_glb, htemp, SAFETY = 0.9d0, PGROW = -0.2d0, &
```
sbryngelson: safety, pgrow, pshrnk, errcon should go in `m_constants`. sorry I didn't catch this earlier.

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev + 0.2d0*dt
+        RKstep = 2
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 2nd time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(2)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(2)%vf, RKstep)
+        call s_update_tmp_rkck(RKstep, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev + 0.3d0*dt
+        RKstep = 3
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 3rd time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(3)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(3)%vf, RKstep)
+        call s_update_tmp_rkck(RKstep, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev
+        RKstep = 1
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 1st time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(1)%vf, q_prim_vf, rhs_ts_adapt(1)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(1)%vf, q_prim_vf, rhs_ts_adapt(1)%vf, RKstep)
+        call s_update_tmp_rkck(RKstep, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev + 0.6d0*dt
+        RKstep = 4
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 4th time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(4)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(4)%vf, RKstep)
+        call s_update_tmp_rkck(RKstep, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev + 1.0d0*dt
+        RKstep = 5
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 5th time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(5)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(5)%vf, 5)
+        call s_update_tmp_rkck(5, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        time_tmp = time_prev + 0.875d0*dt
+        RKstep = 6
+!$acc update device (time_tmp)
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 6th time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(6)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(6)%vf, 6)
+        call s_update_tmp_rkck(6, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
```
sbryngelson: remove goto

src/simulation/m_time_steppers.fpp
```diff
@@ -1041,6 +1246,17 @@ contains
 
         @:DEALLOCATE(rhs_vf)
 
+        ! Deallocating the cell-average RHS variable for adaptive method, Lagrangian solver
+        if (lag_bubbles) then
+            do i = 1, 6
```
sbryngelson: where does this 6 come from? it appears to be a magic number

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+            lag_errmax = min(lag_errmax, 1.0d0)
+            if (num_procs > 1) then
+                call s_mpi_allreduce_max(lag_errmax, errmax_glb)
+                lag_errmax = errmax_glb
+            end if
+            lag_errmax = lag_errmax/lag_rkck_tolerance ! Scale relative to user required tolerance.
+
+            if ((lag_errmax > 1.0d0)) then   ! Truncation error too large, reduce dt and restart time step
+                htemp = SAFETY*RKh*((floor(lag_errmax*1.0d05)/1.0d05)**PSHRNK)
+                RKh = sign(max(abs(htemp), 0.1d0*abs(RKh)), RKh)  ! No more than a factor of 10.
```
sbryngelson: 0.1d0 is a magic number, set it in `m_constant` with a name

src/simulation/m_time_steppers.fpp
sbryngelson: where does 1e5/1e5 come from? either way, if we need something like this it seems like a magic number that should go to `m_constants`

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+                if (proc_rank == 0) print *, '>>>>> WARNING: Truncation error found. Reducing dt and restaring time step, now dt: ', RKh
+                lag_largestep = .false.
+                dt = RKh
+                !$acc update device(lag_largestep, dt)
+                goto 501
```
sbryngelson: no goto

src/simulation/m_start_up.fpp
sbryngelson: to be removed still?

src/simulation/m_start_up.fpp
```diff
@@ -1209,15 +1219,16 @@ contains
...
-            if (cfl_const_dt .and. t_step == 0) call s_compute_dt()
+            if (cfl_const_dt .and. t_step == 0 .and. .not. lag_bubbles) call s_compute_dt()
 
-            if (cfl_adap_dt) call s_compute_dt()
+            if (cfl_adap_dt .and. .not. lag_bubbles) call s_compute_dt()
 
             if (t_step == 0) dt_init = dt
 
-            if (dt < 1d-3*dt_init .and. cfl_adap_dt .and. proc_rank == 0) then
+            if (dt < 1d-3*dt_init .and. cfl_adap_dt .and. proc_rank == 0 .and. .not. lag_bubbles) then
```
sbryngelson: can you do a find/replace on the whole code base to replace `lag_bubbles `with `EL_bubbles`

src/simulation/m_start_up.fpp
```diff
@@ -165,7 +167,15 @@ contains
...
             g_x, g_y, g_z, n_start, t_save, t_stop, &
             cfl_adap_dt, cfl_const_dt, cfl_target, &
-            viscous, surface_tension
+            viscous, surface_tension, &
+            lag_bubbles, lag_write_bubbles, lag_write_bubble_stats, &
+            lag_bubble_model, lag_cluster_type, lag_heatTransfer_model,&
+            lag_massTransfer_model, lag_nBubs_glb, csonhost, vischost, &
+            Thost, gammagas, gammavapor, pvap, cpgas, cpvapor, &
+            kgas, kvapor, Rgas, Rvap, diffcoefvap, sigmabubble, &
+            lag_rkck_tolerance, lag_smooth_type, lag_epsilonb, &
```
sbryngelson: `lag_XXXX` variables (except `lag_bubbles`) should be put into a derived type in my view so they are all encapsulated to one input variable. also makes for clearer code. so the derived type should be, for example `EL%wite_bubbles`, `EL%write_bubble_stats`, etc.

src/simulation/m_start_up.fpp
sbryngelson: so it looks like the method works fine with 2nd/3rd order RK? if so, it appears that the code still restricts EL bubbles to RK45?

src/simulation/m_time_steppers.fpp
sbryngelson: to be fixed still

src/simulation/m_time_steppers.fpp
sbryngelson: to be fixed still

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+        real(kind(0.d0)) :: RKh, RKh_glb, htemp, SAFETY = 0.9d0, PGROW = -0.2d0, &
+                            PSHRNK = -0.25d0, ERRCON = 1.89d-4
+        integer :: i, j, k, l, q, RKstep
+
+        real(kind(0.0d0)) :: maxDV_rhs, minDV_rhs
+        real(kind(0.0d0)), dimension(sys_size) :: maxQcons, minQcons
+        integer, dimension(3) :: cellDV_max
+
+        RKh = min(dt, dt_max)
+        RKh = max(Rkh, 1.0d-12)
```
sbryngelson: 1d-12 magic number

src/simulation/m_time_steppers.fpp
```diff
@@ -980,6 +1004,187 @@ contains
...
+#ifdef DEBUG
+        if (proc_rank == 0) print *, 'RKCK 2nd time-stage at', time_tmp
+#endif
+        call s_compute_rhs(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(2)%vf, pb_ts(1)%sf, rhs_pb, mv_ts(1)%sf, rhs_mv, t_step, time_avg)
+        call s_compute_el_coupled_solver(q_cons_ts(2)%vf, q_prim_vf, rhs_ts_adapt(2)%vf, RKstep)
+        call s_update_tmp_rkck(RKstep, q_cons_ts, rhs_ts_adapt, q_prim_vf)
+        if (lag_largestep) goto 502
+
+        ! Third time-stage
+        time_tmp = time_prev + 0.3d0*dt
```
sbryngelson: 0.3 magic number

sbryngelson:
Many of my new comments were actually on an outdated commit. @dgvacarevelo can you resolve (or comment) on such comments that you resolved in the latest commits?

wilfonba:
@dgvacarevelo You'll need to push with the `lag_bubbles.dat` files from your local copy if you want the examples to work. The files weren't added on the remote because they were targeted by `.gitignore`, but I've corrected that. This will likely break the examples testing on the two example cases, but I can resolve that.

src/simulation/m_global_parameters.fpp
dgvacarevelo: Fixed

src/simulation/m_kernel_functions.fpp
dgvacarevelo: Fixed

src/simulation/m_start_up.fpp
dgvacarevelo: Fixed

src/simulation/m_global_parameters.fpp
dgvacarevelo: Fixed. The derived type variable is called 'lag_params%XYZ'

src/simulation/m_global_parameters.fpp
dgvacarevelo: I created a the file ''m_bubbles" which contains the subroutines shared by both sub grid bubble models. The subroutines for the EE model is in "m_bubbles_EE" and the ones for the EL model is in "m_bubbles_EL". Additionally, I renamed the switch to activate the EL model to 'bubbles_lagrange', and renamed the switch for the EE model from 'bubbles' to 'bubbles_euler'. In the test suite, I didn't make any changes to the trace names, because I did not want to generate new golden files for the EE cases. My point is that I am using the original golden files to test all these changes. Please let me know if these changes address your comment.

src/simulation/m_kernel_functions.fpp
dgvacarevelo: This atomic function is a strategy I came across online to prevent the possibility of different threads simultaneously overwriting the same cell, which could occur if two bubbles are sufficiently close together.

src/simulation/m_kernel_functions.fpp
dgvacarevelo: This atomic function is a strategy I came across online to prevent the possibility of different threads simultaneously overwriting the same cell, which could occur if two bubbles are sufficiently close together.

src/simulation/m_start_up.fpp
dgvacarevelo: fixed

src/simulation/m_start_up.fpp
dgvacarevelo: fixed :) I defined a time_stepper = 4 to activate the rkck algorithm.

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: I am using the GPUs to calculate the maximum truncation error with a reduce function, then the cpus calculate the next dt and finally, the new dt is updated into the GPUs. If you have any comments here to improve this process, please let me know.

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: addressed in one of the above comments.

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: fixed

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: Yes, there was code duplication here. I was able to utilize what was already available for the EE model. Fixed

src/simulation/m_acoustic_src.fpp
sbryngelson: update on this would be useful.. was this resolved or are there questions still? I've kind of gotten confused

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: fixed

src/simulation/m_lag_bubbles.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: 6 is the number of time stages needed for the rkck stepper. All the needed constants were placed in m_constants.

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: I found the neceessity to round the relative truncation error (5th decimal) to avoid the inclusion of random decimals after dividing by a very small number (rkck_tolerance). Fixed.

src/simulation/m_time_steppers.fpp
sbryngelson: i don't know why i thought this was an issue. resolving

src/simulation/m_start_up.fpp
dgvacarevelo: I renamed the EL switch to bubbles_lagrange and also updated the original flag of the EE model from bubbles to bubbles_euler. Please confirm me if you want me to do some additional changes here.

src/simulation/m_start_up.fpp
dgvacarevelo: fixed

src/simulation/m_time_steppers.fpp
dgvacarevelo: fixed

src/simulation/m_acoustic_src.fpp
dgvacarevelo: Now the highlighted line command gets activated when 'time_stepper==4' (rkck switch). This command modifies the physical time, which will not necessarily be equal to t_step*dt if the adaptive functionality is activated. 

This is the update for the test cases: 

- Two cases employ the 3rd RK stepper: one of them uses of the one-way EL coupling approach, and the other uses the two-way EL coupling approach
- Two cases uses the adaptive RKCK stepper: one of them makes use of the one-way EL coupling approach, and the other uses the two-way EL coupling approach

sbryngelson: sounds perfect. thanks! resolving.

sbryngelson (approved):

PR #758, opened 2024-12-15
Chemistry Temperature Optimization

henryleberre:
## Description

- Removes the `T_idx` temperature equation. We were not solving the RHS, just using that index as temporary storage.
- Instead, adds a temperature scalar field we keep around and update. The values it holds are used as the initial guess when we solve for the temperature in the next stage.

Note: In the future, we could convert this one scalar field to a vector of scalar fields to support other features that require it. We would then define indices for each element of this array, like `T_idx = 1` for temperature if chemistry is on.

### Type of change

### Scope


sbryngelson:
Looks quite good, can you add some comments to the subroutines just in case?

henryleberre:
Yes, will do. I am also looking into the NaNs issue…

henryleberre:
@sbryngelson should be good to go now.

henryleberre:
@sbryngelson rebased in the hopes the GT runners run.

sbryngelson (approved):

PR #768, opened 2024-12-22
output_partial_domain for post-processing

ChrisZYJ:
## Description

Adds an option to output only part of the domain during post-processing.

### Type of change

### Scope

## How Has This Been Tested?

- [x] Works as expected when post-processed using Matlab binary reader.
- [x] Passes the original tests with post-processing enabled (with -a)

## Checklist


sbryngelson:
@ChrisZYJ thanks this appears very useful! I'm curious if this works for a couple different scenarios (have you verified):
* Serial vs serial output
* Format == 1 and 2
* Stretched grids

ChrisZYJ:
It works with stretched grids and both serial and parallel outputs. However, it only works for format == 2 (binary) and not format == 1 (silo). The silo format requires more careful data handling, so perhaps that could be completed in another PR. 

I've documented the scenarios where it doesn't work yet in the documentation and added prohibitions in the checkers.

sbryngelson:
@ChrisZYJ thanks this makes sense. May I ask how you are handling your binary output files? Matlab/Python or into Visit/Paraview? I think this would be useful to either document or perhaps even put a helper script in the `misc/` directory for others.

ChrisZYJ:
I've been doing it in Matlab using the `f_binary_reader.m` function, which is already in the `/misc` directory. I could document it and add my wrapper function to `/misc` later if needed.

sbryngelson (approved):

sbryngelson:
Thanks @ChrisZYJ. I find using the binary files rather annoying. If we had a more broad way of doing it (and could suggest to the user when and how to do it) that would be useful.

I'm glad to merge the PR but we have two big PRs that are very near merging that I don't want to disrupt. Hopefully can merge this in a couple days.

ChrisZYJ:
Yeah, I also find binary files quite cumbersome to work with. While I'm not sure what a broader way to handle them would be, I can add some documentation and helper scripts in another PR if you find those helpful.

I'm in no hurry to merge. This PR should be quite independent from everything else, and I could easily merge it into them later if needed.

sbryngelson:
> Yeah, I also find binary files quite cumbersome to work with. While I'm not sure what a broader way to handle them would be, I can add some documentation and helper scripts in another PR if you find those helpful.
> 
> I'm in no hurry to merge. This PR should be quite independent from everything else, and I could easily merge it into them later if needed.

That would be very useful indeed. I hope to merge this soon but don't want to risk messing up these larger long-standing PRs right now.

PR #727, opened 2024-11-13
Hypoelasticity to HLLC, preliminary features for RMT

mrodrig6:
## Description

Please include a summary of the changes and the related issue(s) if they exist.
Please also include relevant motivation and context.

Fixes #(issue) [optional]

### Type of change

Please delete options that are not relevant.

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Something else

### Scope

- [x] This PR comprises a set of related changes with a common goal

If you cannot check the above box, please split your PR into multiple PRs that each have a common goal.

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. 
Provide instructions so we can reproduce. 
Please also list any relevant details for your test configuration

- [ ] Test A
- [ ] Test B

**Test Configuration**:

* What computers and compilers did you use to test this:

## Checklist

- [ ] I have added comments for the new code
- [ ] I added Doxygen docstrings to the new code
- [ ] I have made corresponding changes to the documentation (`docs/`)
- [ ] I have added regression tests to the test suite so that people can verify in the future that the feature is behaving as expected
- [ ] I have added example cases in `examples/` that demonstrate my new feature performing as expected.
They run to completion and demonstrate "interesting physics"
- [ ] I ran `./mfc.sh format` before committing my code
- [ ] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled
- [x] This PR does not introduce any repeated code (it follows the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
- [x] I cannot think of a way to condense this code and reduce any introduced additional line count

### If your code changes any code source files (anything in `src/simulation`)

To make sure the code is performing as expected on GPU devices, I have:
- [x] Checked that the code compiles using NVHPC compilers
- [ ] Checked that the code compiles using CRAY compilers
- [ ] Ran the code on either V100, A100, or H100 GPUs and ensured the new feature performed as expected (the GPU results match the CPU results)
- [ ] Ran the code on MI200+ GPUs and ensure the new features performed as expected (the GPU results match the CPU results)
- [ ] Enclosed the new feature via `nvtx` ranges so that they can be identified in profiles
- [ ] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this PR
[...]

PR #767, opened 2024-12-21
Hyperelastic RMT code with updated modular precision

mrodrig6:
### Type of change

- [x] New feature (non-breaking change which adds functionality)
- [x] Something else

### Scope

- [x] This PR comprises a set of related changes with a common goal

If you cannot check the above box, please split your PR into multiple PRs that each have a common goal.

## How Has This Been Tested?

Please describe the tests that you ran to verify your changes. 

Tested the code on NCSA Delta and passed the full test suite and most example cases.

Provide instructions so we can reproduce. 
Ran on 2 GPUs with one CPU each on an interactive node
./mfc.sh test -j 8 --max-attempt 3 -- -c delta

**Test Configuration**:

* What computers and compilers did you use to test this:

## Checklist

- [ ] I have added comments for the new code
- [ ] I added Doxygen docstrings to the new code
- [ ] I have made corresponding changes to the documentation (`docs/`)
- [ ] I have added regression tests to the test suite so that people can verify in the future that the feature is behaving as expected
- [ ] I have added example cases in `examples/` that demonstrate my new feature performing as expected.
They run to completion and demonstrate "interesting physics"
- [x] I ran `./mfc.sh format` before committing my code
- [x] New and existing tests pass locally with my changes, including with GPU capability enabled (both NVIDIA hardware with NVHPC compilers and AMD hardware with CRAY compilers) and disabled
- [x] This PR does not introduce any repeated code (it follows the [DRY](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) principle)
- [x] I cannot think of a way to condense this code and reduce any introduced additional line count

### If your code changes any code source files (anything in `src/simulation`)

To make sure the code is performing as expected on GPU devices, I have:
- [x] Checked that the code compiles using NVHPC compilers
- [ ] Checked that the code compiles using CRAY compilers
- [x] Ran the code on either V100, A100, or H100 GPUs and ensured the new feature performed as expected (the GPU results match the CPU results)
- [ ] Ran the code on MI200+ GPUs and ensure the new features performed as expected (the GPU results match the CPU results)
- [ ] Enclosed the new feature via `nvtx` ranges so that they can be identified in profiles
- [ ] Ran a Nsight Systems profile using `./mfc.sh run XXXX --gpu -t simulation --nsys`, and have attached the output file (`.nsys-rep`) and plain text results to this PR
[...]

PR #773, opened 2024-12-24
Fix hypoelastic instability

### LINKED ISSUE #771: Instability in Hypoelastic Simulations
### Description

Commit 10b2209 (#727) introduces instability in some hypoelastic simulations compared to previous commit 108805c7.

### Test Case

- Simulation: Simple acoustic wave propagation with sound waves from left focused at a centered stone

- Case file: [case_hypo.txt](https://github.com/user-attachments/files/18234216/case_hypo.txt)

#### Step 300 of the simulation (with commit 10b2209):
image

#### Comparison with the previous commit 108805c7 (top:108805c7 ; bottom:10b2209):
https://github.com/user-attachments/assets/3b05a2a1-08b2-4e0c-ac97-44c6df4f97ae

### Environment & Configuration
- Ubuntu 22.04.4 LTS
- GNU 11.4.0
- Using CPU with both debug enabled and disabled

ChrisZYJ:
## Description

Fixes #771 by reverting the hypoelasticity changes from #727
Adds a test case to prevent similar issues in the future
Also regenerates the hypoelasticity-related test that was generated during #727

### Type of change

### Scope

## How Has This Been Tested?

A new test case has been added that covers a scenario similar to #771 and is more comprehensive than existing tests.

The test was generated using commit 108805c7 (before #727), fails after the changes in #727, and passes after the changes in this PR.


sbryngelson:
@ChrisZYJ I understand why you want to do this. Part of the merged PR #727 was to set the 'groundwork' for pending PR #767 on hyperelasticity and RMT. So, I'm hesitant to merge this because it could mess up the work in #767. Can you work with @mrodrig6 to find a reasonable solution to everything and both PRs?

ChrisZYJ:
@sbryngelson Thanks for raising this! I’ve double-checked, and #767 doesn’t touch m_hypoelastic.fpp, so reverting the hypoelasticity changes shouldn’t conflict with the hyperelasticity/RMT updates. By fixing the stability issue now and adding a test to catch similar regressions, I think #767 would have a better 'groundwork'.

@mrodrig6 Could you please confirm there’s no hidden overlap we’re missing? Thanks!

mrodrig6:
> @sbryngelson Thanks for raising this! I’ve double-checked, and #767 doesn’t touch m_hypoelastic.fpp, so reverting the hypoelasticity changes shouldn’t conflict with the hyperelasticity/RMT updates. By fixing the stability issue now and adding a test to catch similar regressions, I think #767 would have a better 'groundwork'.
> 
> @mrodrig6 Could you please confirm there’s no hidden overlap we’re missing? Thanks!

Yes, that is correct. This does not overlap with the hyperelasticity/RMT updates.

ChrisZYJ:
@mrodrig6 Thank you very much for clarifying! The issue was completely resolved by reverting only m_hypoelastic, likely due to finite-difference stencils in m_helpers.

I agree that grid-stretching is an important feature to keep. May I know if you plan to address the stability issue while preserving grid-stretching in your ongoing commit #767? If so, I suggest merging this #773 first, as its test suites provide correct reference results for easier debugging. You can maintain the existing m_hypoelastic code in #767, and the tests will pass once the issue is fixed.

ChrisZYJ:
I'm not too familiar with the new FD stencil implementations, but I can look into them!

sbryngelson (approved):

PR #737, opened 2024-11-17
(Atomic) Boundary Condition Patches (pt. 1/2)

henryleberre:
## This part (pt. 1)

- Removes (net) ~4k LOC by abstracting the iteration over the buffer region.
- Adds (opt-in) "Boundary Condition Patches" (cuboid & spheroid).
- Adds `--dry-run` to testing so that CI can build all necessary versions of MFC before submitting jobs for testing to a queue system.
- Fixes an IBM method bug for slip-walls (credit @anshgupta1234).
- Fixes a CheMFC bug that prevented the use of analytically defined patches.

## Checklist for getting out of the "draft" stage

- [ ] Check GPU performance.
- [ ] Add spheroid boundary patch code back.
- [ ] Verify the boundary patches still work with any number of ranks.

## Next part (pt. 2)

- Documentation.
- All edge cases addressed. That is, `bc_[x,y,z]` should no longer be used anywhere in the code. One example of such an oversight in pt. 1 is the lack of support for CBCs.

## Why two parts?

\> 8k LOC changes is a lot to carry around in one _atomic_ commit.

PR #789, opened 2025-01-15
Fix slip BCs for IBM

sbryngelson:
This small PR grabs fixes from @henryleberre's PR #737 for IBMs with slip-boundary cases. It apparently came from a problem that @anshgupta1234 found. 

@haochey, can you have a look at the code and let me know if this makes sense or where it comes from?

Also, it appears that we don't have any slip IBM cases in the test suite. Is this true? If so, we need to add them.

### DIFF
--- a/src/simulation/m_ibm.fpp
+++ b/src/simulation/m_ibm.fpp
@@ -18,6 +18,8 @@ module m_ibm
 
     use m_helper
 
+    use m_constants
+
     implicit none
 
     private :: s_compute_image_points, &
@@ -39,7 +41,6 @@ module m_ibm
     type(ghost_point), dimension(:), allocatable :: inner_points
     !$acc declare create(ghost_points, inner_points)
 
-    integer :: gp_layers !< Number of ghost point layers
     integer :: num_gps !< Number of ghost points
     integer :: num_inner_gps !< Number of ghost points
     !$acc declare create(gp_layers, num_gps, num_inner_gps)
@@ -48,8 +49,6 @@ contains
 
     !>  Allocates memory for the variables in the IBM module
     subroutine s_initialize_ibm_module()
-
-        gp_layers = 3
 
         if (p > 0) then
             @:ALLOCATE(ib_markers%sf(-gp_layers:m+gp_layers, &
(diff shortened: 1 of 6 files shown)


haochey:
@sbryngelson I will look at this today.

src/simulation/m_ibm.fpp
```diff
@@ -209,7 +207,9 @@ contains
 
             ! Calculate velocity of ghost cell
             if (gp%slip) then
-                norm = levelset_norm%sf(j, k, l, patch_id, :)
+                norm = levelset_norm%sf(gp%loc(1), gp%loc(2), gp%loc(3), gp%ib_patch_id, :)
+                buf = sqrt(sum(norm**2))
```
haochey, line 211: I think these lines are redundant because `levelset_norm%sf(gp%loc(1), gp%loc(2), gp%loc(3), gp%ib_patch_id, :)` is already a norm vector. 

And `j = gp%loc(1)`,  `k = gp%loc(2)` ,  `l = gp%loc(3)` , so `norm = levelset_norm%sf(j, k, l, patch_id, :)` should be fine.

haochey:
@sbryngelson I just left one review comment. Other than that it looks fine and makes sense to me.

And it's true that we need some slip IBM test cases. I can either create another PR or add them to the test suite if you give me the permission to push to this branch.

sbryngelson:
@haochey I don't see any review comments, you probably didn't "post" them. And can you add them to the test suite after I merge?

haochey:
> @haochey, I don't see any review comments. You probably didn't "post" them. Can you add them to the test suite after I merge them? I will wait until PACE Phoenix is back up and running so CI all passes, probably tomorrow or Friday.

Yes, sounds good.

src/simulation/m_ibm.fpp
sbryngelson: I think I agree, but what about the other part (line 211 and 222 that have `buf = ...` and `norm = norm/buf`? that seems different/new

haochey: These two lines seem to normalize the `norm` to be unit vectors while `norm` are already unit vectors because levelset_norms' are unit vectors.

sbryngelson: i suspect this was not added by @henryleberre for no reason, is there any way to ensure levelset_norm subroutines _always_ normalize without just doing it here?

haochey: Make sense to me! It will rely on levelset subroutines to normalize them before output them. I think this is the reason why Henry added these two lines.

anandrdbz: Btw there's a corner case that needs to be taken care of, when norm = 0

sbryngelson: what do you suggest doing in this case, @anandrdbz?  You can write code to my branch/PR here.

anandrdbz: Okay I'll take a look, do I have push permissions ?

sbryngelson: I believe so.

PR #792, opened 2025-01-25
Checker Error Message

ChrisZYJ:
## Description

Makes case file error messages clearer and more obvious

image

### Type of change

- [x] Something else

### DIFF
--- a/.github/workflows/lint-source.yml
+++ b/.github/workflows/lint-source.yml
@@ -40,7 +40,7 @@ jobs:
 
     - name: Looking for junk code
       run: |
-        ! grep -iR -e '\.\.\.' -e '\-\-\-' -e '===' ./src/*
+        ! grep -iR -e '\.\.\.' -e '\-\-\-' -e '===' ./src/* | grep '!'
 
     - name: Looking for junk comments in examples
       run: |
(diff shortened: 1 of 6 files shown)


sbryngelson:
This was intentional. I don't like the use of long strings of characters because how they appears will depend on the width of your terminal and its font. It seems very old school and somehow bothers me. It also makes it harder to check for such patterns in the future, since I do it via `grep` in CI. I'm open to alternative ways to make the printing of case file errors more readable, but don't want to introduce long patterns of `----` or `=====` or `!!!!!!`. Does that make sense?

ChrisZYJ:
Thanks for teaching me good coding practices! I've moved the `---` and reverted my changes to your grep implementation. It now formats the Fortran output messages better and emphasizes them using colors in Mako. Please let me know if this is better.

.github/workflows/lint-source.yml
```diff
@@ -40,7 +40,7 @@ jobs:
 
     - name: Looking for junk code
       run: |
-        ! grep -iR -e '\.\.\.' -e '\-\-\-' -e '===' ./src/* | grep '!'
```
sbryngelson: what's the reason for this change?

sbryngelson: oh i see, you're just reverting the change it sounds like

ChrisZYJ: yes!

sbryngelson (approved):

PR #798, opened 2025-02-05
Fix sphere smoothing

wilfonba:
## Description

This PR fixes a bug in the smoothing function for spherical patches that was identified during IGR development. The smoothing was only being applied inside the nominal radius of the patch, resulting in a grid level discontinuity. The old code generates the following IC
old
Which shows smoothing only on the sphere's interior and a discontinuity on the outer surface. The new code generates the following IC
new
which shows smoothing on both sides of the nominal radius of the patch. The smoothing coefficient for both ICs is identical.

Fixes #(issue) [optional]

### Type of change

### Scope


wilfonba:
@haochey Says this is likely an issue for some other patches as well and he's going to resolve it in this PR.

sbryngelson:
@haochey let me know when this is ready

sbryngelson:
Glad to merge but it's now failing a bunch of tests?

haochey:
Yeah, I will fix this today.

haochey:
@sbryngelson This should be ready to merge.

PR #813, opened 2025-04-01
Fix #797

### LINKED ISSUE #797: Cannot have `stop` in accelerator region
PR #749 introduces several impermissible statements, at least on AMD GPU hardware and with Cray compilers (v >= 19). The problem is `stop` statements in the GPU kernels. Examples are here:

https://github.com/MFlowCode/MFC/blob/131b41e04828331803747e9235ccd77944a1aaca/src/simulation/m_bubbles_EL.fpp#L992
https://github.com/MFlowCode/MFC/blob/131b41e04828331803747e9235ccd77944a1aaca/src/simulation/m_bubbles_EL.fpp#L1127
https://github.com/MFlowCode/MFC/blob/131b41e04828331803747e9235ccd77944a1aaca/src/simulation/m_bubbles_EL.fpp#L1115
https://github.com/MFlowCode/MFC/blob/131b41e04828331803747e9235ccd77944a1aaca/src/simulation/m_bubbles_EL.fpp#L1102

and so on and so forth.

dgvacarevelo:
## Description

Removes 'stop' commands from the GPU regions, and relocate them in CPU regions.

Fixes #797 

### Type of change

### Scope

### DIFF
--- a/src/simulation/m_checker.fpp
+++ b/src/simulation/m_checker.fpp
@@ -330,8 +330,6 @@ contains
         @:PROHIBIT(bubbles_lagrange .and. file_per_process, "file_per_process must be false for bubbles_lagrange")
         @:PROHIBIT(bubbles_lagrange .and. n==0, "bubbles_lagrange accepts 2D and 3D simulations only")
         @:PROHIBIT(bubbles_lagrange .and. model_eqns==3, "The 6-equation flow model does not support bubbles_lagrange")
-        @:PROHIBIT(bubbles_lagrange .and. lag_params%cluster_type<=0, "cluster_type must be specified")
-        @:PROHIBIT(bubbles_lagrange .and. lag_params%cluster_type>=2 .and. lag_params%smooth_type/=1, "cluster_type=2 requires smooth_type=1")
     end subroutine s_check_inputs_bubbles_lagrange
 
     !> Checks miscellaneous constraints,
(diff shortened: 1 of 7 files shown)


src/simulation/m_checker.fpp
```diff
@@ -330,6 +330,8 @@ contains
         @:PROHIBIT(bubbles_lagrange .and. file_per_process, "file_per_process must be false for bubbles_lagrange")
         @:PROHIBIT(bubbles_lagrange .and. n==0, "bubbles_lagrange accepts 2D and 3D simulations only")
         @:PROHIBIT(bubbles_lagrange .and. model_eqns==3, "The 6-equation flow model does not support bubbles_lagrange")
+        @:PROHIBIT(bubbles_lagrange .and. lag_params%cluster_type<=0, "cluster_type must be specified")
```
sbryngelson: i think this line might pose a problem. what is the default cluster type? if it is 0 then this will always be tripped

dgvacarevelo: I realized that I missed including the 'bubbles_lagrange' logical in the prohibit condition in the other PR. I believe this might be the cause of the issue. The default value of the cluster_type, as defined in m_global_parameters, is dflt_int (-100), but the cluster_type should be >= 1 (user defined value).

sbryngelson:
@dgvacarevelo FYI the Phoenix runner won't work for a day or two, that machine is offline for maintenance.

PR #814, opened 2025-04-01
Fix reflective BC for stresses

ChrisZYJ:
## Description

Fixes symmetric boundary (`bc=-2`) conditions for stresses (elasticity simulations).

### Type of change

### Scope

## How Has This Been Tested?

- [x] Old (wrong)

https://github.com/user-attachments/assets/b0caba5a-01c6-47fd-8ae7-8c9aa19afec0

- [x] New (corrected)

https://github.com/user-attachments/assets/d1e2edd3-640e-40a5-9835-7676b1c459f8

## Checklist


ChrisZYJ:
@sbryngelson Do you know what's causing this error? It works locally and I haven't seen this before. I also didn't change any compilation settings or FFTW-related stuff. Thanks!

```
[ 62%] Performing configure step for 'fftw'
-- Configuring incomplete, errors occurred!
```

sbryngelson:
Try recompiling/re-running the runners. it could be that the run got killed by something. another possibility is that FFTW ended up with a bug in a new release or something, but this seems less likely.

sbryngelson:
If only Ubuntu runners are failing, I wonder if something went wrong with a new FFTW version grabbed by those Ubuntu runs. Looking at the workflow file, we have

```shell
      - name: Setup Ubuntu
        if:   matrix.os == 'ubuntu' && matrix.intel == false
        run: |
           sudo apt update -y
           sudo apt install -y cmake gcc g++ python3 python3-dev hdf5-tools \
                    libfftw3-dev libhdf5-dev openmpi-bin libopenmpi-dev
```
You could look at a previous PR to see if the version has changed in the last day or so. If MacOS is also failing then it is probably a different problem because we `brew install fftw` and `brew` and `apt` are surely not in perfect synchrony with their package versioning. That's one way to debug.

ChrisZYJ:
@sbryngelson Thank you very much for the tips! I've tried a few things and move the discussion over to Slack.

sbryngelson (approved):

PR #819, opened 2025-04-14
Boundary Condition Paches

okBrian:
## Description

Added Boundary Condition Patches (Rectangle, Circle, and Line Segment). 

### Type of change

### Scope

## How Has This Been Tested?

- [x] Tested with custom test cases that test the different boundary patches in each direction locally with GCC 14.2.1
- [x] Used the example 2D_jet to test in Phoenix, Frontier, and Locally. 

**Test Configuration**:
* GCC 14.2.1
* GT Phoenix
* Frontier

## Checklist

 [link to nsys report too large for the pr](https://drive.google.com/file/d/1y4cGD-LnamlLFvTagFuvtM90fP2u6ueM/view?usp=sharing)


src/common/include/inline_boundary_conditions.fpp
```diff
@@ -0,0 +1,280 @@
...
+                    -q_prim_vf(shear_BC_flip_indices(${DIR}$, i))%sf(${SRC}$)
+            end do
+        end if
+
+        if (hyperelasticity) then
+            q_prim_vf(xibeg + ${DIR}$-1)%sf(${DEST}$) = &
+                -q_prim_vf(xibeg + ${DIR}$-1)%sf(${SRC}$)
+        end if
+    end do
+#:enddef
```
wilfonba: I'm not sure if this code is still correct after #814, this should be verified

src/common/include/inline_boundary_conditions.fpp
```diff
@@ -0,0 +1,280 @@
...
+
+#:def COLOR_FUNC_BC(DEST, SRC)
+    do i = 1, num_dims + 1
+        do j = 1, buff_size
+            c_divs(i)%sf(${DEST}$) = c_divs(i)%sf(${SRC}$)
+        end do
+    end do
+#:enddef
```
wilfonba: I know I initially named this, but it should be renamed to `COLOR_FUNC_EXTRAPOLATION` for clarity.

src/common/m_boundary_common.fpp
```diff
@@ -0,0 +1,582 @@
...
+            @:ALLOCATE(bc_buffers(2,-1)%sf(-buff_size:m+buff_size,1:sys_size,0:p))
+            @:ALLOCATE(bc_buffers(2,1)%sf(-buff_size:m+buff_size,1:sys_size,0:p))
+            @:ACC_SETUP_SFs(bc_buffers(2,-1), bc_buffers(2,1))
+            if (p > 0) then
+                @:ALLOCATE(bc_buffers(3,-1)%sf(-buff_size:m+buff_size,-buff_size:n+buff_size,1:sys_size))
+                @:ALLOCATE(bc_buffers(3,1)%sf(-buff_size:m+buff_size,-buff_size:n+buff_size,1:sys_size))
+                @:ACC_SETUP_SFs(bc_buffers(3,-1), bc_buffers(3,1))
+            end if
+        end if
+#endif
```
wilfonba: I think you can remove the `#ifdef MFC_PRE_PROCESS` copy of these allocations, and discard the `#ifdef MFC_SIMULATION` surrounding these allocations. The `@:ALLOCATE` macro works in pre-process. Also, these should only be allocated if they're needed, ie if there's dirichlet BCs.

src/common/m_boundary_common.fpp
```diff
@@ -0,0 +1,582 @@
...
+        if (n > 0) then
+            deallocate (bc_buffers(2, -1)%sf)
+            deallocate (bc_buffers(2, 1)%sf)
+            if (p > 0) then
+                deallocate (bc_buffers(3, -1)%sf)
+                deallocate (bc_buffers(3, 1)%sf)
+            end if
+        end if
+
+        deallocate (bc_buffers)
```
wilfonba: Replace `deallocate` with `@:DEALLOCATE` and only deallocate if needed (see earlier comment about allocating).

src/common/m_constants.fpp
```diff
@@ -21,9 +21,11 @@ module m_constants
     integer, parameter :: fourier_rings = 5                       !< Fourier filter ring limit
     integer, parameter :: num_fluids_max = 10                     !< Maximum number of fluids in the simulation
     integer, parameter :: num_probes_max = 10                     !< Maximum number of flow probes in the simulation
-    integer, parameter :: num_patches_max = 10
+    integer, parameter :: num_patches_max = 25
+    integer, parameter :: num_bc_patches_max = 10
     integer, parameter :: pathlen_max = 400
     integer, parameter :: nnode = 4    !< Number of QBMM nodes
+    real(wp), parameter :: dflt_alf_factor = 10._wp
```
wilfonba: Remove this. It's not needed

src/common/m_constants.fpp
```diff
@@ -21,9 +21,11 @@ module m_constants
     integer, parameter :: fourier_rings = 5                       !< Fourier filter ring limit
     integer, parameter :: num_fluids_max = 10                     !< Maximum number of fluids in the simulation
     integer, parameter :: num_probes_max = 10                     !< Maximum number of flow probes in the simulation
-    integer, parameter :: num_patches_max = 10
+    integer, parameter :: num_patches_max = 25
```
wilfonba: Revert this to 10

src/common/m_mpi_common.fpp
```diff
@@ -51,6 +51,30 @@ contains
...
+        else
+            halo_size = -1 + buff_size*sys_size
+        end if
+
+        v_size = sys_size
+
+        allocate (q_prims_buff_send(0:halo_size))
+
+        allocate (q_prims_buff_recv(0:ubound(q_prims_buff_send, 1)))
+#endif
```
wilfonba: The `#ifdef MFC_PRE_PROCESS` here and the `#ifdef MFC_SIMULATION` following this can be removed and `halo_size` can be used for both without any pre processor statements.

src/pre_process/m_checker.fpp
```diff
@@ -197,4 +198,51 @@ contains
 
     end subroutine s_check_inputs_misc
 
+    subroutine s_check_bc
+
+        integer :: i
+
+        @:PROHIBIT(num_bc_patches > num_bc_patches_max, "num_bc_patches must be <= 10")
```
wilfonba: Rewrite to remove the hardcoded number 10. There's a routine called `s_int_to_str` in `m_helper.fpp` that can be used for this. There should be an example somewhere else in the checker routines on how to do this.

src/pre_process/m_perturbation.fpp
```diff
@@ -14,9 +14,11 @@ module m_perturbation
 
     use m_eigen_solver          ! Subroutines to solve eigenvalue problem for
 
-    use m_boundary_conditions   ! Boundary conditions module
+    use m_boundary_common   ! Boundary conditions module
     ! complex general matrix
 
+    use m_helper
+
```
wilfonba: Is `m_helper` used in this module?

src/pre_process/m_perturbation.fpp
```diff
@@ -29,7 +31,7 @@ module m_perturbation
 
     real(wp), allocatable, dimension(:, :, :, :) :: q_prim_temp
 
-    real(wp) :: bcxb, bcxe, bcyb, bcye, bczb, bcze
+    ! real(wp) :: bcxb, bcxe, bcyb, bcye, bczb, bcze
```
wilfonba: Delete instead of comment

src/pre_process/m_start_up.fpp
```diff
@@ -770,12 +780,14 @@ contains
             pref = 1._wp
         end if
         call s_initialize_mpi_common_module()
+        ! call s_initialize_mpi_proxy_module()
```
wilfonba: Delete instead of comment if this isn't needed

src/pre_process/m_start_up.fpp
```diff
@@ -921,12 +933,14 @@ contains
 
         ! Deallocation procedures for the modules
         call s_finalize_mpi_common_module()
+        ! call s_finalize_mpi_proxy_module()
```
wilfonba: Delete instead of comment if this isn't needed

toolchain/mfc/run/case_dicts.py
```diff
@@ -98,9 +98,12 @@ def analytic(self):
     'surface_tension': ParamType.LOG,
     'elliptic_smoothing': ParamType.LOG,
     'elliptic_smoothing_iters': ParamType.INT,
+    'num_bc_patches': ParamType.INT,
+    'viscous': ParamType.LOG,
+    'bubbles_lagrange': ParamType.LOG,
 })
 
-for ib_id in range(1, 10+1):
+for ib_id in range(1, 25+1):
```
wilfonba: revert to 10

wilfonba (changes requested):

src/common/include/inline_boundary_conditions.fpp
ChrisZYJ: The elasticity part looks correct to me, but I'm not sure what line 15
`q_prim_vf(i)%sf(${DEST}$) = -q_prim_vf(i)%sf(${SRC}$)`
does. Maybe I'm missing something, but should it be
`q_prim_vf(momxb + ${DIR}$-1)%sf(${DEST}$) = -q_prim_vf(momxb + ${DIR}$-1)%sf(${SRC}$)`
instead?

wilfonba: Hmm, I don't know how the test suite passed with that. It's definitely wrong though. I fixed it.

src/common/include/inline_boundary_conditions.fpp
sbryngelson, line 1: this makes the code very hard to read and edit

wilfonba: I did this because I felt that it made the code much easier to read and edit...

wilfonba: This is an easy change to revert, but it would add back a significant amount of nearly duplicated code.

wilfonba: Without these macros everything would (probably) have to be inlined explicitly, adding a ton of nearly duplicate code. I was worried about calling the existing subroutines with additional arguments for the grid cell index because I wasn't sure the compiler would be able to inline everything properly and keep things fast.

sbryngelson: i think we should use functions and subroutines rather than just inlining arbitrary code (which one then has to fish out). at this point there is no reason not to (before we used these macros for speedups due to compiler non-optimizations).

sbryngelson: similarly, `inline_riemann` should be subroutines/functions

sbryngelson: with two-pass compilation we are fast on NVIDIA, and CCE is aggressive about inlining in the Frontier case.

sbryngelson: using macros this way just isn't sustainable. if we find some slowdown somewhere we should try to fix it a more 'fortran appropriate' way rather than continuing the workarounds.

wilfonba: Fair enough. Brian or I will revert this then.

okBrian: Yea I just finished a big exam yesterday, so I should have time to revert them.

ChrisZYJ: Sorry to be a busybody here, just adding that macros can cause gdb to miss breakpoints, making debugging a lot harder... Not sure if you guys have seen it too

sbryngelson: I have seen this. I'm fine with macros for macros sake, but really want to avoid using them as a copy/paste code solution. for example, in these macros we have things like
```fortran
do i = 1,10
	do j = 5,idx%end
		blah = blah(i) + thing(x)
	end do
end do
```

if someone changes the name of one of these variables anywhere, then the macro will break. if you change the macro, the code will break.

src/common/m_boundary_common.fpp
```diff
@@ -0,0 +1,1726 @@
...
+                                -q_prim_vf(momxb)%sf(j - 1, k, l)
+
+                            !$acc loop seq
+                            do i = momxb + 1, sys_size
+                                q_prim_vf(i)%sf(-j, k, l) = &
+                                    q_prim_vf(i)%sf(j - 1, k, l)
+                            end do
+
+                            if (elasticity) then
+                                do i = 1, shear_BC_flip_num
```
sbryngelson: missing acc statement

src/common/m_boundary_common.fpp
```diff
@@ -0,0 +1,1726 @@
...
+                                    q_prim_vf(shear_BC_flip_indices(1, i))%sf(-j, k, l) = &
+                                        -q_prim_vf(shear_BC_flip_indices(1, i))%sf(j - 1, k, l)
+                                end do
+                            end if
+
+                            if (hyperelasticity) then
+                                q_prim_vf(xibeg)%sf(-j, k, l) = &
```
sbryngelson: missing acc statement?

sbryngelson:
some suggestions:

<html><head></head><body><p><strong>Reviewing GitHub PR</strong></p>




### 1  API & data‑layout changes

| File                                                          | Change                                                                                                                 | Comments                                                                                                                                                                                                                                                                                                                                                                |
| ------------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
| **m\_boundary\_common.fpp**                                   | New optional code‑path for `bc_type == -17` (ghost‑cell extrapolation vs Dirichlet) gated by `#ifdef MFC_PRE_PROCESS`  | The *behavioural* difference between simulation and pre‑process is now hidden in two nested `#ifdef/elseif` blocks (≈ 6 places). Consider collapsing this behind a small pure function, e.g. `logical function is_ghost_patch(bc_code, build_phase)` to avoid *N* compile‑time branches and keep the call‐sites clean. ([[GitHub](https://github.com/MFlowCode/MFC/commit/ce31cdd.patch)][1])                                    |
| **m\_boundary\_common.fpp**                                   | Addition of macro templates `COLOR_FUNC_*` and heavy macro nesting                                                     | Gives nice brevity, but the code now relies on four layers of textual substitution (#\:def → C‑preprocessor → OpenACC & Fortran).  In practice this is hard to debug with optimisation + GPU offload.  A generic elemental/pure procedure with an assumed‑rank `c_div` argument would be easier to profile and keeps the OpenACC loop separate from the boundary logic. |
| **m\_derived\_types.fpp**                                     | Removed `vel`, `alpha_rho`, `alpha`, `pres` from the patch/plane derived‑type                                          | Double‑check every caller; several post‑processing kernels (e.g. statistics routines) still index these components on `master`.  If they are truly redundant, mark the commit message **breaking‐API** and bump the module major version. ([[GitHub](https://github.com/MFlowCode/MFC/commit/ce31cdd.patch)][1])                                                                                                                 |
| **m\_boundary\_common.fpp / s\_populate\_capillary\_buffers** | Function now receives `bc_type` and uses helper scalars `bcxb, bcxe, bcyb…`                                            | Nice step toward removing the long cascade of `elseif`s.  However the new scalars are implicitly imported globals—make them arguments or members of a boundary‑info derived‐type so that the routine stays *pure* and testable. ([[GitHub](https://github.com/MFlowCode/MFC/commit/ce31cdd.patch)][1])                                                                                                                           |

---

### 2  Maintainability & readability

1. **Replace magic numbers with an enumerator module**
   `-1 ↔ periodic`, `-2 ↔ slip`, `-16 ↔ no‑slip`, `-17 ↔ ghost‑patch` is now sprinkled in \~40 places.  Add an `enum, bind(c)` or at least `integer, parameter :: BC_PERIODIC = -1, …` to a central header.  It eliminates comment rot and helps the debugger show symbolic names.

2. **Reduce pre‑processor surface area**
   The project already uses OpenACC, MPI and metaprogramming macros.  Each new `#ifdef` multiplies test cases (SIMULATION × PRE\_PROCESS × GPU/CPU × DIM).  Where possible:

   ```fortran
   select case (phase)
   case (PHASE_PREPROCESS)
      call handle_ghost()
   case default
      call handle_dirichlet()
   end select
   ```

   This keeps the logic inside the language and tools like fpm/f18 can still inline and remove dead code.

4. **Vector‑friendly access**
   The new macro builds lines like

   ```fortran
   c_divs(i)%sf(-j,k,l) = -c_divs(i)%sf(j-1,k,l)
   ```

   At run‑time that is fine, but on GPUs the negative index forces an address‑check.  A simple local pointer such as

   ```fortran
   real(wp), pointer :: left(:,:,:)
   left => c_divs(i)%sf
   left(-j,k,l) = -left(j-1,k,l)
   ```

   lets the compiler expose aligned loads/stores.

---

### 3  Correctness & safety

* **Intent declarations** – `bc_type` is `intent(in)` in the new signature but is also dereferenced on the device.  If an external routine ever writes to it, race conditions appear because there is no `update device/host`.  Add `intent(in)` explicitly and compile with `-fcheck=all` once.
* **Ghost‑patch extrapolation order** – The macro `PRIM_GHOST_CELL_EXTRAPOLATION_BC` copies cell 0 into all *j* ghost cells; previously you used linear extrapolation from cell 1.  Validate high‑order capillary terms—the curvature operator is sensitive to constant padding.
* **Range checks** – Several loops now run from `-buff_size` to `m+buff_size` on arrays with base‑1 allocations.  Enable `-fcheck=bounds` in CI to avoid silent OOB writes.


---

### 5  Suggested follow‑ups

1. **Unit‑test the new patch type**: A 1‑D linear advection with `bc = -17` should conserve the step function within machine epsilon.
2. **Refactor boundary handling into a generically typed strategy object** so the simulation loop calls `call patch_handler%apply(q)`.  That will let you add future BCs (sponge, inflow) without re‑touching the hot loops.

---

PR #829, opened 2025-05-06
Fix chemistry‑related bugs

DimAdam-01:
## Description

1) Integrate the new Pyrometheus branch to calculate transport coefficients,
resolve the issue that prevented multi‑rank execution when using non‑hardcoded initial conditions;
and fix minor bugs in both the inline Riemann solver and the HLL Riemann solver.

### Type of change

-  [x]  Bug fix (non-breaking change which fixes an issue)

### Scope

## How Has This Been Tested?

- [x] Test A: 1D_reactive_shocktube
 Run the 1D shocktube reaction case, compared the result and they had the exact match, 

**Test Configuration**:

Run on Delta and Delta AI ( A100 and GH GPUs), as well as on my personal laptop
## Checklist


src/simulation/m_riemann_solvers.fpp
```diff
@@ -1015,8 +1015,8 @@ contains
                                     Y_L = qL_prim_rs${XYZ}$_vf(j, k, l, i)
                                     Y_R = qR_prim_rs${XYZ}$_vf(j + 1, k, l, i)
 
-                                    flux_rs${XYZ}$_vf(j, k, l, i) = (s_M*Y_R*rho_R*vel_R(dir_idx(norm_dir)) &
-                                                                     - s_P*Y_L*rho_L*vel_L(dir_idx(norm_dir)) &
+                                    flux_rs${XYZ}$_vf(j, k, l, i) = (s_M*Y_R*rho_R*vel_R(dir_idx(1)) &
+                                                                     - s_P*Y_L*rho_L*vel_L(dir_idx(1)) &
```
wilfonba, line 1019: Has this been verified in 2D/3D?

DimAdam-01: In 2D yes,

wilfonba: Sounds good

sbryngelson: it's unclear to me why this fixes the issue, can you explain @DimAdam-01

sbryngelson:
has the code that automatically turns a 1D case into an extrapolated 2D one been put into MFC? @DimAdam-01

src/simulation/m_riemann_solvers.fpp
DimAdam-01: Previously, we indexed the velocity component with dir_idx(norm_dir). That worked for the x‑direction—norm_dir was 1, and dir_idx(1) returned 1, so we correctly used U(1) for the flux. However, when scanning in the y‑direction (norm_dir = 2), our permutation (the dir_idx array changes depending on the direction) [2, 1, 3] makes dir_idx(2) return 1 instead of 2. As a result, the code still picks U(1) rather than the intended U(2), and the flux is computed with the wrong velocity component., @sbryngelson

DimAdam-01:
I consider it a distinct feature that I intended to add separately, but I can include it in this pull request. @sbryngelson

src/simulation/m_riemann_solvers.fpp
sbryngelson: so should this fix also work in 3D?

DimAdam-01: I’m confident it will work, I don’t see any reason it would fail.

toolchain/pyproject.toml
```diff
@@ -38,8 +38,8 @@ dependencies = [
     "matplotlib",
 
     # Chemistry
-    "cantera==3.0.1",
-    "pyrometheus==1.0.2"
+    "cantera==3.1.0",
+    "pyrometheus@git+https://github.com/DimAdam-01/pyrometheus.git@main"
```
sbryngelson: you can't make master MFC use your fork of pyrometheus....

DimAdam-01: The original version encountered build errors on Frontier, so I introduced a few modifications to verify whether it would compile successfully there, and it did.

sbryngelson: then we need to create a PR for the Pyrometheus fix first

DimAdam-01: Yes, I’m speaking with Esteban at the moment.

toolchain/pyproject.toml
```diff
@@ -38,8 +38,8 @@ dependencies = [
     "matplotlib",
 
     # Chemistry
-    "cantera==3.0.1",
-    "pyrometheus==1.0.2"
+    "cantera==3.1.0",
+    "pyrometheus@git+https://github.com/pyrometheus/pyrometheus.git@main"
```
henryleberre: @DimAdam-01 I think it would be preferable to reference an official (versioned) release of Pyro instead of referencing the main branch. You can see the original code referenced the 1.0.2 release on Pypi

sbryngelson: @DimAdam-01 i just created a new pyro release 1.0.3 (not sure if it's on pypi just yet)

DimAdam-01: I just tested it, it's not on pypi yet: (ERROR: No matching distribution found for pyrometheus==1.0.3)

sbryngelson: yeah i created it 3 min ago

sbryngelson: you can change to `pyrometheus==1.0.3` now

sbryngelson (approved):

PR #831, opened 2025-05-08
MPI Refactor

wilfonba:
## Description

This PR refactors a lot of MPI code to reduce duplicated code and shorten the MPI-related code in the codebase. Significant testing is needed to verify the changes' correctness, but I'm opening this as a draft now so that people know what's being changed and can start reviews/make suggestions early.

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [ ] 3D IBM one rank vs. multi-rank comparison (simple geometry)
- [ ] 2D IBM one rank vs. multi-rank comparison (simple geometry)
- [ ] 3D IBM one rank vs. multi-rank comparison (complex geometry)
- [ ] 2D IBM one rank vs. multi-rank comparison (complex geometry)
- [ ] 3D surface tension one rank vs. multi-rank comparison
- [ ] 2D surface tension one rank vs. multi-rank comparison
- [ ] 3D QBMM one rank vs. multi-rank comparison
- [ ] 2D QBMM one rank vs. multi-rank comparison
- [ ] 1D QBMM one rank vs. multi-rank comparison
- [ ] EL Bubbles simulation and post-process verification
- [ ] Verify that the existing 2MPI rank golden files are correct

## Checklist


src/post_process/m_global_parameters.fpp
```diff
@@ -474,7 +474,7 @@ contains
         integer :: i, j, fac
 
         ! Setting m_root equal to m in the case of a 1D serial simulation
-        if (num_procs > 1 .and. n == 0) m_root = m_glb
+        if (n == 0) m_root = m_glb
```
sbryngelson, line 476: There must be some reason why that if statement was in there.... do you know what it was?

wilfonba: There was a line in the `s_decompose_computational_domain.fpp` in `src/post_process/m_mpi_proxy.fpp` that set `m_root = m` _before_ `m` was converted from a global cell count to a local cell count https://github.com/MFlowCode/MFC/blob/40164cf2e1a13d5f77a42ee50d39d7d2bdb24f2e/src/post_process/m_mpi_proxy.fpp#L573, so `m_root` was always being set to `m_glb`. In 1D, post processing is done by doing an MPI reduce to the root processor and saving only from the root. The `num_procs > 1` condition is removed because this assignment no longer happens in the domain decomposition code, but `m_root` should always be equal to `m_glb` in the 1D case.

sbryngelson:
I'm going to do a code review, but I think I only have two lingering questions for verification:

* Can you confirm that RDMA MPI works on Frontier and an NVIDIA machine? You don't need to test a bunch of cases; just one is fine.
* Can you confirm that the output files, in particular the serial output files, are the same as they were before the PR? I noticed you modified those files. Can't remember if you modified the probe output, but if so then confirming that as well would be useful.

Mostly asking because the above two things aren't covered in CI (though probably should be).

wilfonba:
The scaling plots I included show RDMA MPI on Frontier, so that's covered. I'm not sure which NVIDIA machines we have that support RDMA MPI. I'll try Phoenix since I know that's what Max has been using it on. I'll have to go back and look at my changes to serial output files to remember what they even were, but I can check on that as well.

wilfonba:
@sbryngelson I've updated the scaling plots and added RDMA scaling on Phoenix. The I/O files have actually been tested already. Serial I/O is covered in the test suite, and parallel I/O is covered in all of the example simulations I gave in the initial comment.

PR #836, opened 2025-05-14
Remove redundant timestep code

danieljvickers:
## Description

Fixes #(issue) [optional]

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

## Checklist

### DIFF
--- a/src/simulation/m_time_steppers.fpp
+++ b/src/simulation/m_time_steppers.fpp
@@ -357,6 +357,52 @@ contains
 
     end subroutine s_initialize_time_steppers_module
 
+    subroutine s_evolve_q_pb_mv(index1, index2, scaler1, scaler2, scaler3, scaler4)  !! TODO :: Get a better name for this
+
+        integer, intent(in) :: index1, index2  !! TODO :: I have no idea what index is meant to represent. Rename this.
+        real(wp), intent(in) :: scaler1, scaler2, scaler3, scaler4
+        integer :: i, j, k, l, q
+
+        !$acc parallel loop collapse(4) gang vector default(present)
+        do i = 1, sys_size
+            do l = 0, p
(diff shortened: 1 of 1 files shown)


wilfonba:
Instead of `scaler1,...,scalarN` could you use an array to store the coefficients and simply pass the correct row/column of that array?

src/simulation/m_time_steppers.fpp
```diff
@@ -339,6 +339,65 @@ contains
 
     end subroutine s_initialize_time_steppers_module
 
+    subroutine s_evolve_q_pb_mv(index, scaler1, scaler2, scaler3, scaler4)  !! TODO :: Get a better name for this
+
+        integer, intent(in) :: index  !! TODO :: Rename this
```
sbryngelson, line 344: make intents `inout` for testing purposes

src/simulation/m_time_steppers.fpp
```diff
@@ -339,6 +339,65 @@ contains
...
+        real(wp), intent(in) :: scaler1, scaler2, scaler3, scaler4  !! TODO :: Rename these too
```
sbryngelson, line 345: same here

sbryngelson:
@danieljvickers, this is getting a little silly/comedic. Why don't you move on to a different issue that you feel you're suited for (send me a message of what you think it is, perhaps), and I'll think about this one a bit. This is really an edge case that probably needs a little more thought from our previous snafus with different types of runners (though this one seems especially _special_ AKA stupid)...

You're in a good place to create some containers for MFC. We could try creating those in a few different formats.

sbryngelson:
stale

PR #837, opened 2025-05-15
Improve mpi buffer allocation in post_process

hyeoksu-lee:
## Description

In the subroutine `s_initialize_mpi_proxy_module` in post_process, `q_cons_buffer_in` and `q_cons_buffer_out` are allocated with the size `buff_size*sys_size*(m + 2*buff_size + 1)*(n + 2*buff_size + 1)*(p + 2*buff_size + 1)/(min(m, n, p)+ 2*buff_size + 1)`. However, as the Fortran computes the operations sequentially, it could be too large before dividing by `(min(m, n, p)+ 2*buff_size + 1)`.

For example, I noticed in a 3D simulation (1023 x 1023 x 511) with method of classes bubbles (`nb = 11`) that `buff_size*sys_size*(m + 2*buff_size + 1)*(n + 2*buff_size + 1)*(p + 2*buff_size + 1)` is 6.4242e+10, which is greater than the max value of integer 2^31 - 1 =2.1475e+09. Therefore, it becomes a wrong negative integer value, -2528337, and thus it throws an error. 

Therefore, I made a quick fix to this by changing the order of operations.

### Type of change

### Scope

## How Has This Been Tested?

- [x] Large 3D simulation (1023 x 1023 x 511, sys_size = 29)

* What computers and compilers did you use to test this: Carpenter

## Checklist

### DIFF
--- a/src/post_process/m_mpi_proxy.fpp
+++ b/src/post_process/m_mpi_proxy.fpp
@@ -67,18 +67,18 @@ contains
 
                     allocate (q_cons_buffer_in(0:buff_size* &
                                                sys_size* &
-                                               (m + 2*buff_size + 1)* &
-                                               (n + 2*buff_size + 1)* &
-                                               (p + 2*buff_size + 1)/ &
+                                               (m + 2*buff_size + 1)/ &
                                                (min(m, n, p) &
-                                                + 2*buff_size + 1) - 1))
+                                                + 2*buff_size + 1)* &
+                                               (n + 2*buff_size + 1)* &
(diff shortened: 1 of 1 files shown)


sbryngelson:
Good find. There's actually a "better" way of doing this by making them floating point additions/multiplies then turning the result into the nearest integer. @wilfonba  can you incorporate this fix into your MPI PR?

hyeoksu-lee:
@sbryngelson @wilfonba Sounds great. I will close this PR then. Thanks!

PR #842, opened 2025-05-21
Frontier module update

wilfonba:
## Description

Updates Frontier modules for CCE 19.0.X and the profiling updates.

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

Ran
`./mfc.sh run examples/3D_performance_test/case.py -c frontier -n 1`
`./mfc.sh run examples/3D_performance_test/case.py -c frontier -# PERF_Test -n 1 --rcu`
`./mfc.sh run examples/3D_performance_test/case.py -c frontier --rsys`
and all worked as expected.


toolchain/templates/frontier.mako
```diff
@@ -44,7 +44,11 @@ export MPICH_GPU_SUPPORT_ENABLED=1
     ${helpers.run_prologue(target)}
```
sbryngelson, line 48: can you change the `export MPICH_GPU_SUPPORT_ENABLED=1` to be 0 if it is not a GPU simulation?

wilfonba: Yep. Testing now and then I'll push.

sbryngelson:
@wilfonba did you test that the CPU case works?

If so, can you add it as a CI run? So each CI test run will run the CPU version and the GPU version.

I'm realizing I'm ballooning your PR here, apologies. I've been wanting to do this for a while, actually.

wilfonba:
@sbryngelson Yeah, I can. No worries, shouldn't take long at all

wilfonba:
@sbryngelson I'll send a message when this is ready to merge. I was looking at the Frontier logs and noticed that both the CPU and GPU runners ran the CPU code due to an error I made in the workflow files, so I did what I think corrected that, but I'll have to wait for the runners.

sbryngelson:
this appears to be working, ?

sbryngelson (approved):

PR #844, opened 2025-05-26
Extend initial conditions from 1D to 2D and 2D to 3D 

DimAdam-01:
## Description

his pull request adds the ability to generate initial conditions by mapping a 1D simulation’s results onto 2D and 2D into 3D grids. It also introduces support for importing 1D profiles (for example, a premixed‐flame solution from Cantera) and using them to initialize and run a 1D simulation.
Fixes #(issue) [optional]

### Type of change

### Scope

## How Has This Been Tested?

- [x] Test A: 1D → 2D Mapping (Uniform Domain)

1. Use a 2D domain whose x-length exactly matches the 1D simulation.

2. Distribute the grid across multiple MPI ranks.

3. Verify that the reconstructed 2D initial field matches the original 1D data.

-[x] Test 2: 1D → 2D Mapping (Partial Domain + Patch)

1. Extend only a portion of the 2D domain with the 1D profile, fill the remainder via a prescribed patch in case.py.

2. Run across multiple ranks with varying grid decompositions (different rank counts and grid-size ratios).

3. Test different horizontal offsets (e.g.\ placing the 1D profile at the left edge, centered, etc.) and confirm correct placement and values.

The same two tests were then carried over to 2D → 3D extensions, exercising uniform-domain mapping and partial-domain + patch scenarios in three dimensions. 

**Test Configuration**:

My Laptop and Delta (A100)
## Checklist


src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,5 +1,33 @@
...
+    character(len=100), dimension(nFiles) :: fileNames
+    ! Arrays to store all data from files
+    real(wp), dimension(nRows, nFiles) :: stored_values
+    real(wp), dimension(nRows) :: x_coords
+    logical :: files_loaded = .false.
+    real(wp) :: domain_start, domain_end
+    character(len=*), parameter :: init_dir = "/home/YourDirectory"
+    character(len=20) :: file_num_str     ! For storing the file number as a string
+    character(len=20) :: zeros_part       ! For the trailing zeros part
+    character(len=6), parameter :: zeros_default = "000000"  ! Default zeros (can be changed)
```
sbryngelson: use doxygen formatting...

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -8,6 +36,42 @@
...
+                if (ios /= 0) then
+                    print *, "Error opening file: ", trim(fileNames(f))
+                    cycle  ! Skip this file on error
+                end if
+                ! Read all rows at once into memory
+                do iter = 1, nRows
+                    read (unit, *, iostat=ios) x_coords(iter), stored_values(iter, f)
+                    if (ios /= 0) then
+                        print *, "Error reading file ", trim(fileNames(f)), " at row ", iter
+                        exit  ! Exit loop on error
```
sbryngelson: don't use print like this, it will print 1000 lines of same line if you use 1000 MPI ranks. 

don't use exit or stop. use s_mpi_abort('Reason for exit')

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -8,6 +36,42 @@
...
+                close (unit)
+            end do
+            ! Calculate domain information for mapping
+            domain_start = x_coords(1)
+            domain_end = x_coords(nRows)
+            x_step = x_cc(1) - x_cc(0)
+
+            delta = x_cc(0) - domain_start - x_step/2.0
+            global_offset = nint(abs(delta)/x_step)
+            print *, "All data files loaded. Domain range:", domain_start, "to", domain_end
```
sbryngelson: see above

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -5,6 +5,34 @@
     real(wp) :: rhoH, rhoL, pRef, pInt, h, lam, wl, amp, intH, intL, alph
     real(wp) :: factor
 
+    integer, parameter :: nFiles = 14   ! Can be changed to any number
```
sbryngelson: what does this number mean? i don't know how this works

sbryngelson: this is why you add documentation and better code comments :D

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -5,6 +5,34 @@
...
+    character(len=20) :: zeros_part       ! For the trailing zeros part
+    character(len=6), parameter :: zeros_default = "000000"  ! Default zeros (can be changed)
+    ! Generate file names in a loop
+    do f = 1, nFiles
+        ! Convert file number to string with proper formatting
+        if (f < 10) then
+            write (file_num_str, '(I1)') f  ! Single digit
+        else
+            write (file_num_str, '(I2)') f  ! Double digit
+            ! For more than 99 files, you might need to adjust this format
```
sbryngelson: what are these files you speak of

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -130,6 +158,50 @@
...
 
+    case (207)
+        if (.not. files_loaded) then
+            ! Print status message
+            index_1 = i
+            do f = 1, nFiles
+                ! Open the file for reading
+                open (newunit=unit, file=trim(fileNames(f)), status='old', action='read', iostat=ios)
+                if (ios /= 0) then
+                    print *, "Error opening file: ", trim(fileNames(f))
```
sbryngelson: is this running in parallel if `mpirun -np 10 ...` ? if so, don't use print like this

src/pre_process/include/3dHardcodedIC.fpp
```diff
@@ -56,6 +90,89 @@
...
+        do f = 1, nFiles
+            if (f > 3) then
+                jump = 1
+            else
+                jump = 0
+            end if
+            q_prim_vf(f + jump)%sf(i, j, k) = stored_values(data_i, data_j, f)
+        end do
+        ! Set z velocity to zero
+        q_prim_vf(4)%sf(i, j, k) = 0.0_wp
```
sbryngelson: don't use explicit numbers as indices in `q_prim_vf(X)`...

src/pre_process/include/3dHardcodedIC.fpp
```diff
@@ -56,6 +90,89 @@
...
+            print *, "All ", nFiles, " files loaded successfully!"
+
+            files_loaded = .true.
+        end if
+
+        data_i = i + 1 + global_offset_x - index_x  ! x-direction index in data grid
+        data_j = j + 1 + global_offset_y - index_y  ! y-direction index in data grid
+
+        do f = 1, nFiles
+            if (f > 3) then
```
sbryngelson: what happens at 3? magic number

src/pre_process/include/1dHardcodedIC.fpp
sbryngelson, line 1: I'm not sure if this works with other hardcoded ICs? ask @wilfonba

src/pre_process/include/2dHardcodedIC.fpp
DimAdam-01: They are produced if we set paraller io = F in the case file, the files names are for example prim.1.00.00000.dat

src/pre_process/include/2dHardcodedIC.fpp
DimAdam-01: It runs fine if I use 10 ranks

sbryngelson: that's not the point

src/pre_process/include/2dHardcodedIC.fpp
sbryngelson: what if the user uses parallel_IO=T? also you only get prim.1.00.00000.dat if you have prim_var_wrt true.

DimAdam-01: This is noted in the 1D_hardcoded description: the files are generated when parallel I/O is disabled in case.py, I will add it in the other hardcoded.fpp files.

src/pre_process/include/2dHardcodedIC.fpp
DimAdam-01: I removed the print statement, but it was being printed by every MPI rank. Should I add it back and wrap and add a condition like if (MPI_rank == 0)?

sbryngelson: yes, you will see similar patterns in the other parts of the code

src/pre_process/include/2dHardcodedIC.fpp
sbryngelson: does this now apply to all hardcoded patches, then? because obviously some of them we don't want to do this to...

examples/2D_Detonation/case.py
sbryngelson, line 1: how does this case work? can i just run it and it will run the 1d case and then restart in 2D? if not, then we need two case files and an explanation (readme?) describing how to use them. it also looks like you are disabling chemistry and enabling scaling?

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,5 +1,53 @@
 #:def Hardcoded1DVariables()
+
+!> @file 1DHardcodedIC.fpp
```
sbryngelson: confused on how exactly this works... it looks like it will do this for every hardcoded 1d case? we don't want that

sbryngelson: and obviously not every hardcoded case is supposed to work like this (with chemistry and so on). maybe i'm just misunderstanding how this file works.

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,5 +1,53 @@
...
+    real(wp) :: x_len, x_step
+    integer :: global_offset            ! MPI subdomain offset
+    real(wp) :: delta
+    character(len=100), dimension(nFiles) :: fileNames ! Arrays to store all data from files
+    character(len=200) :: errmsg
+    real(wp), dimension(nRows, nFiles) :: stored_values  ! Imported Data
+    real(wp), dimension(nRows) :: x_coords
+    logical :: files_loaded = .false.
+    real(wp) :: domain_start, domain_end
+    character(len=*), parameter :: init_dir = "/home/YourDirectory" ! For example /home/MFC/examples/1D_Shock/D/
```
sbryngelson: if we had hardcoded cases in the test suite, this wouldn't work, right? and we might want that in the future

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -1,10 +1,49 @@
...
+    integer :: global_offset  ! MPI subdomain offset
+    real(wp) :: delta_x
+    character(len=100), dimension(nFiles) :: fileNames
+    character(len=200) :: errmsg
+    ! Arrays to store all data from files
+    real(wp), dimension(nRows, nFiles) :: stored_values  ! Imported Data
+    real(wp), dimension(nRows) :: x_coords
+    logical :: files_loaded = .false.
+    real(wp) :: domain_start, domain_end
+    character(len=*), parameter :: init_dir = "/home/pain/MFC-Adam/examples/1D_reactive_shocktube/D/"
```
sbryngelson: you can't leave it like this i don't think... also since it's a parameter the user can't change it without recompiling, right?

src/pre_process/include/3dHardcodedIC.fpp
```diff
@@ -1,10 +1,56 @@
...
+    !> @param yRows          Number of grid points in the y-direction
+    !> @param init_dir       Directory containing the `prim.*.dat` files
 
+    ! Place any declaration of intermediate variables here
     real(wp) :: rhoH, rhoL, pRef, pInt, h, lam, wl, amp, intH, alph
 
     real(wp) :: eps
 
+    integer, parameter :: nFiles = 15   ! Number of files (variables)
+    integer, parameter :: xRows = 401   ! Number of points in x
```
sbryngelson: can't you make this part of the case dictionary? it looks like everytime someone wants to do this they have to modify the source code

sbryngelson: maybe @wilfonba has an opinion on these hard coded patches because I don't use them much and am not super familiar with how one _usually_ uses them

src/pre_process/include/2dHardcodedIC.fpp
DimAdam-01: I moved the reading of the files to the specific case (for example case(270) )

src/pre_process/include/2dHardcodedIC.fpp
DimAdam-01: I changed it, so it uses mpi_abort.

examples/2D_Detonation/case.py
DimAdam-01: You need to first run the 1D case, for example, the 1D_reactive_shocktube, and then run the 2D case (2D_Detonation).

src/pre_process/include/1dHardcodedIC.fpp
DimAdam-01: Currently with the last changes, it is only necessary to specify the init_dir and the number of rows (nRows). These parameters could be moved to the individual case file for better modularity. While specifying nRows is not strictly required, omitting it will result in significantly higher memory usage, since the parameter m (the actual grid size) can be used instead of nRows to determine the required allocation.

src/pre_process/include/1dHardcodedIC.fpp
DimAdam-01: I am pretty confident that it wouldn't cause an error, I will talk with @wilfonba and try to test it.

src/pre_process/include/3dHardcodedIC.fpp
DimAdam-01: It was more practical for me to define the parameters directly in the hardcoded file, as most of the modifications in the cases I implemented, such as adding perturbations were concentrated there, rather than in case.py.

examples/2D_Detonation/case.py
sbryngelson: that's not obvious at all

src/pre_process/include/1dHardcodedIC.fpp
sbryngelson: im not sure i understand what you mean

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,13 +1,92 @@
...
+!>     `prim.<variable>.00.<timestep>.dat`
+!>     (produced when parallel I/O is disabled in `case.py`).
+!>   - Configurable parameters:
+!>       - @c nRows: total number of grid points in the 1D profile.
+!>   - Default file directory: `examples/Case_File/D`
+!>
+!> @param nRows Total number of grid points in the imported 1D profile.
+
     ! Place any declaration of intermediate variables here
+    integer, parameter :: nRows = 512   ! Number of grid points
```
sbryngelson: per MPI rank or global? what if you run with multiple ranks for the 1D case?

docs/documentation/case.md
```diff
@@ -236,6 +236,16 @@ and use `patch_icpp(i)%%geometry = 7` and `patch_icpp(i)%%hcid = 200` in the inp
...
 
+Also Several custom hard-coded patches are already defined to support common workflows. These include:
+
+case(100) in 1dHardcodedIC.fpp: used to initialize the domain with data from a previous 1D simulation or external source (e.g. Cantera), allowing reuse of existing 1D profiles.
+
+case(270) in 2dHardcodedIC.fpp: used to create 2D fields by extruding a 1D profile along the second spatial dimensions, This allows the solution of a 1D case at a
+specific timestep to be extended uniformly in the transverse direction and used as a full 2D initial condition.
+
+case(370) in 3dHardcodedIC.fpp: used to build 3D domains by extruding 2D data along the third dimension.
+
```
sbryngelson: Are these the only ones provided?

src/pre_process/include/1dHardcodedIC.fpp
DimAdam-01: I mentioned in the Doxygen comments that it represents the total number of grid points from the 1D file, but I forgot to include this description when declaring the variable.

docs/documentation/case.md
DimAdam-01: Yes, for three different dimensions. I usually add perturbations directly in the hardcoded case file rather than in the  case.py file.

sbryngelson: I'm confused why these are the only hardcoded ICs in the entire documentation or code. It seems like we should have many (and several examples). What's going on here? @wilfonba should know - or you should figure out from him

src/pre_process/include/1dHardcodedIC.fpp
sbryngelson: you can't read that info from the file itself?

src/pre_process/include/3dHardcodedIC.fpp
sbryngelson: it seems like these parameters can be inferred from the input file. right now you have too many things you're hoping the user to do themself.

DimAdam-01: I suppose I could allocate memory, read the file to determine the number of grid points, and then resize the memory accordingly. The only thing the user would need to specify is the directory containing the files, along with the filename pattern (e.g., prim.xx.yy.zzzzz.dat).

sbryngelson: can't you declare the variable as allocatable and then allocate it once you know how big the file is?

docs/documentation/case.md
```diff
@@ -236,6 +236,11 @@ and use `patch_icpp(i)%%geometry = 7` and `patch_icpp(i)%%hcid = 200` in the inp
 Additional variables can be declared in `Hardcoded1[2,3]DVariables` and used in `hardcoded1[2,3]D`.
 As a convention, any hard coded patches that are part of the MFC master branch should be identified as 1[2,3]xx where the first digit indicates the number of dimensions.
 
+Also Several custom hard-coded patches are already defined to support common workflows. These include:
+
+By convention, patch ID `case(170)` load a 1D profile, ID `case(270)` extrude that 1D data into 2D, and ID `case(370)` extrude 2D data into 3D. 
+You only need to supply `init_dir` (and the filename‐pattern via `zeros_default`). All related variables (including init_dir, zeros_default, and the “files_loaded” flag)
```
sbryngelson: it should be `init_dir` , etc.

src/pre_process/include/1dHardcodedIC.fpp
sbryngelson: still a question

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,13 +1,75 @@
...
-
 #:enddef
 
 #:def Hardcoded1D()
 
     select case (patch_icpp(patch_id)%hcid)
-    case (100)
-        ! Put your variable assignments here
+    case (170)
+
```
sbryngelson: put a comment about what this case does here?

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -158,6 +157,77 @@
...
+            do
+                read (unit2, *, iostat=ios2) dummy_x, dummy_y
+                if (ios2 /= 0) exit ! Exit since files has been read
+                line_count = line_count + 1
+            end do
+            close (unit2)
+
+            index_x = i
+            xRows = line_count
+            allocate (x_coords(xRows))
```
sbryngelson: @:ALLOCATE?

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -1,13 +1,75 @@
...
+            end if
+            do
+                read (unit2, *, iostat=ios2) dummy_x, dummy_y
+                if (ios2 /= 0) exit  ! Exit since the file has been read
+                line_count = line_count + 1
+            end do
+            close (unit2)
+
+            xRows = line_count
+            allocate (x_coords(xRows))
```
sbryngelson: @:ALLOCATE ...

sbryngelson (commented):
repeated code in 1d/2d/3d hardcodedic.fpp files. add a macro that puts these somewhere else?

src/pre_process/include/ExtrusionHardcodedIC.fpp
```diff
@@ -0,0 +1,45 @@
...
+    real(wp) :: domain_xstart, domain_xend, domain_ystart, domain_yend
+    character(len=*), parameter :: init_dir = "/home/YourDirectory" ! For example /home/MFC/examples/1D_Shock/D/
+    character(len=20) :: file_num_str     ! For storing the file number as a string
+    character(len=20) :: zeros_part       ! For the trailing zeros part
+    character(len=6), parameter :: zeros_default = "00000"  ! Default zeros (can be changed)
+#:enddef
+
+#:def HardcodedDellacation()
+    if (allocated(stored_values)) then
+        deallocate (stored_values)
```
sbryngelson: @:DEALLOCATE

docs/documentation/case.md
```diff
@@ -236,6 +236,11 @@ and use `patch_icpp(i)%%geometry = 7` and `patch_icpp(i)%%hcid = 200` in the inp
 Additional variables can be declared in `Hardcoded1[2,3]DVariables` and used in `hardcoded1[2,3]D`.
 As a convention, any hard coded patches that are part of the MFC master branch should be identified as 1[2,3]xx where the first digit indicates the number of dimensions.
 
+Also Several custom hard-coded patches are already defined to support common workflows. These include:
```
sbryngelson: what's going on here. These include: 

but then there are no lines below it with a list....

sbryngelson: also capitalization error

Issue #828, opened 2025-05-05
Updated storage for equation indices and boundary condition types


wilfonba:
Implement the suggetion made here (https://github.com/MFlowCode/MFC/pull/821#discussion_r2052708379) for equation indices and the suggestion made here (https://github.com/MFlowCode/MFC/pull/819#issuecomment-2847058417) for `bcx[y,z]b[e]`.

Malmahrouqi3:
will look into this

PR #860, opened 2025-06-05
Equation Indices and BC Types - Issue (#828)

Malmahrouqi3:
## Description

making derived types and addressing issue #828 

Fixes #(issue) [optional]

### Type of change

### Scope

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

## Checklist


.vscode/settings.json
```diff
@@ -16,8 +16,8 @@
     "python.defaultInterpreterPath": "${workspaceFolder}/build/venv/bin/python3",
 
     "files.associations": {
-        "*.f90": "FortranFreeForm",
-        "*.fpp": "FortranFreeForm",
+        "*.f90": "fortran-modern",
```
sbryngelson: does this help anything?

src/common/m_boundary_common.fpp
```diff
@@ -22,7 +22,7 @@ module m_boundary_common
     type(scalar_field), dimension(:, :), allocatable :: bc_buffers
     !$acc declare create(bc_buffers)
 
-    real(wp) :: bcxb, bcxe, bcyb, bcye, bczb, bcze
+    type(boundary_flags) :: bc_flag
```
sbryngelson: These aren't flags but rather bounds. bc_x_b = boundarycondition_xdirection_beginning.

.vscode/settings.json
Malmahrouqi3: missed it

src/common/m_derived_types.fpp
```diff
@@ -345,6 +345,11 @@ module m_derived_types
         type(vec3_dt), allocatable, dimension(:) :: var
     end type mpi_io_airfoil_ib_var
 
+    !> Derived type for boundary flags
+    type boundary_flags
```
sbryngelson: see above comment about the word 'flags'

.vscode/settings.json
Malmahrouqi3: not really

src/simulation/m_ibm.fpp
```diff
@@ -170,46 +192,39 @@ contains
...
+                q_prim_vf(c_idx)%sf(j, k, l) = gp%ip%c
             end if
 
             if (model_eqns /= 4) then
                 ! If in simulation, use acc mixture subroutines
                 if (elasticity) then
-                    call s_convert_species_to_mixture_variables_acc(rho, gamma, pi_inf, qv_K, alpha_IP, &
-                                                                    alpha_rho_IP, Re_K, j, k, l, G_K, Gs)
+                    call s_convert_species_to_mixture_variables_acc(rho, gamma, pi_inf, qv_K, &
+                                                                    gp%ip%alpha, gp%ip%alpha_rho, Re_K, j, k, l, G_K, Gs)
```
sbryngelson: one advantage of this strategy of using more derived types is that we can just pass in `gp%ip` (for ibm) or `state%L` (or something like this) for riemann solvers. basically gathering all the data that could exist at a given point and lumping it into a derived type, and then the argument list becomes very short. for example, you just pass in 

`call s_routine(state)`

where
```fortran
subroutine s_routine(mystate)
	type(state) :: mystate
	
	[...]
	
	mystate%rho = [......]
end subroutine
```

src/simulation/m_ibm.fpp
```diff
@@ -743,22 +757,15 @@ contains
 
     !> Function that uses the interpolation coefficients and the current state
     !! at the cell centers in order to estimate the state at the image point
-    pure subroutine s_interpolate_image_point(q_prim_vf, gp, alpha_rho_IP, alpha_IP, pres_IP, vel_IP, c_IP, r_IP, v_IP, pb_IP, mv_IP, nmom_IP, pb, mv, presb_IP, massv_IP)
+    pure subroutine s_interpolate_image_point(q_prim_vf, gp, pb, mv)
```
sbryngelson, line 760: this is an good example of what i was talking about above... for removing so many arguments.

Malmahrouqi3: yup yup I see that

src/simulation/m_ibm.fpp
```diff
@@ -865,10 +872,27 @@ contains
 
     !> Subroutine to deallocate memory reserved for the IBM module
     impure subroutine s_finalize_ibm_module()
+        integer :: i
+
+        if (allocated(ghost_points)) then
+            do i = 1, size(ghost_points)
+                if (allocated(ghost_points(i)%ip%alpha_rho)) deallocate (ghost_points(i)%ip%alpha_rho)
```
sbryngelson: FYI from perplexity:

Perfect! Given that your `point_data` type uses **allocatable components** (not pointers), your manual deallocation loop is completely unnecessary. 

## The Clean Solution

Since all the components you're manually deallocating (`alpha_rho`, `alpha`, `r`, `v`, `pb`, `mv`, `nmom`, `presb`, `massv`) are declared as `allocatable`, Fortran will **automatically deallocate them** when the parent object is deallocated.

You can replace your entire verbose loop with just:

```fortran
if (allocated(ghost_points)) deallocate(ghost_points)
```

That's it! When `ghost_points` is deallocated, Fortran automatically:
1. Deallocates all allocatable components within each `ghost_points(i)%ip`
2. Deallocates the `ghost_points` array itself

## Why This Works

The Fortran standard guarantees that when a derived type object with allocatable components is deallocated (or goes out of scope), all of its allocatable components are automatically deallocated first. This is called **automatic deallocation** and is one of the key advantages of using `allocatable` over `pointer`.

## Additional Benefits

This approach is not only cleaner but also:
- **Safer**: No risk of forgetting to deallocate a component
- **More maintainable**: Adding new allocatable components to `point_data` requires no changes to deallocation code
- **Better performance**: No unnecessary allocation status checks in loops

Your original manual approach was doing work that Fortran handles automatically for allocatable components. The verbose loop is only necessary when dealing with `pointer` components, which require explicit deallocation.

Malmahrouqi3: neat will remove the loop

src/simulation/m_ibm.fpp
```diff
@@ -95,6 +95,35 @@ contains
         @:ALLOCATE(ghost_points(1:num_gps))
```
sbryngelson, line 95: i'm not sure how this ever worked. i would ask on the MFC Slack and tag @abbotts . this is an array of derived types. it seems like you would need to use something like this macro for AMD GPUs: https://github.com/MFlowCode/MFC/blob/32c034746fadfd2797ff00ce60ae287b349a8090/src/common/include/macros.fpp#L48-L62

sbryngelson: the reason is perhaps that `sf` is another derived type. when you have a derived type that has another allocatable derived type within it, you need to use a macro like this (except instead of `sf` it would be something else)

src/simulation/m_ibm.fpp
```diff
@@ -95,6 +95,35 @@ contains
...
+                allocate (ghost_points(i)%ip%r(nb))
+                allocate (ghost_points(i)%ip%v(nb))
+                if (.not. polytropic) then
+                    allocate (ghost_points(i)%ip%pb(nb))
+                    allocate (ghost_points(i)%ip%mv(nb))
+                end if
+            end if
+
+            if (qbmm) then
+                allocate (ghost_points(i)%ip%nmom(nb*nmom))
```
sbryngelson: we use the `@:ALLOCATE` macro in almost all cases i think, since it declares the variable on the device.

sbryngelson (commented):
this is a good PR but isn't quite right. the issues are subtle though, and are somewhat MFC specific. ask about the things i brought up on the slack channel (or ask @wilfonba @abbotts @anandrdbz etc.)

sbryngelson:
This CCE failure suggests something is wrong with allocation, deallocation, or memory access. It happens on Cray CPU cases.

```
+ srun --nodes 1 --ntasks-per-node 1 --cpus-per-task 7 /lustre/orion/cfd154/scratch/sbryngelson/runner/actions-runner/_work/MFC/MFC/build/install/877bac7bd7/bin/simulation
corrupted size vs. prev_size in fastbins
srun: error: frontier10353: task 0: Aborted (core dumped)
srun: Terminating StepId=3499970.414
```

and chatgpt says 

Thanks — since your simulation is written in **Fortran**, the `corrupted size vs. prev_size in fastbins` error still points to **heap corruption**, but it's likely caused by one of the following Fortran-specific issues:

---

### 🔍 **Common Causes in Fortran**

1. **Out-of-bounds array access**

   * Fortran does **not** check array bounds unless explicitly compiled with bounds checking.
   * For example:

     ```fortran
     real, allocatable :: arr(:)
     allocate(arr(10))
     arr(11) = 1.0  ! <- invalid access, no error unless bounds checking is on
     ```

2. **Use of uninitialized pointers or allocatables**

   * Deallocating something not allocated, or reusing pointers improperly, can corrupt memory.

     ```fortran
     real, pointer :: p(:)
     deallocate(p)  ! <- invalid if never allocated
     ```

3. **Double `deallocate()`**

   * Fortran does not always protect against this unless you're careful.

4. **Mixing C interop with Fortran allocatables**

   * If you’re using `ISO_C_BINDING` and passing Fortran arrays to C (e.g., through OpenACC, CUDA Fortran, or MPI buffers), and there's a mismatch in memory handling, heap corruption can result.

---

### 🛠️ **Recommended Actions**

#### ✅ Compile with bounds and memory checks:

Add flags to **enable runtime checking**:

* For **gfortran**:

  ```bash
  -g -O0 -fcheck=all -fbacktrace -Wall
  ```
* For **ifort** (Intel Fortran):

  ```bash
  -g -check all -traceback -warn all
  ```

#### ✅ Use `valgrind` with MPI wrapper or standalone:

```bash
valgrind ./simulation
```

Or with MPI:

```bash
mpirun -n 1 valgrind ./simulation
```

#### ✅ Reproduce with small input

Try to isolate the problem with a **minimal input case**. Heap corruption is often non-deterministic, so smaller runs can help localize it.

#### ✅ Consider AddressSanitizer (if using gfortran):

Modern `gfortran` supports it:

```bash
-fsanitize=address -fsanitize=undefined -g
```

> ⚠️ This slows down execution but will immediately point out invalid memory use.

Malmahrouqi3:
So I made the following derived type, but the build always corrupts mainly in the post-process because sys_size clashes with all processes while being in the form `eqn_idx%sys_size` for some reason. So for now, I will undo all changes pertaining to sys_size. Other than that, all ought to be sound and functional I hope so.
``` fortran
    !> @name Annotations of the structure of the state and flux vectors in terms of the
    !! size and the configuration of the system of equations to which they belong
    !> @{
    type system_of_equations
        integer :: sys_size                            !< Number of unknowns in system of eqns.
        type(int_bounds_info) :: cont                  !< Indexes of first & last continuity eqns.
        type(int_bounds_info) :: mom                   !< Indexes of first & last momentum eqns.
        integer :: E                                   !< Index of energy equation
        integer :: n                                   !< Index of number density
        type(int_bounds_info) :: adv                   !< Indexes of first & last advection eqns.
        type(int_bounds_info) :: internalEnergies      !< Indexes of first & last internal energy eqns.
        type(bub_bounds_info) :: bub                   !< Indexes of first & last bubble variable eqns.
        integer :: alf                                 !< Index of void fraction
        integer :: gamma                               !< Index of specific heat ratio func. eqn.
        integer :: pi_inf                              !< Index of liquid stiffness func. eqn.
        type(int_bounds_info) :: B                     !< Indexes of first and last magnetic field eqns.
        type(int_bounds_info) :: stress                !< Indexes of first and last shear stress eqns.
        type(int_bounds_info) :: xi                    !< Indexes of first and last reference map eqns.
        integer :: b_size                              !< Number of elements in the symmetric b tensor, plus one
        integer :: tensor_size                         !< Number of elements in the full tensor plus one
        type(int_bounds_info) :: species               !< Indexes of first & last concentration eqns.
        integer :: c                                   !< Index of color function
        integer :: damage                              !< Index of damage state variable (D) for continuum damage model
        integer, dimension(3) :: dir
        real(wp), dimension(3) :: dir_flg
        integer, dimension(3) :: dir_tau !!used for hypoelasticity=true
        integer, dimension(2) :: Re_size
        integer, allocatable, dimension(:, :) :: Re
    end type system_of_equations
```

src/simulation/m_ibm.fpp
```diff
@@ -872,21 +872,8 @@ contains
...
-                if (allocated(ghost_points(i)%ip%r)) deallocate (ghost_points(i)%ip%r)
-                if (allocated(ghost_points(i)%ip%v)) deallocate (ghost_points(i)%ip%v)
-                if (allocated(ghost_points(i)%ip%pb)) deallocate (ghost_points(i)%ip%pb)
-                if (allocated(ghost_points(i)%ip%mv)) deallocate (ghost_points(i)%ip%mv)
-                if (allocated(ghost_points(i)%ip%nmom)) deallocate (ghost_points(i)%ip%nmom)
-                if (allocated(ghost_points(i)%ip%presb)) deallocate (ghost_points(i)%ip%presb)
-                if (allocated(ghost_points(i)%ip%massv)) deallocate (ghost_points(i)%ip%massv)
-            end do
-        end if
+
```
Malmahrouqi3, line 875: @abbotts 

Do you mind checking out this line?
I suspect it to be causing all IBM cases to fail.

src/common/m_derived_types.fpp
```diff
@@ -166,6 +135,37 @@ module m_derived_types
...
+        integer :: b_size                              !< Number of elements in the symmetric b tensor, plus one
+        integer :: tensor_size                         !< Number of elements in the full tensor plus one
+        type(int_bounds_info) :: species               !< Indexes of first & last concentration eqns.
+        integer :: c                                   !< Index of color function
+        integer :: damage                              !< Index of damage state variable (D) for continuum damage model
+        integer, dimension(3) :: dir
+        real(wp), dimension(3) :: dir_flg
+        integer, dimension(3) :: dir_tau !!used for hypoelasticity=true
+        integer, dimension(2) :: Re_size
+        integer, allocatable, dimension(:, :) :: Re
```
sbryngelson, line 166: you may need to be careful with this on GPU cases (especially Frontier/Cray). i forget exactly how we deal with cases where we have an allocatable in a derived type but i would look for other examples of the code that do this. also, isn't Re a real and not an integer? and why is its size unknown at compile time?

sbryngelson:
> So I made the following derived type, but the build always corrupts mainly in the post-process because sys_size clashes with all processes while being in the form `eqn_idx%sys_size` for some reason. So for now, I will undo all changes pertaining to sys_size. Other than that, all ought to be sound and functional I hope so.
> 
> ```fortran
>     !> @name Annotations of the structure of the state and flux vectors in terms of the
>     !! size and the configuration of the system of equations to which they belong
>     !> @{
>     type system_of_equations
>         integer :: sys_size                            !< Number of unknowns in system of eqns.
>         type(int_bounds_info) :: cont                  !< Indexes of first & last continuity eqns.
>         type(int_bounds_info) :: mom                   !< Indexes of first & last momentum eqns.
>         integer :: E                                   !< Index of energy equation
>         integer :: n                                   !< Index of number density
>         type(int_bounds_info) :: adv                   !< Indexes of first & last advection eqns.
>         type(int_bounds_info) :: internalEnergies      !< Indexes of first & last internal energy eqns.
>         type(bub_bounds_info) :: bub                   !< Indexes of first & last bubble variable eqns.
>         integer :: alf                                 !< Index of void fraction
>         integer :: gamma                               !< Index of specific heat ratio func. eqn.
>         integer :: pi_inf                              !< Index of liquid stiffness func. eqn.
>         type(int_bounds_info) :: B                     !< Indexes of first and last magnetic field eqns.
>         type(int_bounds_info) :: stress                !< Indexes of first and last shear stress eqns.
>         type(int_bounds_info) :: xi                    !< Indexes of first and last reference map eqns.
>         integer :: b_size                              !< Number of elements in the symmetric b tensor, plus one
>         integer :: tensor_size                         !< Number of elements in the full tensor plus one
>         type(int_bounds_info) :: species               !< Indexes of first & last concentration eqns.
>         integer :: c                                   !< Index of color function
>         integer :: damage                              !< Index of damage state variable (D) for continuum damage model
>         integer, dimension(3) :: dir
>         real(wp), dimension(3) :: dir_flg
>         integer, dimension(3) :: dir_tau !!used for hypoelasticity=true
>         integer, dimension(2) :: Re_size
>         integer, allocatable, dimension(:, :) :: Re
>     end type system_of_equations
> ```

`sys_size` is used in so many places in the code that it's possible you missed something somewhere. also, did you make sure to `!$acc update device(eqn_idx%sys_size)` after it is set in `m_global_parameters`? i'm thinking about this line here https://github.com/MFlowCode/MFC/blob/db44da17d608bc94968ba48d344aadabd5161d7c/src/simulation/m_global_parameters.fpp#L1219

sbryngelson:
nitpick but probably can change the naming pattern of `bub_bounds_info` to `bub_bounds` and similiar. these types are used in the code a lot so it makes sense to keep the names of types short-ish.

sbryngelson:
FYI i moved pressure relaxation from RHS to its own module, which is why the merge conflict there looks so crazy

src/common/m_derived_types.fpp
Malmahrouqi3: I kinda omitted `_idx` from all variables listed. It was `Re_idx `and I just copied the declaration into under `system_of_equations `type.

.vscode/settings.json
sbryngelson: remove from pr please

src/simulation/include/inline_riemann.fpp
```diff
@@ -93,14 +93,14 @@
...
-                    (rho_R*(s_R - vel_R(dir_idx(1))) - rho_L*(s_L - vel_L(dir_idx(1))))* &
+                    (s_L - vel_L(eqn_idx%dir(1)))*(s_R - vel_R(eqn_idx%dir(1)))*(vel_R(eqn_idx%dir(1)) - vel_L(eqn_idx%dir(1)))/ &
+                    (rho_R*(s_R - vel_R(eqn_idx%dir(1))) - rho_L*(s_L - vel_L(eqn_idx%dir(1))))* &
                     (zcoef - 1._wp)
         else if (low_Mach == 2) then
-            vel_L_tmp = 5e-1_wp*((vel_L(dir_idx(1)) + vel_R(dir_idx(1))) + zcoef*(vel_L(dir_idx(1)) - vel_R(dir_idx(1))))
-            vel_R_tmp = 5e-1_wp*((vel_L(dir_idx(1)) + vel_R(dir_idx(1))) + zcoef*(vel_R(dir_idx(1)) - vel_L(dir_idx(1))))
-            vel_L(dir_idx(1)) = vel_L_tmp
-            vel_R(dir_idx(1)) = vel_R_tmp
+            vel_L_tmp = 5e-1_wp*((vel_L(eqn_idx%dir(1)) + vel_R(eqn_idx%dir(1))) + zcoef*(vel_L(eqn_idx%dir(1)) - vel_R(eqn_idx%dir(1))))
```
sbryngelson, line 100: i think you should put dir_idx into the derived type, it's used not as a global parameter of sorts but one that changes every time step. it's also used in different modules i believe, perhaps even as a local variable.

src/simulation/m_global_parameters.fpp
```diff
@@ -1021,32 +1000,32 @@ contains
...
+                if (fluid_pp(i)%Re(2) > 0) eqn_idx%Re_size(2) = eqn_idx%Re_size(2) + 1
             end do
 
-            if (Re_size(1) > 0._wp) shear_stress = .true.
-            if (Re_size(2) > 0._wp) bulk_stress = .true.
+            if (eqn_idx%Re_size(1) > 0._wp) shear_stress = .true.
+            if (eqn_idx%Re_size(2) > 0._wp) bulk_stress = .true.
 
-            !$acc update device(Re_size, viscous, shear_stress, bulk_stress)
+            !$acc update device(eqn_idx, viscous, shear_stress, bulk_stress)
```
sbryngelson, line 1010: i think this might be a problem. i'm not sure you can just `update device(eqn_idx)` on a derived type, especially one that includes allocatables. i could be wrong. look for other patterns in the code that need to do this.

src/simulation/m_start_up.fpp
```diff
@@ -1634,7 +1634,7 @@ contains
         if (chemistry) then
             !$acc update device(q_T_sf%sf)
         end if
-        !$acc update device(nb, R0ref, Ca, Web, Re_inv, weight, R0, V0, bubbles_euler, polytropic, polydisperse, qbmm, R0_type, ptil, bubble_model, thermal, poly_sigma, adv_n, adap_dt, adap_dt_tol, n_idx, pi_fac, low_Mach)
+        !$acc update device(nb, R0ref, Ca, Web, Re_inv, weight, R0, V0, bubbles_euler, polytropic, polydisperse, qbmm, R0_type, ptil, bubble_model, thermal, poly_sigma, adv_n, adap_dt, adap_dt_tol, eqn_idx%n, pi_fac, low_Mach)
```
sbryngelson, line 1637: yeah, see, here you update device with a specific variable inside of a derived type, but in m_global you tried to update device with the whole eqn_idx. i can't quite remember if that works or not.

src/simulation/m_time_steppers.fpp
```diff
@@ -941,8 +941,8 @@ contains
 
         elseif (bubbles_lagrange) then
 
-            call s_populate_variables_buffers(q_prim_vf, pb_ts(1)%sf, mv_ts(1)%sf, bc_type)
-            call s_compute_bubble_EL_dynamics(q_prim_vf, stage)
+            call s_populate_variables_buffers(q_prim_vf, pb_ts(1)%sf, mv_ts(1)%sf, bc_type, bc_bound)
+            call s_compute_bubble_EL_dynamics(q_cons_ts(1)%vf, q_prim_vf, t_step, rhs_vf, stage)
```
sbryngelson, line 945: is this correct? looks like a breaking change to the subroutine call? or maybe this is from a merge conflict you fixed

src/common/m_derived_types.fpp
```diff
@@ -135,6 +135,37 @@ module m_derived_types
         integer, dimension(:, :), allocatable :: moms !< Moment indices for qbmm
```
sbryngelson, line 135: Interestingly, this derived type appears never to be used. git blame says I added it years ago.

src/common/m_derived_types.fpp
sbryngelson: yep this works

sbryngelson: have a look at https://github.com/MFlowCode/MFC/blob/db44da17d608bc94968ba48d344aadabd5161d7c/src/simulation/m_acoustic_src.fpp#L59 to see how derived types that have allocatables in them are handled. that said, you only have one allocatable in the type. i suggest removing it and "putting back" the old Re_idx. this should make life much simpler as a stepping stone. notice that basically all of the other types that are used do not have allocatables in them. they require special treatment.

sbryngelson (commented):
i think i found a couple of the pain points here. look at the comments i left.

Malmahrouqi3:
Many conflicts to resolve, outdated changes to edit, and bugs to fix.

Therefore, I will close the PR in order to break it down into two (EI & BC) - easing up debugging, resolving conflicts, and book keeping. Following the same approach pursued, I will re-apply the same changes - not blindly copy-paste . I will address all review comments provided.

PR #872, opened 2025-06-09
deallocate fix for openacc

### LINKED ISSUE #871: ACC exit data and deallocate are in reverse order in the macro
ACC exit data and deallocate are in reverse order in the `@:DEALLOCATE()` macro. Switch and verify working.

sbryngelson:
attempt to fix #871


sbryngelson:
I think QBMM needs a finalize module subroutine to deallocate module vars

sbryngelson:
fyi @abbotts @bmuralid  this is fixed now (along with one other bug).

Issue #750, opened 2024-11-30
Benchmark CI failure/improvement


sbryngelson:
Right now the benchmark CI workflow prints useful information about grind time and runtime to the CI output. This CI workflow always passes, so long as the code executes.

I would like the CI to **fail** if the master and PR runtime or grind times are sufficiently different (within some tolerance). Without this, it is easy to have a performance regression but not see it because one did not check CI carefully enough.

Malmahrouqi3:
will look into the CI workflow and see if doable

PR #876, opened 2025-06-11
Tolerances for CI Workflow - Grind & Exec Times (#750)

mohdsaid497566:
## Description

Enhancement for CI workflow. Raises MFCException if grind time is less than an acceptable threshold. For Exec, it prints an error.

### DIFF
--- a/toolchain/mfc/bench.py
+++ b/toolchain/mfc/bench.py
@@ -81,89 +81,62 @@ def bench(targets = None):
 # pylint: disable=too-many-branches
 def diff():
     lhs, rhs = file_load_yaml(ARG("lhs")), file_load_yaml(ARG("rhs"))
-
     cons.print(f"[bold]Comparing Benchmarks: Speedups from [magenta]{os.path.relpath(ARG('lhs'))}[/magenta] to [magenta]{os.path.relpath(ARG('rhs'))}[/magenta] are displayed below. Thus, numbers > 1 represent increases in performance.[/bold]")
+    
     if lhs["metadata"] != rhs["metadata"]:
-        def _lock_to_str(lock):
-            return ' '.join([f"{k}={v}" for k, v in lock.items()])
-
-        cons.print(f"""\
-[bold yellow]Warning[/bold yellow]: Metadata in lhs and rhs are not equal.
-    This could mean that the benchmarks are not comparable (e.g. one was run on CPUs and the other on GPUs).
-    lhs:
-    * Invocation: [magenta]{' '.join(lhs['metadata']['invocation'])}[/magenta]
-    * Modes:      {_lock_to_str(lhs['metadata']['lock'])}
-    rhs:
-    * Invocation: {' '.join(rhs['metadata']['invocation'])}
-    * Modes:      [magenta]{_lock_to_str(rhs['metadata']['lock'])}[/magenta]
-        """)
+        _lock_to_str = lambda lock: ' '.join([f"{k}={v}" for k, v in lock.items()])
+        cons.print(f"[bold yellow]Warning[/bold yellow]: Metadata in lhs and rhs are not equal.\n"
+                  f"   This could mean that the benchmarks are not comparable (e.g. one was run on CPUs and the other on GPUs).\n"
+                  f"         Invocation,\t\t Modes\n"
+                  f"    lhs: {' '.join(lhs['metadata']['invocation'])} | {_lock_to_str(lhs['metadata']['lock'])}\n"
+                  f"    rhs: {' '.join(rhs['metadata']['invocation'])} | {_lock_to_str(rhs['metadata']['lock'])}")
 
     slugs = set(lhs["cases"].keys()) & set(rhs["cases"].keys())
     if len(slugs) not in [len(lhs["cases"]), len(rhs["cases"])]:
         cons.print(f"""\
-[bold yellow]Warning[/bold yellow]: Cases in lhs and rhs are not equal.
+[bold yellow]Warning[/bold yellow]: Cases in lhs and rhs differ. Using intersection: {slugs} with {len(slugs)} elements.
     * rhs cases: {', '.join(set(rhs['cases'].keys()) - slugs)}.
-    * lhs cases: {', '.join(set(lhs['cases'].keys()) - slugs)}.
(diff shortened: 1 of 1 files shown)


wilfonba:
I picked the 0.98 tolerance for this. I looked through some old CI runs, and the grind times reported were basically all 0.99 or 1.00, so something worse than 0.98 seemed like it was worth looking at. @sbryngelson can comment on whether he thinks this is a reasonable tolerance or not.

sbryngelson:
@mohdsaid497566 @wilfonba 
Right now, the thing that I hate the most is that the benchmark CI fails about 30% of the time due to some slurm/filesystem error. I have to re-run benchmarking a few times in some cases to get it to finish without error. This used to have an 80% failure rate, but I made it somewhat more robust. Nothing you guys can fix, but one day I'll get with PACE to fix it up. Until then, having this isn't super useful since benchmarking fails spuriously already (so those would become false positives basically).

I really like the idea of this, but I would much rather have continuous benchmarking, which is more robust/repeatable, and insightful (fixing this PR is just a bit of a convenience for me).

Malmahrouqi3:
Sample benchmark output. It needs a minor fix.
``` txt
 Comparing Benchmarks: Speedups from ../master/bench-cpu.yaml to bench-cpu.yaml are displayed below. Thus, numbers > 1 represent increases in performance.
 Warning: Exec time speedup for pre_process is less than 0.9 - Case: ibm
                                                                                
  Case                      Pre Process          Simulation       Post Process  
 ────────────────────────────────────────────────────────────────────────────── 
  5eq_rk3_weno3_hllc   Exec: Exec: 1.08    Exec: Exec: 1.00   Exec: Exec: 1.15  
                                             & Grind: N/A &                     
                                                Grind: 1.00                     
  ibm                  Exec: Exec: 0.84    Exec: Exec: 1.08   Exec: Exec: 1.08  
                                             & Grind: N/A &                     
                                                Grind: 1.09                     
  viscous_weno5_sgb…   Exec: Exec: 1.04    Exec: Exec: 1.00   Exec: Exec: 1.01  
                                             & Grind: N/A &                     
                                                Grind: 1.01                     
  hypo_hll             Exec: Exec: 1.02    Exec: Exec: 0.99   Exec: Exec: 0.99  
                                             & Grind: N/A &                     
                                                Grind: 1.02                     
                                                                                

```

sbryngelson:
We only really care about simulation grind time and exec. time and the pre/post process times aren't reliable anyway. we could probably even just get rid of them.

sbryngelson:
can you make the PR source code slow so we can see it fail benchmarking?

toolchain/mfc/bench.py
```diff
@@ -56,11 +56,11 @@ def bench(targets = None):
 
         with open(log_filepath, "w") as log_file:
             system(
-                ["./mfc.sh", "run", case.path, "--case-optimization"] +
+                ["./mfc.sh", "run", case.path] +
                 ["--targets"] + [t.name for t in targets] +
                 ["--output-summary", summary_filepath] +
                 case.args +
-                ["--", "--gbpp", ARG('mem')],
+                ["--", "--gbpp", 0.5],
```
Malmahrouqi3: I removed case optimization and capped gbpp to only 0.5.
This should slow all cases down drastically. I am not sure but It might exceed the allocated time. If happens to be the case, I will hop into Phoenix and figure out proper time for the bench job until I reach grind time failure mode.

Malmahrouqi3: Currently testing the failure mode on Delta and will post the outcomes here
I guess I will cap just the memory per process and keep case optimization to make it run slower but do not take forever

Malmahrouqi3: I overthought over it needlessly. I can just induce an artificial failure by modifying the grind times in the pr/master benchmark yaml files locally then run `bench_diff`.

Malmahrouqi3:
Screenshot 2025-06-15 165001

Malmahrouqi3:
Failed indeed as intended to be. To replicate the artificial failure, download benchmark yaml artifacts from any PR results then tweak the numbers to create unreasonable discrepancies. Finally, run `./mfc.sh bench_diff master/bench-cpu.yaml pr/bench-cpu.yaml`
Now, I will undo the last commit changes.

toolchain/mfc/bench.py
Malmahrouqi3: it worked out luckily

sbryngelson:
very good thanks!

sbryngelson:
it looks like it failed the CPU benchmark because the exec. time was below 0.9 on preprocess? we should only run the test for the difference in speed on simulation

Malmahrouqi3:
@sbryngelson no, grind time is below threshold actually - 0.97. Ours is set for >=0.98. Exec time is meant to throw only warnings but cant terminate the job.

toolchain/mfc/bench.py
```diff
@@ -114,56 +113,39 @@ def _lock_to_str(lock):
...
+                    cons.print(f"[bold yellow]Warning[/bold yellow]: Exec time speedup for {target.name} is less than 0.9 - Case: {slug}")
+                speedups[i] = f"Exec: {exec_time_value:.2f}"
+                if target == SIMULATION:
+                    if not math.isfinite(lhs_summary[target.name]["grind"]) or not math.isfinite(rhs_summary[target.name]["grind"]):
+                        err = 1
+                        cons.print(f"lhs_summary or rhs_summary reports non-real grind time for {target.name} - Case: {slug}")
+
+                    grind_time_value = lhs_summary[target.name]["grind"] / rhs_summary[target.name]["grind"]
+                    speedups[i] += f" & Grind: {grind_time_value:.2f}"
+                    if grind_time_value <0.98:
```
Malmahrouqi3: threshold for grind time

sbryngelson: got it. well i think 0.98 is too strict against @wilfonba's suggestion. I would use 0.95 for now, can adjust later.

Malmahrouqi3: kinda stringent I believe so. I will change it now.

sbryngelson:
cool it will probably pass and i will merge. thanks!

PR #878, opened 2025-06-11
Test Suite Flag `--rdma-mpi` Implemented (#598)

Malmahrouqi3:
## Description
I added the flag under `./mfc.sh test`. It can be invoked with `-r` or `--rdma-mpi`
It can be used on any test cases. It is intended to skip tests whenever the `rdma_mpi `is false or unspecified (false by default).

Closes https://github.com/MFlowCode/MFC/issues/598

``` bash
/rdma_mpi/MFC-mo2$ ./mfc.sh test  -r -o HLLD

Test Summary: 0 passed, 0 failed, 2 skipped.

 Skipped Cases

 1D -> MHD -> HLLD
 2D -> MHD -> HLLD
```


Malmahrouqi3:
Six 2-Rank MPI cases will be created. Pairs with the RDMA enabled and disabled. RDMA-On cases will be run only when the flag is given. Otherwise, they will be written but skipped.
> 0FCCE9F1   1D -> 2 MPI Ranks
>   8C7AA13B   2D -> 2 MPI Ranks
>   CE232828   3D -> 2 MPI Ranks


**Example Usage:**
``` bash
./mfc.sh test --rdma-mpi                                        #only MPI Ranks cases.
./mfc.sh test -o Hypoelasticity --rdma-mpi                      #Hypoelasticity cases with the RDMA enabled.
./mfc.sh test -a --rdma-mpi                                     #all tests along with RDMA 2-Ranks MPI cases.
```
--------------------
```bash
./mfc.sh test --help
```
```txt
  -r, --rdma-mpi        Enable RDMA MPI for tests (default: False)
  --no-build            (Testing) Do not rebuild MFC. (default: False)
  --no-examples         Do not test example cases. (default: False)
  --case-optimization   (GPU Optimization) Compile MFC targets with some case parameters hard-coded. (default: False)
  --dry-run             Build and generate case files but do not run tests. (default: False)
  --generate            (Test Generation) Generate golden files. (default: False)
  --add-new-variables   (Test Generation) If new variables are found in D/ when running tests, add them to the golden files. (default: False)
  --remove-old-tests    (Test Generation) Delete tests directories that are no longer. (default: False)
```

Malmahrouqi3:
Just need to update the testing docs for the flag and add `./mfc.sh test -a --rdma-mpi` to Frontier CI.

sbryngelson:
I'm not sure if this is the case right now, but if one enables the flag `--rdma-mpi` (i would remove the `-r` option for it) then it should run _all_ the cases, not only the ones with RDMA MPI. that's what i was trying to suggest.

Malmahrouqi3:
@sbryngelson 
When the flag is enabled, it should run whatever category specified or just all + RDMA cases.

toolchain/mfc/test/test.py
```diff
@@ -179,8 +188,13 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
     if ARG("dry_run"):
         cons.print(f"  [bold magenta]{case.get_uuid()}[/bold magenta]     SKIP     {case.trace}")
         return
+   
+    if ARG("rdma_mpi"):
+        cmd = case.run([PRE_PROCESS, SIMULATION], gpus=devices)
+        cmd = case.run([PRE_PROCESS, SIMULATION], gpus=devices, rdma_mpi=True)
+    else:
+        cmd = case.run([PRE_PROCESS, SIMULATION], gpus=devices)
```
Malmahrouqi3: @sbryngelson this block (line 192-196) would run whatever cases then those with RDMA.

Malmahrouqi3:
@sbryngelson I do not think it is viable to debug locally. I do not have access to Frontier, is RDMA available on (Phoenix, Bridges2, or Delta)??

sbryngelson:
> @sbryngelson I do not think it is viable to debug locally. I do not have access to Frontier, is RDMA available on (Phoenix, Bridges2, or Delta)??

You can ask the MFC Slack what other computers have compilers that support it. But the Frontier failure appears to have nothing to do with your PR. It's having some network issues.

Malmahrouqi3:
Working on it right now on Phoenix

Malmahrouqi3:
This should do it hopefully. The difference between the two commands is right now 3 tests. I will keep an eye on the test suite
`./mfc.sh test -j $(nproc) -a
`
`./mfc.sh test -j $(nproc) -a --rdma-mpi`

I removed `-r` so it only can be run by `--rdma-mpi` at this moment.

Malmahrouqi3:
Self-Hosted Frontier failed due to irrelevant reasons at all. I will rerun momentarily.

Malmahrouqi3:
@anandrdbz Self-Hosted Frontier has failed on multiple re-runs for unclear reasons. Do you mind checking it out?

sbryngelson:
the problem seems to be that it's trying to use the internet on a compute node (in the test part of the workflow). the compute nodes don't have internet access. it's trying to access silo. I am not sure why it thinks it isn't available, because all of the other PR workflows are working fine... which suggests something specific to this PR

Malmahrouqi3:
I have not clue to be honest

sbryngelson:
frankly me either, your PR doesn't seem to tickle that part of the code. i would try commenting out some of your changes until it works. not an elegant method. probably leave the flag but comment out the extra test cases and how you modify the case files, and make sure that works. if that works, add in how you modify the case files but not the tests themselves.

as i look closer, i'm unsure if how you added the RDMA flag is proper - though i'm not entirely sure. what happens if you try to use it on Phoenix or something? If you run the `./mfc.sh test --dry-run -a` and then run `./mfc.sh test --rdma-mpi -a` does it try to build silo again? You can just try that out on Phoenix on your own (or whatever computer, for that matter).

Malmahrouqi3:
on Phoenix it ran just fine, but failed on the RDMA cases only

Malmahrouqi3:
I will keep playing with it tonight

sbryngelson:
basically the idea of `--dry-run` is that it builds _everything_ needed to run all of the tests. if you messed up the flags somehow, then it may try to build MFC again on the compute node automatically. (e.g., it sees a new flag `dry-run` and it says "i need to rebuild!" and that's the problem). 

frankly i'm pretty confident now that something like that is happening. MFC rebuilds itself when it sees certain flag differences. 
for example, if you compile with `--gpu` and then run with `--no-gpu` it will rebuild MFC for no-gpu cases. i think it may see the flag difference during testing/running and think it needs to rebuild. i could be wrong.

sbryngelson:
> on Phoenix it ran just fine, but failed on the RDMA cases only

well, in this case i know it should run fine, but the question is _did it try to rebuild silo_ when you run test *after you run `./mfc.sh test --dry-run`*

Malmahrouqi3:
I will wait for the re-run now and if it does not pass, I will try the flag on different clusters to investigate the issue further

sbryngelson:
Yeah i don't think you even need a cluster. you could probably even do this on your laptop. you just need to see if it tries to fetch silo from github because that's the failure mode.

toolchain/mfc/test/test.py
```diff
@@ -224,6 +232,10 @@ def _handle_case(case: TestCase, devices: typing.Set[int]):
     if ARG("test_all"):
         case.delete_output()
         cmd = case.run([PRE_PROCESS, SIMULATION, POST_PROCESS], gpus=devices)
```
Malmahrouqi3: Bug here. Duplicate of the below basically but **`cmd`** is gonna be re-written anyways.

sbryngelson: good find

toolchain/mfc/test/test.py
sbryngelson: yep

Malmahrouqi3:
checked the logs and 
Phoenix gpu benchmark failed since the master branch oddly did not complete all benchmark tests.
Self-Hosted Frontier cpu failed from the compiler end.

sbryngelson:
interesting copilot also generated these but left them in a dropdown not a comment because it had low confidence

<summary>Comments suppressed due to low confidence (4)</summary>

**toolchain/mfc/args.py:83**
* The implementation documentation mentions a short form `-r` for this flag, but it isn’t defined here. Consider adding `-r` alongside `--rdma-mpi` for consistency with the description.
```
    test.add_argument(      "--rdma-mpi",     action="store_true",     default=False,     help="Run tests with RDMA MPI enabled")
```
**toolchain/mfc/test/test.py:60**
* New filtering logic skips RDMA MPI cases when the flag is unset; consider adding unit or integration tests for `__filter` to verify both skip and include behaviors.
```
    for case in cases[:]:
```
**toolchain/mfc/test/test.py:234**
* The RDMA-enabled path for the post-process step is commented out under `test_all`, so `--rdma-mpi` has no effect there. You should either restore that logic or remove the commented block to avoid misleading code.
```
        # if ARG("rdma_mpi"):
```
**toolchain/mfc/test/test.py:190**
* [nitpick] This `else` branch is indented with two spaces instead of four, which breaks the project’s indentation style. Please adjust for consistency.
```
        cmd = case.run([PRE_PROCESS, SIMULATION], gpus=devices)
```

Malmahrouqi3:
Will cruise back into this PR in a sec

Malmahrouqi3:
[github/workflows/frontier/test.sh](https://github.com/MFlowCode/MFC/pull/878/files#diff-c5d4a0387e2130964b29910426e48e9b705ee21ce6339a68193c472e84fa3db5) has nothing sketchy looking tbh, just normal intended use of the flag.
[docs/documentation/testing.md](https://github.com/MFlowCode/MFC/pull/878/files#diff-f79b9d6ca01899b691bce5e2ea36ec83ccd186f0c296c0f028ebae6bc60ed38d) removed -r as you requested from both code and docs.
[toolchain/mfc/args.py](https://github.com/MFlowCode/MFC/pull/878/files#diff-5aeb20baff8d1d62ad5b7fe4d75a0697e743d6a25a843ec3231a7dc6456f610a) recognized the flag in general.
`    test.add_argument(      "--rdma-mpi",     action="store_true",     default=False,     help="Run tests with RDMA MPI enabled")`

[toolchain/mfc/test/cases.py](https://github.com/MFlowCode/MFC/pull/878/files#diff-fb12c96151a1a899dbb894f0c6145dbed66bc200374699b694b66a52932b47da) intended to append the additional cases to the existing list.
``` python
    def alter_ppn(dimInfo):
        if len(dimInfo[0]) == 3:
            cases.append(define_case_d(stack, '2 MPI Ranks', {'m': 29, 'n': 29, 'p': 49}, ppn=2))
            cases.append(define_case_d(stack, '2 MPI Ranks -> RDMA MPI', {'m': 29, 'n': 29, 'p': 49, 'rdma_mpi': 'T'}, ppn=2))
        else:
            cases.append(define_case_d(stack, '2 MPI Ranks', {}, ppn=2))
            cases.append(define_case_d(stack, '2 MPI Ranks -> RDMA MPI', {'rdma_mpi': 'T'}, ppn=2))
```

[toolchain/mfc/test/test.py](https://github.com/MFlowCode/MFC/pull/878/files#diff-f6e8934e27b5173e5e507afd442b7e24ef61cf90dcb204dcc76f3eb9437b63ac) skips RMDA cases if not enabled
``` python
    for case in cases[:]:
        if "RDMA MPI" in case.trace and not ARG("rdma_mpi"):
            cases.remove(case)
            skipped_cases.append(case)
```

sbryngelson:
The persisting problem is that your flag is tripping MFC to try to build itself again (or at least build silo, which is a dependency of post_process). Frontier compute nodes don't have access to the internet so this fails.

Malmahrouqi3:
I do not know why there is solely a single block that is supposed to run in [toolchain/mfc/test/test.py](https://github.com/MFlowCode/MFC/pull/878/files#diff-f6e8934e27b5173e5e507afd442b7e24ef61cf90dcb204dcc76f3eb9437b63ac)
If removed, the flag would be part of the args intake but not the execution.

sbryngelson:
> I do not know why there is solely a single block that is supposed to run in [toolchain/mfc/test/test.py](https://github.com/MFlowCode/MFC/pull/878/files#diff-f6e8934e27b5173e5e507afd442b7e24ef61cf90dcb204dcc76f3eb9437b63ac) If removed, the flag would be part of the args intake but not the execution.

Not sure what this means. In the end, it just needs to tell the toolchain what cases to run. It doesn't matter much to me how it gets there.

Malmahrouqi3:
Rather subtle, for the current implementation, RDMA tests themselves trigger the re-build as they are not part of `--dry-run` in `./github/workflows/frontier/build.sh`.

The files show the difference between using the flag and not in the dry run.
[WO_RDMA.txt](https://github.com/user-attachments/files/20959430/WO_RDMA.txt)
[W_RDMA.txt](https://github.com/user-attachments/files/20959431/W_RDMA.txt)

**Edit 1**: the module-load & dry-run occur on the login node, so that kinda explains a lot why it was attempting to clone silo while offline on the compute node.

**Edit 2**: Silo still is attempted to being cloned. Maybe, --no-build flag would do something.

**Edit 3**: I checked some PR's self-hosted frontier tests. The keyword 'silo', except for this PR, does not exist in the logs at all, therefore I am highly suspecting of make instructions.

sbryngelson:
> Rather subtle, for the current implementation, RDMA tests themselves trigger the re-build as they are not part of `--dry-run` in `./github/workflows/frontier/build.sh`.
> 
> The files show the difference between using the flag and not in the dry run. [WO_RDMA.txt](https://github.com/user-attachments/files/20959430/WO_RDMA.txt) [W_RDMA.txt](https://github.com/user-attachments/files/20959431/W_RDMA.txt)
> 
> **Edit 1**: the module-load & dry-run occur on the login node, so that kinda explains a lot why it was attempting to clone silo while offline on the compute node.
> 
> **Edit 2**: Silo still is attempted to being cloned. Maybe, --no-build flag would do something.
> 
> **Edit 3**: I checked some PR's self-hosted frontier tests. The keyword 'silo', except for this PR, does not exist in the logs at all, therefore I am highly suspecting of make instructions.

Silo doesn't appear with Frontier tests because we never used `-a` with Frontier tests due to wall clock time limitations...

Malmahrouqi3:
I will edit frontier template to reflect that no-need for silo. I will wait for the self-hosted tests to pass, then push another change.

.github/workflows/frontier/test.sh
```diff
@@ -4,7 +4,7 @@ gpus=`rocm-smi --showid | awk '{print $1}' | grep -Eo '[0-9]+' | uniq | tr '\n'
 ngpus=`echo "$gpus" | tr -d '[:space:]' | wc -c`
 
 if [ "$job_device" = "gpu" ]; then
-    ./mfc.sh test --max-attempts 3 -j $ngpus -- -c frontier
+    ./mfc.sh test -a --rdma-mpi --max-attempts 3 -j $ngpus -- -c frontier
 else
-    ./mfc.sh test --max-attempts 3 -j 32 -- -c frontier
+    ./mfc.sh test -a --rdma-mpi --max-attempts 3 -j 32 -- -c frontier
```
sbryngelson, line 9: this case doesn't make sense because RDMA is only for GPUs

Malmahrouqi3: Never knew that. I will make it GPU-exclusive.

toolchain/mfc/test/test.py
```diff
@@ -120,6 +124,9 @@ def test():
     # Some cases require a specific build of MFC for features like Chemistry,
     # Analytically defined patches, and --case-optimization. Here, we build all
     # the unique versions of MFC we need to run cases.
+    if ARG("rdma_mpi") == True:
+        ARGS()["test_all"] = False
```
Malmahrouqi3: This presumably should re-set `test_all` to False if `rdma` flag is present. Just remembered to correct the argument to be while only on a dry-run as well. Otherwise, regardless of cluster and outside dry-run, the flag should work as anticipated to include/exclude RDMA tests.

Malmahrouqi3: > This presumably should re-set `test_all` to False if `rdma` flag is present. Just remembered to correct the argument to be while only on a dry-run as well. Otherwise, regardless of cluster and outside dry-run, the flag should work as anticipated to include/exclude RDMA tests.

    if ARG("rdma_mpi") and ARG("dry_run"):
        ARGS()["test_all"] = False

sbryngelson: this is right because `rdma_mpi` doesn't require rebuilding

sbryngelson (approved):

sbryngelson:
Hmm something wrong now...

```shell
 Failed Cases

 Failed test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 TestCase.run() got an unexpected keyword argument 'rdma_mpi'
 Failed test tests/1B300F28: 2D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 TestCase.run() got an unexpected keyword argument 'rdma_mpi'
 Failed test tests/2C9844EF: 3D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 TestCase.run() got an unexpected keyword argument 'rdma_mpi'
```

Edit/update: Looks like this has been an error for several commits.

sbryngelson (approved):

Malmahrouqi3:
``` 
 Failed test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI after 1 attempt(s).
 Test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI: Failed to execute MFC.
 ```
Prior, it was when run on Phoenix
 ``` 
 Failed test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 TestCase.run() got an unexpected keyword argument 'rdma_mpi'
 ```

**Comment by Malmahrouqi3** [comment]
Self Hosted (frontier) were canceled and needs to be re-run whenever convenient. appreciate that.

**Comment by sbryngelson** [comment]
yeah frontier keeps failing, I suspect someone has been using the computer for a long time or there's some backlog of jobs. I'm going to look shortly and try to fix it up -- hopefully easy fix.

**Review by sbryngelson — APPROVED** [review]

**Comment by Malmahrouqi3** [comment]
``` 
Failed Cases

 Failed test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 Test tests/FA4D8FEF: 1D -> 2 MPI Ranks -> RDMA MPI: The golden file does not exist! To generate golden files, use the '--generate' flag.
 Failed test tests/1B300F28: 2D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 Test tests/1B300F28: 2D -> 2 MPI Ranks -> RDMA MPI: The golden file does not exist! To generate golden files, use the '--generate' flag.
 Failed test tests/2C9844EF: 3D -> 2 MPI Ranks -> RDMA MPI after 3 attempt(s).
 Test tests/2C9844EF: 3D -> 2 MPI Ranks -> RDMA MPI: The golden file does not exist! To generate golden files, use the '--generate' flag.
```

**Comment by sbryngelson** [comment]
Umm, yeah, this makes sense. I would create the goldenfile on an RDMA-capable computer and call it a day. This should work since you 'SKIP' the non-RDMA case when appropriate, or skip RDMA case when appropriate. Unless this is because the goldenfile on the master branch doesn't exist? Or is something else going wrong that I'm not noticing?

**Comment by Malmahrouqi3** [comment]
yup, I could not find their golden files anywhere.

**Comment by sbryngelson** [comment]
In that case, failure is expected. Tests work by comparing the UUID's goldenfile to the results that come out. There is a clean way to generate goldenfiles via `test --generate` (or something similar). You would only commit the _new_ goldenfiles (associated with the new UUIDs).

Note that the non-RDMA tests have different UUIDs (hashes):
```shell
[I] MFC/MFC $ ./mfc.sh test -l -o 2\ MPI\ Ranks
mfc: OK > (venv) Entered the Python 3.11.7 virtual environment (>= 3.9).

      .=++*:          -+*+=.        | sbryngelson@login03 [Linux]
     :+   -*-        ==   =* .      | ---------------------------
  :*##-.....:*+   .#%+++=--+=:::.   | --jobs 1
  -=-++-======#=--**+++==+*++=::-:. | --mpi --gpu --no-debug --no-gcov --no-unified --no-single
 .:++=----------====+*= ==..:%..... |

         MFC Test Cases

    UUID     Trace
 ──────────────────────────────
  0FCCE9F1   1D -> 2 MPI Ranks
  8C7AA13B   2D -> 2 MPI Ranks
  CE232828   3D -> 2 MPI Ranks
```

PR #889, opened 2025-06-16
`PMD` Comments Removal

Malmahrouqi3:
## Description

This is subsequent to (#882) with the mere addition of filtering out inline comments and commented lines. For the PMD details, check out the original issue (https://github.com/MFlowCode/MFC/issues/646)

To show off the difference:
Filter-off: [pmd-old branch](https://github.com/Malmahrouqi3/MFC-mo2/tree/pmd-old) https://github.com/Malmahrouqi3/MFC-mo2/actions/runs/15693861383/job/44214858621
Filter-on: [pmd-new branch](https://github.com/Malmahrouqi3/MFC-mo2/tree/pmd-new) https://github.com/Malmahrouqi3/MFC-mo2/actions/runs/15694477828/job/44216673771

Improves PMD.


.github/workflows/pmd.yml
```diff
@@ -30,7 +30,50 @@ jobs:
...
+                  sed -E \
+                      -e '/^\s*!/s/.*//' \
+                      -e '/^[cC*dD]/s/.*//' \
+                      -e '/^[ \t]*[cC*dD]/s/.*//' \
+                      -e 's/([^"'\'']*("[^"]*"[^"'\'']*|'\''[^'\'']*'\''[^"'\'']*)*[^"'\'']*)[!].*$/\1/' \
+                      "$file" > "$TMP_FILE"
+                  
+                  # Check if file was actually modified before replacing
+                  if cmp -s "$file" "$TMP_FILE"; then
+                      echo -e "No changes needed for $file"
```
sbryngelson, line 120: remove echo statements it just makes the output harder to read

.github/workflows/pmd.yml
```diff
@@ -30,7 +30,50 @@ jobs:
...
+                      rm "$TMP_FILE"
+                  else
+                      # Overwrite the original file with the processed content
+                      mv "$TMP_FILE" "$file"
+                      echo -e "Successfully processed $file"
```
sbryngelson, line 125: remove

Malmahrouqi3:
@sbryngelson the number of violations is identical (63) whether the filter is on or off.

sbryngelson:
strange

sbryngelson:
idk about the number of violations, but the violations are different. This looks like an improvement to me.

.github/workflows/pmd.yml
```diff
@@ -30,7 +30,50 @@ jobs:
...
+                  #    - Lines starting with 'c', 'C', '*', 'd', 'D' in column 1 (fixed form comments)
+                  # 2. Remove end-of-line comments (anything after '!' that isn't in a string)
+                  # 3. Preserve strings containing '!' characters
+                  sed -E \
+                      -e '/^\s*!/s/.*//' \
+                      -e '/^[cC*dD]/s/.*//' \
+                      -e '/^[ \t]*[cC*dD]/s/.*//' \
+                      -e 's/([^"'\'']*("[^"]*"[^"'\'']*|'\''[^'\'']*'\''[^"'\'']*)*[^"'\'']*)[!].*$/\1/' \
+                      "$file" > "$TMP_FILE"
+
```
sbryngelson: also remove line breaks: https://chatgpt.com/s/t_6850a2d138c0819187272dc044cfb208

```shell
sed -E '
  :a                      # ① label for looping
  /&$/ {                  # ② if the *current* line ends with “&”
        N                 #    read the *next* line into the pattern space
        s/&[[:space:]]*\n[[:space:]]*&/ /   # ③ delete the trailing “&”,
                                            #    the following newline, the
                                            #    leading “&” on the next line,
                                            #    and join them with one space
        ba                # ④ repeat until the next line no longer starts
                          #    with “&” (handles runs of many “&” lines)
  }
  s/&//g                  # ⑤ finally strip any stray “&” left in the file
'  input.txt  >  output.txt
```

(i haven't checked to see if this works myself). PS this will mess up the line numbers but whatever.

sbryngelson: this might break for openacc line continuations which have something like

```
!$acc <......> &
!$acc <......>
```

idk what it will do to code that looks like this.

Malmahrouqi3: It deems them as comments as far as I know

sbryngelson: supposedly this will work for the acc statements but the first command might break acc w/ line continuation. i would run the acc one first, then the one above

```
sed -E '
  :a
  /[[:space:]]*&[[:space:]]*$/ { N; s/[[:space:]]*&[[:space:]]*\n[[:space:]]*/ /; ba }
  s/^[[:space:]]*!\$acc[[:space:]]*/!$acc /
  :b
  s/[[:space:]]+!\$acc[[:space:]]*/ /g
  tb
'  source.f90  >  clean.f90
```

sbryngelson: > It deems them as comments as far as I know

it has no notion of comments, it just looks at text. i'm trying to make the text as simple as possible, and having line continuations in different places is not a meaningful code difference so we should remove them before CPD

Malmahrouqi3: There were some syntax errors and I got rid off em. Also I lowered the number of tokens=20 kinda semi-helpful cuz at 40 we only see big blocks mostly that are barely affected by comments. Further, I added these flags to avoid dummy errors/violations.  ` --no-fail-on-violation \--no-fail-on-error` so the test will fail iff pmd.yml itself is messed up - syntax wise.

Current implementation
``` 
                  sed -E '
                    :a
                    /&$/ {
                          N
                          s/&[[:space:]]*\n[[:space:]]*&/ /
                          ba
                    }
                    s/&//g
                  ' "$file" | \
                  sed -E \
                      -e '/^\s*!/s/.*//' \
                      -e '/^[cC*dD]/s/.*//' \
                      -e '/^[ \t]*[cC*dD]/s/.*//' \
                      -e 's/([^"'\'']*("[^"]*"[^"'\'']*|'\''[^'\'']*'\''[^"'\'']*)*[^"'\'']*)[!].*$/\1/' \
                      > "$TMP_FILE"
```

Malmahrouqi3: > (i haven't checked to see if this works myself). PS this will mess up the line numbers but whatever.

You can just copy and paste the redundant lines and search them up. It ain't gonna be a wild hunting job.

sbryngelson:
something isn't right with concatenating the line continuations because this is a repeated pattern in pmd:

```fortran
                                q_sf(j, k, l) = q_sf(j, k, l) 
                                                + q_prim_vf0(mom_idx%beg)%sf(j, k, l)*fd_coeff_x(r, j)* 
                                                q_prim_vf0(mom_idx%beg)%sf(r + j, k, l) 
                                                + q_prim_vf0(mom_idx%beg + 1)%sf(j, k, l)*fd_coeff_y(r, k)* 
                                                q_prim_vf0(mom_idx%beg)%sf(j, r + k, l)
```

if they were concatted then this would be on one line

Malmahrouqi3:
> something isn't right with concatenating the line continuations because this is a repeated pattern in pmd:
> 
> ```fortran
>                                 q_sf(j, k, l) = q_sf(j, k, l) 
>                                                 + q_prim_vf0(mom_idx%beg)%sf(j, k, l)*fd_coeff_x(r, j)* 
>                                                 q_prim_vf0(mom_idx%beg)%sf(r + j, k, l) 
>                                                 + q_prim_vf0(mom_idx%beg + 1)%sf(j, k, l)*fd_coeff_y(r, k)* 
>                                                 q_prim_vf0(mom_idx%beg)%sf(j, r + k, l)
> ```
> 
> if they were concatted then this would be on one line

I guess this should fix it
```fortran
                                    q_sf(j, k, l) = q_sf(j, k, l) &
                                                    + q_prim_vf0(mom_idx%beg)%sf(j, k, l)*fd_coeff_x(r, j)* &
                                                    q_prim_vf0(mom_idx%beg)%sf(r + j, k, l) &
                                                    + q_prim_vf0(mom_idx%beg + 1)%sf(j, k, l)*fd_coeff_y(r, k)* &
                                                    q_prim_vf0(mom_idx%beg)%sf(j, r + k, l) &
                                                    + q_prim_vf0(mom_idx%end)%sf(j, k, l)*fd_coeff_z(r, l)* &
                                                    q_prim_vf0(mom_idx%beg)%sf(j, k, r + l)/y_cc(k)
```

New Implementation

Malmahrouqi3:
I guess it should now presumably detect more duplicate lines if they are exactly the same operation but split out into lines differently.

```fortran

q_sf(j, k, l) = q_sf(j, k, l)+q_prim_vf0(mom_idx%beg)%sf(j, k, l)*fd_coeff_x(r, j)*q_prim_vf0(mom_idx%beg)%sf(r+j, k, l)+q_prim_vf0(mom_idx%beg+1)%sf(j, k, l)*fd_coeff_y(r, k)*q_prim_vf0(mom_idx%beg)%sf(j, r+k, l)+q_prim_vf0(mom_idx%end)%sf(j, k, l)*fd_coeff_z(r, l)*q_prim_vf0(mom_idx%beg)%sf(j, k, r+l)/y_cc(k)
```

sbryngelson:
yes that's what i was thinking. you could even strip spaces (maybe?)

sbryngelson:
this is actually a very valuable tool!

Malmahrouqi3:
I left behind `=  .or. .and.` and few subtle things.
Other than that, all spaces should be taken off around math/comparison operators, inside indexing parentheses and brackets.

sbryngelson:
> I left behind `= .or. .and.` and few subtle things. Other than that, all spaces should be taken off around math/comparison operators, inside indexing parentheses and brackets.

yes agreed. so you already did this or not yet?

Malmahrouqi3:
yup, you can check out the last commit PMD check.

Malmahrouqi3:
Filter Full Implementation
``` bash
                  sed -E '
                    # First handle & continuation style (modern Fortran)
                    :ampersand_loop
                    /&[[:space:]]*$/ {
                      N
                      s/&[[:space:]]*\n[[:space:]]*(&)?/ /g
                      tampersand_loop
                    }

                    # Handle fixed-form continuation (column 6 indicator)
                    :fixed_form_loop
                    /^[[:space:]]{0,5}[^[:space:]!&]/ {
                      N
                      s/\n[[:space:]]{5}[^[:space:]]/ /g
                      tfixed_form_loop
                    }

                    # Remove any remaining continuation markers
                    s/&//g

                    # Normalize spacing - replace multiple spaces with single space
                    s/[[:space:]]{2,}/ /g

                    # Remove spaces around mathematical operators
                    s/[[:space:]]*\*[[:space:]]*/*/g
                    s/[[:space:]]*\+[[:space:]]*/+/g
                    s/[[:space:]]*-[[:space:]]*/-/g
                    s/[[:space:]]*\/[[:space:]]*/\//g
                    s/[[:space:]]*\*\*[[:space:]]*/\*\*/g

                    # Remove spaces in common Fortran constructs (array indexing, function calls)
                    s/\([[:space:]]*([^,)[:space:]]+)[[:space:]]*,/(\1,/g      # First argument
                    s/,[[:space:]]*([^,)[:space:]]+)[[:space:]]*,/,\1,/g       # Middle arguments
                    s/,[[:space:]]*([^,)[:space:]]+)[[:space:]]*\)/,\1)/g      # Last argument
                    s/\([[:space:]]*([^,)[:space:]]+)[[:space:]]*\)/(\1)/g     # Single argument

                    # Remove spaces around brackets and parentheses
                    s/\[[[:space:]]*/</g
                    s/\[[[:space:]]*/>/g
                    s/\[[[:space:]]*/</g
                    s/[[:space:]]*\]/]/g
                    s/\([[:space:]]*/(/g
                    s/[[:space:]]*\)/)/g

                    # Remove spaces around comparison operators
                    s/[[:space:]]*<=[[:space:]]*/</g
                    s/[[:space:]]*>=[[:space:]]*/>/g
                    s/[[:space:]]*<[[:space:]]*/</g
                    s/[[:space:]]*>[[:space:]]*/>/g
                    s/[[:space:]]*==[[:space:]]*/==/g

                    # Remove full-line comments
                    /^\s*!/d
                    /^[cC*dD]/d
                    /^[ \t]*[cC*dD]/d

                    # Remove end-of-line comments, preserving quoted strings
                    s/([^"'\''\\]*("[^"]*")?('\''[^'\'']*'\''?)?[^"'\''\\]*)[!].*$/\1/
                  ' "$file" > "$TMP_FILE"
```

sbryngelson:
i just realized that once you have removed any line continuations, you can delete all blank lines and spaces in the source code... they don't actually mean anything and the code doesn't have to compile, it just has to be parsed by PMD. by doing this we ensure we find all duplicate patterns

Malmahrouqi3:
Yup yup, it is kinda reasonable to think this way if you would rather unveil longer patterns than just ones with few lines.

sbryngelson:
> Yup yup, it is kinda reasonable to think this way if you would rather unveil longer patterns than just ones with few lines.

cool should be easier to do this as well. don't see a new commit yet but will look at the PR once it's updated. I _did_ notice that the current run appears to have shorter "worst" offenders at the top of PMD output (fewer tokens). is this because you stripped their whitespace / comments and the code is the same as it was before (so # tokens dropped but the Fortran part is the same)? or is it not finding the worst offenders it was a few commits ago? I remember seeing > 200 tokens as the worst cases

Edit: I checked myself. The violations (the first few worst ones) are the same, so you didn't break anything 👍 .

Malmahrouqi3:
Added to remove blank lines
 /^[[:space:]]*$/d

sbryngelson:
Looks pretty good to me. Do you have any other modifications planned?

Malmahrouqi3:
Nope

Issue #614, opened 2024-09-07
New modules for Bridges2 GPU


sbryngelson:
It seems the previous modules for Bridges2 are no longer working. I found these to work:
```
  1) openmpi/4.0.5-nvhpc22.9   2) python/3.8.6   3) cuda/11.7.1
```
along with an
```
export FC=nvfortran
export CC=nvc
export CXX=nvc++
```

sbryngelson:
Update: Related to #613 . I found that the above works when turning `            set(NVHPC_USE_TWO_PASS_IPO TRUE)` to `            set(NVHPC_USE_TWO_PASS_IPO FALSE)` in `CMakeLists.txt`, but does __not__ work if twopassIPO is TRUE. This serves as a confirmation that _at least_ nvhpc 23.X is required for two-pass IPO.

sbryngelson:
Briges2 does not have nvhpc23.X so I cannot test with these

Malmahrouqi3:
Current modules on `mfc.sh load`
``` txt
b     PSC Bridges2
b-all python/3.8.6
b-cpu allocations/1.0 gcc/10.2.0 openmpi/4.0.5-gcc10.2.0
b-gpu openmpi/4.0.5-nvhpc22.9 nvhpc/22.9 cuda
b-gpu CC=nvc CXX=nvc++ FC=nvfortran
```

Newest modules on Bridges2
``` txt
b     PSC Bridges2
b-all python/3.8.6
b-cpu allocations/1.0 gcc/10.2.0 openmpi/4.0.5-gcc10.2.0
b-gpu openmpi/4.0.5-nvhpc22.9 nvhpc/22.9 cuda/12.4.0
b-gpu CC=nvc CXX=nvc++ FC=nvfortran
```

Quick note: if outdated python version error ever occurs on Bridges2, quick way fix is to load anaconda3. It always works for me somehow.
``` bash
[almahrou@bridges2-login012 MFC]$ ./mfc.sh build
mfc: ERROR > Python 3.8.6 (python3) is out of date. Required >= 3.9.
```

sbryngelson:
so you can only build if you load anaconda? if so, we should add it to the bridges2 modules

Malmahrouqi3:
kinda I am still tackling an issue with environment variables and yeah below is a version of `mfc.sh` that is supposedly working I am still figuring out things right now.

``` bash
#!/bin/bash
# Custom MFC build script that uses Python 3.11

[...]

PR #905, opened 2025-06-27
Updated Bridges2 Modules (CPU/GPU)

Malmahrouqi3:
## Description
Concerning (#614),

Updated modules for Bridges2, tested on `--qos=cpuinteract/gpuinteract`. MFC compiles and builds neatly. To address the outdated python version issue, `anaconda3 `is loaded as it houses a very recent version. Foreseen concern would be dependency conflict with the co-existing py venv & conda env. Therefore, please keep conda as-is and do not attempt to install any dependency on it at all. To run tests, append `-- --binary mpirun`.

### DIFF
--- a/toolchain/modules
+++ b/toolchain/modules
@@ -14,11 +14,9 @@ s-gpu nvhpc/22.11 cuda/nvhpc
 s-gpu CC=nvc CXX=nvc++ FC=nvfortran
 
 b     PSC Bridges2
-b-all python/3.8.6 hdf5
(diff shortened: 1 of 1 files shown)


sbryngelson:
Why this part
>  To run tests, append -- --binary mpirun.

it should run tests with `-- -c bridges2` (this is the MFC way, at least)

Malmahrouqi3:
I edited the description.

**Edit**: just confirming, either way works fine.

sbryngelson (approved):

Issue #650, opened 2024-10-15
Valgrind `verrou` to check for sketchy float operations


sbryngelson:
Use Valgrind's `verrou` tool to check for poorly conditioned float operations. 

Link: https://edf-hpc.github.io/verrou/vr-manual.html 

These are likely the source of why some tests randomly fail when you compile them differently.

Related to #622 and https://github.com/MFlowCode/MFC/pull/525 and #474

sbryngelson:
I seem to, perhaps, have found something useful. Hard to believe, I know.

This is the output from running the 1D sod shocktube example for just 10 time steps. We see there is already some weird stuff going on. This is with `--no-mpi` on a debug build with GCC right off the master branch. @henryleberre @okBrian you might find this interesting.

```console
instinct: examples/1D_sodshocktube $ valgrind -s --tool=verrou --rounding-mode=random /nethome/sbryngelson3/MFC/build/staging/90baab9f56/simulation  > valgrind.txt
==1409376== Verrou, Check floating-point rounding errors
==1409376== Copyright (C) 2014-2019, EDF (F. Fevotte & B. Lathuiliere). 2019-2023, EDF (B. Lathuiliere). 2020-2021, TriScale innov (F. Fevotte)
==1409376==
==1409376== Using Valgrind-3.23.0.verrou-2.6.0 and LibVEX; rerun with -h for copyright info
==1409376== Command: /nethome/sbryngelson3/MFC/build/staging/90baab9f56/simulation
==1409376==
==1409376== First seed : 2204177
==1409376== Backend verrou : 1.x-dev
==1409376== Backend mcaquad : 1.x-dev
==1409376== Backend checkcancellation : 1.x-dev
==1409376== Backend check_float_max : 1.x-dev
==1409376== Backend checkdenorm : 1.x-dev
==1409376== Instrumented operations :
==1409376== 	add : yes
==1409376== 	sub : yes
==1409376== 	mul : yes
==1409376== 	div : yes
==1409376== 	mAdd : yes
==1409376== 	mSub : yes
==1409376== 	sqrt : yes
==1409376== 	cmp : no
==1409376== 	conv : yes
==1409376== 	max : no
==1409376== 	min : no
==1409376== Instrumented vectorized operations :
==1409376== 	scal : no
==1409376== 	llo  : yes
==1409376== 	vec2  : yes
==1409376== 	vec4  : yes
==1409376== 	vec8  : yes
==1409376== 	unk  : yes
[...]

PR #937, opened 2025-07-12
`verrou` into Workflow

Malmahrouqi3:
## Description
Concerning (#650),
Intended as standalone floating-point error checker.

Debugging info,
The current Valgrind setup ought to be sufficient, so debugging locally (nektos/act extension) or on the CI are all viable.

To-dos,
- [ ] Proper Valgrind commands to generate the most amount of errors.
- [ ] Save and process all reports such that non-empty error summaries are displayed.

Ref [verrou docs](https://edf-hpc.github.io/verrou/vr-manual.html)

### DIFF
--- a/.github/workflows/verrou.yml
+++ b/.github/workflows/verrou.yml
@@ -1,0 +1,31 @@
+name: Floating-Point Compliance
+
+on: [pull_request, push]
+
+jobs:
+  self:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone PR
+        uses: actions/checkout@v4
+
+      - name: valgrind Setup
+        run: |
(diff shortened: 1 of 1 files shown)


Malmahrouqi3:
**Status Update**: Valgrind builds nicely and PR is awaiting (#938) to merge in to replicate the results of (#650).

sbryngelson:
I assume you know this but the use of verrou in our case is very specific (we are looking for quite specific problems involving dangerous floating point operations that could cause round-off error problems). This is what this 'linter' of sorts should be looking for.

sbryngelson:
this doesn't do what the GH issue discussed and has been stagnant. closing.

PR #942, opened 2025-07-14
Add hard coded patched for tests

danieljvickers:
## Description

The primary intent of this PR is to remove hardcoded patches from the test suite, preventing the need to recompile them and saving time during testing. In doing so, I also refactored the geometries because we had no existing support for creating hard-coded patches on any geometries other than line segments, rectangles, and cuboids. Now we can generate hard-coded patches on almost any geometries except for the airfoil geometries (I left a TODO about this) and the model (STL) geometry (this will require it's own dedicated refactor). I also changed every test that used analytic patches to use new hard coded patches, added comments to some code in the toolchain, and updated the docs to note that the old analytic geometry patches are deprecated. I left a print statement warning that those geometries are deprecated, and we should delete them after some time. For now, I had then call subroutines that recreated their original functionality.

TODO :: After this PR, we should go back and add the analytic patches as examples that will be skipped so that they are present for external reference.

### Type of change

- [ ] Something else

### Scope

## How Has This Been Tested?

- [ ] This was tested on the full test suite. (`./mfc.sh test -j $(nproc)`)

**Test Configuration**:

All of my testing was done locally on Ubuntu with the intel and non-intel compilers.

## Checklist


docs/documentation/case.md
```diff
@@ -996,16 +997,16 @@ This boundary condition can be used for subsonic inflow (`bc_[x,y,z]%[beg,end]`
...
-| 7    | 2D Hardcoded 	    | 2     | N      | Assigns the primitive variables as analytical functions. |
+| 6    | N/A 		        | N/A   | N/A    | No longer exists. Empty. |
+| 7    | N/A        	    | N/A   | N/A    | No longer exists. Empty. |
 | 8    | Sphere 		    | 3     | Y      | Requires `[x,y,z]_centroid` and `radius` |
 | 9    | Cuboid 		    | 3     | N      | Coordinate-aligned. Requires `[x,y,z]_centroid` and `length_[x,y,z]`. |
 | 10   | Cylinder 		    | 3     | Y      | Requires `[x,y,z]_centroid`, `radius`, and `length_[x,y,z]`. |
 | 11   | Sweep plane 	    | 3     | Y      | Not coordinate-aligned. Requires `x[y,z]_centroid` and `normal(i)`. |
 | 12   | Ellipsoid 		    | 3     | Y      | Requires `[x,y,z]_centroid` and `radii(i)`. |
-| 13   | 3D Hardcoded 	    | 3     | N      | Assigns the primitive variables as analytical functions |
+| 13   | N/A        	    | N/A   | N/A    | No longer exists. Empty. |
```
sbryngelson, line 1007: so these really aren't used? @wilfonba

src/pre_process/m_patches.fpp
```diff
@@ -1290,62 +1359,15 @@ contains
...
-                                                        eta, q_prim_vf, patch_id_fp)
-
-                @:Hardcoded1D()
-
-                ! Updating the patch identities bookkeeping variable
-                if (1._wp - eta < 1.e-16_wp) patch_id_fp(i, 0, 0) = patch_id
-
-            end if
-        end do
+        print *, "WARNING :: The use of the 1D analytical patch geometry has &
```
sbryngelson: you don't do this in MFC, you add to `m_checker` for this case parameter with `@:PROHIBIT` (with a message if you want) and then it never reaches this subroutine and you can delete it.

src/pre_process/m_patches.fpp
```diff
@@ -1410,70 +1438,15 @@ contains
         !! @param patch_id_fp Array to track patch ids
         !! @param q_prim_vf Array of primitive variables
     subroutine s_2D_analytical(patch_id, patch_id_fp, q_prim_vf)
-
         integer, intent(in) :: patch_id
         integer, dimension(0:m, 0:n, 0:p), intent(inout) :: patch_id_fp
         type(scalar_field), dimension(1:sys_size), intent(inout) :: q_prim_vf
 
-        integer :: i, j, k !< generic loop iterators
+        print *, "WARNING :: The use of the 2D analytical patch geometry has &
```
sbryngelson: oh i see, you are going to make this for actual analytic semantics. but what if the 2D analytical patch isn't a rectangle (say, it could be a circle)? I guess you could put that as a parameter to the subroutine/input file.

sbryngelson: btw all print statements need to have ` if (proc_rank == 0 )` before them so they don't print out 24 (or, even worse, 30K) lines that say the same thing, one for each MPI rank.

src/pre_process/m_patches.fpp
```diff
@@ -2037,6 +1959,9 @@ contains
                                                                     eta, q_prim_vf, patch_id_fp)
 
                             @:analytical()
+                            if (patch_icpp(patch_id)%hcid /= dflt_int) then
+                                @:Hardcoded3D()
+                            end if
 
                             ! Updating the patch identities bookkeeping variable
                             if (1._wp - eta < 1.e-16_wp) patch_id_fp(i, j, k) = patch_id
```
sbryngelson: while we are at it can we remove this magic number 1e-16 and use something in `common/m_constants`

examples/3D_TaylorGreenVortex/case.py
```diff
@@ -85,10 +85,11 @@
             "patch_icpp(1)%length_x": 2 * math.pi * L,
             "patch_icpp(1)%length_y": 2 * math.pi * L,
             "patch_icpp(1)%length_z": 2 * math.pi * L,
-            "patch_icpp(1)%vel(1)": f"{V0}*sin(x/{L})*cos(y/{L})*sin(z/{L})",
-            "patch_icpp(1)%vel(2)": f"-{V0}*cos(x/{L})*sin(y/{L})*sin(z/{L})",
+            "patch_icpp(1)%vel(1)": 0.0,
+            "patch_icpp(1)%vel(2)": 0.0,
```
sbryngelson, line 89: can we just remove these lines for unused variables from the input file? might need an if statement in the checker that if the variables are all set to default values, then we require hcid /= 0, else fail.

docs/documentation/case.md
sbryngelson: if so, then if someone puts this in their input file it needs to be added as a PROHIBIT statement in the pre_process/m_checker file

src/pre_process/m_patches.fpp
danieljvickers: Noted on the print statements.

I put the warning here and called rectangle because the (soon to be) legacy version of hard-coded patches could ONLY be run on rectangular geometries. Adding this print is just making sure that I don't break anyone's code that isn't checked into the repo, and instead giving them time to see the warning and let them fix it as they notice the warning. Later we should remove it in another PR.

src/pre_process/m_patches.fpp
danieljvickers: Noted. I will look into that. I did want there to be some time before we just prohibit to give people a chance to swap over, since writing in this call to the rectangle/line segment/cuboid subroutines means that old version of the case files will still work.

docs/documentation/case.md
danieljvickers: This PR deprecates their use. You now no longer need a specific geometry for hard-coded patches. Instead, you just request the geometry that you want to run your patch on. It just removes some code and clarifies the use of geometries with hard-coded patches.

examples/3D_TaylorGreenVortex/case.py
danieljvickers: No, I just tested this specific file to make sure, and you get failures if you do not have those values left in as zeros. We can add that, but I think we will want some notion about what variable are set with head `hcid` since I image that we will want to perform this check if that value actually needs to be set

docs/documentation/case.md
wilfonba: There was an example case that wasn't compatible with the existing way that hardcoded patches work. Dan and I talked about it, and the solution was to get rid of the line/rectangle/cuboid hardcoded patches (15/7/13) and allow all patches to have an `hcid` specified that references the code in `XD_hardcoded.fpp`. This adds functionality by making it so that the variables for any patch geometry can be hardcoded, not just the line/rectangle/cuboid.

examples/3D_TaylorGreenVortex/case.py
wilfonba: The variable in the case file serves as a "default value" if the hardcoded patch doesn't overwrite it. This makes it so that you can hardcode only one or two variables and leave the rest defined by the default in the case file. I'm not sure there would be an easy way to check for this in the checker code, besides just requiring "defaults" to be set for all variables.

src/pre_process/m_patches.fpp
wilfonba: @danieljvickers there's an existing constant in `m_constants.fpp` called `sgm_eps` that can be used here.

examples/3D_TaylorGreenVortex/case.py
sbryngelson: yeah. agreed. leave it like this for now and it can be manicured later.

docs/documentation/case.md
sbryngelson: check. thanks.

src/pre_process/m_patches.fpp
```diff
@@ -253,13 +244,19 @@ contains
...
+        ! Placeholders for the cell boundary values
+        real(wp) :: pi_inf, gamma, lit_gamma
+        @:HardcodedDimensionsExtrusion()
+        @:Hardcoded1DVariables()
 
         pi_inf = fluid_pp(1)%pi_inf
         gamma = fluid_pp(1)%gamma
         lit_gamma = (1._wp + gamma)/gamma
+        j = 0.0_wp
+        k = 0.0_wp
```
anandrdbz, line 260: typo, j = 0 (0.0_wp is for reals)

src/pre_process/m_patches.fpp
```diff
@@ -253,13 +244,19 @@ contains
...
-        integer :: i, j, k !< Generic loop operators
+        ! Placeholders for the cell boundary values
+        real(wp) :: pi_inf, gamma, lit_gamma
+        @:HardcodedDimensionsExtrusion()
+        @:Hardcoded1DVariables()
 
         pi_inf = fluid_pp(1)%pi_inf
         gamma = fluid_pp(1)%gamma
         lit_gamma = (1._wp + gamma)/gamma
+        j = 0.0_wp
```
danieljvickers:
```suggestion
        j = 0
```

src/pre_process/m_patches.fpp
```diff
@@ -253,13 +244,19 @@ contains
...
+        k = 0.0_wp
```
danieljvickers:
```suggestion
        k = 0
```

src/pre_process/m_check_patches.fpp
```diff
@@ -74,11 +76,15 @@ contains
                 elseif (patch_icpp(i)%geometry == 12) then
                     call s_check_ellipsoid_patch_geometry(i)
                 elseif (patch_icpp(i)%geometry == 13) then
-                    call s_check_3D_analytical_patch_geometry(i)
+                    call s_mpi_abort('geometry 13 (formerly "3D Hardcoded")'// &
```
sbryngelson: we have @:prohibit and prohibit_abort for this

src/pre_process/m_check_patches.fpp
```diff
@@ -74,11 +76,15 @@ contains
...
+                                     '. Exiting.')
                 elseif (patch_icpp(i)%geometry == 14) then
                     call s_check_spherical_harmonic_patch_geometry(i)
                 elseif (patch_icpp(i)%geometry == 15) then
-                    call s_check_1d_analytical_patch_geometry(i)
+                    call s_mpi_abort('geometry 15 (formerly "1D Hardcoded")'// &
+                                     'is no longer supported for patch '//trim(iStr)// &
+                                     '. Exiting.')
                 elseif (patch_icpp(i)%geometry == 16) then
                     print *, '1d pressure sinusoid'
```
sbryngelson: i think we can remove all of these elseifs that just print something. i added them years ago and i don't think they are helping anyone.

src/pre_process/m_patches.fpp
```diff
@@ -437,13 +448,19 @@ contains
...
+
                     end if
                 end if
             end do
         end do
+        @:HardcodedDellacation()
 
     end subroutine s_circle
 
+    ! TODO :: Determine if we want analytical and hardcoded patches for the airfoil geometries
```
sbryngelson: we don't, can remove this

sbryngelson:
This seems fine to me, pending removing some comments / extra junk. Also, if @wilfonba approves, since he uses a lot of these features and added some of them to the code to begin with.

docs/documentation/case.md
```diff
@@ -313,7 +314,7 @@ These parameters should be prepended with `patch_ib(j)%` where $j$ is the patch
 #### Parameter Descriptions
 
 - `geometry` defines the type of geometry of a patch with an integer number.
-Definitions for currently implemented patch types are list in table [Immersed Boundary Patch Type](#immersed-boundary-patch-types)
+Definitions for currently implemented patch types are list in table [Patch Types](#patch-types).
```
wilfonba: This should be reverted unless I'm missing something. The "immersed boundary patch types" table appears to still exist.

src/pre_process/include/2dHardcodedIC.fpp
```diff
@@ -161,6 +160,37 @@
...
 
+    case (280)
+        ! This is patch is hard-coded for test suite optimization used in the
+        ! 2D_isentropicvortex case:
+        ! This analytic patch uses geometry 2
+        if (patch_id == 1) then
+            q_prim_vf(E_idx)%sf(i, j, 0) = 1.0*(1.0 - (1.0/1.0)*(5.0/(2.0*3.141592653589793))*(5.0/(8.0*1.0*(1.4 + 1.0)*3.141592653589793))*exp(2.0*1.0*(1.0 - (x_cc(i) - patch_icpp(1)%x_centroid)**2.0 - (y_cc(j) - patch_icpp(1)%y_centroid)**2.0)))**(1.4 + 1.0)
+            q_prim_vf(contxb + 0)%sf(i, j, 0) = 1.0*(1.0 - (1.0/1.0)*(5.0/(2.0*3.141592653589793))*(5.0/(8.0*1.0*(1.4 + 1.0)*3.141592653589793))*exp(2.0*1.0*(1.0 - (x_cc(i) - patch_icpp(1)%x_centroid)**2.0 - (y_cc(j) - patch_icpp(1)%y_centroid)**2.0)))**1.4
+            q_prim_vf(momxb + 0)%sf(i, j, 0) = 0.0 + (y_cc(j) - patch_icpp(1)%y_centroid)*(5.0/(2.0*3.141592653589793))*exp(1.0*(1.0 - (x_cc(i) - patch_icpp(1)%x_centroid)**2.0 - (y_cc(j) - patch_icpp(1)%y_centroid)**2.0))
+            q_prim_vf(momxb + 1)%sf(i, j, 0) = 0.0 - (x_cc(i) - patch_icpp(1)%x_centroid)*(5.0/(2.0*3.141592653589793))*exp(1.0*(1.0 - (x_cc(i) - patch_icpp(1)%x_centroid)**2.0 - (y_cc(j) - patch_icpp(1)%y_centroid)**2.0))
```
wilfonba: There's a constant for `pi` in `m_constants` that will clean these lines up a lot

src/pre_process/include/1dHardcodedIC.fpp
```diff
@@ -3,12 +3,23 @@
...
+        ! This is patch is hard-coded for test suite optimization used in the
+        ! 1D_shuoser cases: "patch_icpp(2)%alpha_rho(1)": "1 + 0.2*sin(5*x)"
+        if (patch_id == 2) then
+            q_prim_vf(contxb + 0)%sf(i, 0, 0) = 1 + 0.2*sin(5*x_cc(i))
+        end if
+
+    case (181)
+        ! This is patch is hard-coded for test suite optimization used in the
+        ! 1D_titarevtorro cases: "patch_icpp(2)%alpha_rho(1)": "1 + 0.1*sin(20*x*pi)"
+        q_prim_vf(contxb + 0)%sf(i, 0, 0) = 1 + 0.1*sin(20*x_cc(i)*3.141592653589793)
```
wilfonba: There's a constant for 'pi' in 'm_constants' that will clean these lines up a lot

wilfonba (changes requested):

wilfonba:
I added some minor comments. I don't see any other issues with these changes

src/pre_process/m_check_patches.fpp
danieljvickers: Thanks. I need to look at the code for the prohibit and s_prohibit_abort subroutines so that I am sure when to be using what, but I can use this for now.

sbryngelson: i must admit i don't fully understand why we have both the probit macro and the prohibit abort subroutine. it would be useful to see where they are used to see why both exist, then pick the appropriate one. it's possible that one or the other should just be deleted.

docs/documentation/case.md
danieljvickers: The text sounds like it should be the table for the patch types, now the immersed bounadry types. To compromise, I will put in both.

sbryngelson: @wilfonba is right that the IBM table exists, but i'm a little confused why we linted to it here? don't we have a patch types table and that's the relevant thing to link to (presumably @danieljvickers's fix here)? [admittedly i did not look at the documentatino before posting this comment]

sbryngelson: @danieljvickers beat me to the punch by 2 minutes. Is the IBM table linked to anywhere else in the docs? presumably it should be linked to somewhere....

src/pre_process/include/2dHardcodedIC.fpp
sbryngelson: correct

src/pre_process/include/1dHardcodedIC.fpp
sbryngelson: correct

docs/documentation/case.md
danieljvickers: @sbryngelson I just decided to list both here instead. Hopefully this is a good compromise. I have my eyes on like a 3-hour branch where I only change some of the things that I noticed are missing/misplaced in the docs.

sbryngelson: very good, thanks @danieljvickers . resolving this conversation.

sbryngelson:
@danieljvickers would you consider this ready to merge?

PR #990, opened 2025-08-31
MPI Transfer for IBM Markers on multiple GPUs

anandrdbz:
## Description

Fix a Bug with MPI transfer of IBM Markers on multiple GPUs

Fixes #(issue) [optional]

### Type of change

- [x ] Bug fix (non-breaking change which fixes an issue)
- [ ] Something else

### Scope

- [ x] This PR comprises a set of related changes with a common goal

## How Has This Been Tested?

- [ ] Test A
- [ ] Test B

**Test Configuration**:

## Checklist

### DIFF
--- a/src/simulation/m_mpi_proxy.fpp
+++ b/src/simulation/m_mpi_proxy.fpp
@@ -345,12 +345,40 @@ contains
         #:endfor
         call nvtxEndRange ! Packbuf
 
-        call nvtxStartRange("IB-MARKER-SENDRECV")
-        call MPI_SENDRECV( &
-            ib_buff_send, buffer_count, MPI_INTEGER, dst_proc, send_tag, &
-            ib_buff_recv, buffer_count, MPI_INTEGER, src_proc, recv_tag, &
-            MPI_COMM_WORLD, MPI_STATUS_IGNORE, ierr)
-        call nvtxEndRange ! RHS-MPI-SENDRECV-(NO)-RDMA
+        #:for rdma_mpi in [False, True]
+            if (rdma_mpi .eqv. ${'.true.' if rdma_mpi else '.false.'}$) then
+                #:if rdma_mpi
+                    #:call GPU_HOST_DATA(use_device='[ib_buff_send, ib_buff_recv]')
+                        call nvtxStartRange("IB-MARKER-SENDRECV-RDMA")
+
+                        call MPI_SENDRECV( &
+                            ib_buff_send, buffer_count, MPI_INTEGER, dst_proc, send_tag, &
+                            ib_buff_recv, buffer_count, MPI_INTEGER, src_proc, recv_tag, &
+                            MPI_COMM_WORLD, MPI_STATUS_IGNORE, ierr)
+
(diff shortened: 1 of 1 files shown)


anandrdbz:
@sbryngelson MacOS CI has a problem building cmake btw

sbryngelson:
> @sbryngelson MacOS CI has a problem building cmake btw

Please fix

sbryngelson (approved):

sbryngelson:
does doesn't run with case optimization on CPU/Phoenix for some reason... @anandrdbz please check logs

anandrdbz:
Actually this case revealed an important corner case, the IBM boundary and processor boundary match almost exactly. You actually need a larger buffer size than I even thought , I kept a limit of 6 = 2*gp_layers for the image point, but this isn't strictly true. In this particular case, it worked when increasing it to 8. 

But there's actually no deterministic way of calculating this though as it depends on the IBM geometry, For now I'll push a fix to get this test to pass, but I'll put a print message that tells the user to increase buff_size if they run into this issue again in the source code.

sbryngelson:
hmm... @anandrdbz given that halo exchanges / ghost cell exchanges are relatively cheap except in extreme strong scaling cases, can we just put a rather generous upper bound on the ghost cells? This is a good find though, and I don't want to lose it in case it comes up again. I'm not sure how you want to add an warning message, but try to place it carefully/mindfully. Presumably the vast majority of cases with an IB will fail for a reason that has nothing to do with this 😄

anandrdbz:
Yeah this will fail uniquely, with a print message that's unique too, so it should be clear ( you can checkout my latest commit)

 I have kept a generous upper bound of 10 for IBM cases, but there's no way to calculate a strict upper bound as it's geometry dependent.

src/common/m_helper_basic.fpp
```diff
@@ -142,6 +143,10 @@ contains
             buff_size = max(buff_size, 6)
         end if
 
+        if (ib) then
+            buff_size = max(buff_size, 10)
```
sbryngelson, line 147: Does this depend at all on the grid resolution, the size of the IBM compared to the grid resolution, or if an STL file is used? If so, we could use that as a heuristic for how large the lower bound should be.

anandrdbz: Yes, but it also depends on the curvature of the IBM. Essentially to determine ghost points, you are checking if there's non-IBM points within a rectangle of size (6dx, 6dy) around the point (3dx in each direction).  This will both depend on the grid resolution as well as the geometry. And if this is true, then you will need to ensure that the corresponding image point is within buff_size of the local domain. 

So, the limit for a circle and airfoil for example will be quite different even at the same grid resolution.

anandrdbz:
Also @sbryngelson the benchmark case with IBM will eventually fail the check in this latest commit due to the increase in buffer size from 6 to 10 (I checked this for CPUs on Phoenix), with a 20% or so slowdown in grind time (I think the limit is about 5-10%).

sbryngelson (approved):

PR #1191, opened 2026-02-21
Fix s_write_intf_data_file using wrong loop index for distance check

### LINKED ISSUE #1211: Add Fortran/Fypp static analysis linter to catch copy-paste bugs
## Feature Request

Add a Python-based static analysis linter for Fortran/Fypp source code that catches patterns known to cause silent bugs. Integrate it into the pre-commit hook (\`./mfc.sh precheck\`) and CI (\`lint-source.yml\`).

### Proposed Checks

| Check | What it catches |
|---|---|
| **Fypp list duplicates** | Duplicate entries in \`#:for VAR in [...]\` broadcast/loop lists |
| **Duplicate consecutive lines** | Identical adjacent non-trivial lines (copy-paste errors) |
| **Hardcoded byte size** | \`int(8._wp, ...)\` patterns that assume 8-byte reals |

These checks are motivated by real bugs found in the codebase. The linter would prevent regressions after the bugs are fixed.

sbryngelson:
## Summary

**Severity:** HIGH — interface contour extraction uses stale loop index, produces wrong point distances.

**File:** `src/post_process/m_data_output.fpp`, lines 1549-1563

In `s_write_intf_data_file`, the Euclidean distance `euc_d` between a candidate interface point and existing stored points is computed using `x_d1(i)` and `y_d1(i)`, where `i` is stale from a previous outer loop (value = `m+1` after the maxalpha search loop). This references out-of-bounds or wrong data. Additionally, `euc_d` is computed once outside the inner loop, so the distance check is the same for all stored points.

### Before
```fortran
! euc_d uses stale 'i' from the previous loop, not the stored-point index
euc_d = sqrt((x_cc(j) - x_d1(i))**2 + (y_cc(k) - y_d1(i))**2)
tgp = sqrt(dx(j)**2 + dy(k)**2)
do i = 1, counter        ! i is now a different loop variable
    if (euc_d < tgp) then
        cycle             ! wrong: should exit, not cycle
    elseif (euc_d > tgp .and. i == counter) then
        ! add point
```

### After
```fortran
tgp = sqrt(dx(j)**2 + dy(k)**2)
do i = 1, counter
    euc_d = sqrt((x_cc(j) - x_d1(i))**2 + (y_cc(k) - y_d1(i))**2)
    if (euc_d < tgp) then
        exit              ! too close to existing point, reject candidate
    elseif (i == counter) then
        ! add point — passed all distance checks
```

Three fixes:
1. Moved `euc_d` computation inside the loop so it checks distance to each stored point
2. Changed `cycle` to `exit` for correct early termination (reject if too close to **any** point)
3. Removed dead code in the `counter == 0` branch (distance to self)

## Test plan
- [ ] Run post-processing with interface data output
- [ ] Verify extracted contour points are properly spaced

Fixes #1211

### DIFF
--- a/src/post_process/m_data_output.fpp
+++ b/src/post_process/m_data_output.fpp
@@ -1546,19 +1546,16 @@ contains
                         counter = counter + 1
                         x_d1(counter) = x_cc(j)
                         y_d1(counter) = y_cc(k)
-                        euc_d = sqrt((x_cc(j) - x_d1(i))**2 + (y_cc(k) - y_d1(i))**2)
-                        tgp = sqrt(dx(j)**2 + dy(k)**2)
                     else
-                        euc_d = sqrt((x_cc(j) - x_d1(i))**2 + (y_cc(k) - y_d1(i))**2)
                         tgp = sqrt(dx(j)**2 + dy(k)**2)
                         do i = 1, counter
+                            euc_d = sqrt((x_cc(j) - x_d1(i))**2 + (y_cc(k) - y_d1(i))**2)
                             if (euc_d < tgp) then
-                                cycle
-                            elseif (euc_d > tgp .and. i == counter) then
+                                exit
+                            elseif (i == counter) then
                                 counter = counter + 1
                                 x_d1(counter) = x_cc(j)
                                 y_d1(counter) = y_cc(k)
-
                             end if
                         end do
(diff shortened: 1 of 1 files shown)


sbryngelson:

PR #1290, opened 2026-03-03
Moving EL Bubbles with MPI Decomposition

wilfonba:
## Description

This PR adds support for moving Euler-Lagrange bubbles and various force models. Tracer particles that simply follow the background flow are implemented, as are Newton's 2nd Law particles that respond to body forces, pressure gradients, and drag. This PR is marked as a draft for now as there are several things that I need to complete before a proper merge can occur, but I wanted these changes to be visible to other contributors working on related features.

### Type of change

- [x] New feature

### 16 rank CPU versus 16 rank GPU
This test was ran using the `examples/3D_lagrange_bubblescreen` test case and compares the results for a 16 rank CPU simulation to a 16 rank GPU simulation. The visualization shows the following things:
1. The difference in pressure in the vertical plane, with darker colors corresponding to larger errors. There seems to be a reflection of some sort from the subsonic buffer outflow condition, and the acoustic source introduces a small amount of difference, though neither of these features was modified by this PR. (I intend to add a video with no bubbles in the domain to see if these artifacts still exist)
2. The difference in void fraction in the horizontal plane. The difference is 1e-38 everywhere in ParaView, which corresponds to bitwise identical results.
3. The bubbles for each simulation overlayed on top of each other. The bubbles from the CPU result are colored in blue while the bubbles from the GPU result are colored in white. The small bit of blue that is observed in the video is a result of the rendering spheres with a fixed number of facets.
4. The pressure field in the lower front quadrant as a surface render with darker colors indicating higher pressure.
5. The MPI domain decomposition as a black lattice.

https://github.com/user-attachments/assets/262a31aa-2843-4dd1-9a93-a4151da73073

### 2D Lagrange bubbles in crossflow

The following tests all use the `examples/2D_lagrange_in_crossflow` example case. The first crossflow velocity is 0.2. This means that regardless of the drag model, the EL bubbles should accelerate to a velocity of 0.2 and then maintain a constant velocity, with the acceleration dictated by the drag model. The drag models in increasing order of drag force are slip drag -> no-slip drag -> Levich drag, and the results below agree with that.

image

The following three images show comparisons between 1 GPU, 4 GPU, and 4 CPU simulations. The left-hand axis shows the quantity of interest, and the right-hand axis shows the error on a log scale. The errors are nearly identical to zero in all cases, so the error lines are rarely visible. The left column shows position, and the right column shows velocity. The five rows correspond to the five different bubbles in the initial condition.
#### Slip Drag
image
#### No-Slip Drag
image
#### Levich Drag
image

I also performed a simulation with "tracer particles" that match the carrier velocity instantly. The results below show the expected behavior.
image

### 3D Lagrange Bubblescreen
I also ran the `examples/3D_lagrange_bubblescreen` example case with different numbers of CPUs and GPUs to ensure the results match. The figure below shows that the maximum pressure difference at the center of the bubble screen is on the order of `10^-15`.
image

### 2D Lagrange rising bubble

The following test uses the `examples/2D_lagrange_rising_bubble` example case. Five bubbles start at rest in a domain subject to gravitational force in hydrostatic equilibrium. The bubbles begin moving upwards, but reach a terminal velocity due to drag. The following figure shows the results for 1 GPU, 4 GPUs, and 4 CPUs. The left-hand axis shows the bubble position, and the right-hand axis shows the difference between results. Errors do grow as large as 10^-13, though there does not appear to be an obvious divergence between solutions.
image

### 2D 1 rank vs. 4 rank
The following test uses the `examples/2D_moving_lag_bub` example case. The left frame of the video shows the pressure difference between the two simulations, which is identically zero at all time steps. The right pane shows the bubble cloud of the 1- and 4-rank simulations overlaid in black and white, with no obvious divergence visible. The pressure field is shown in the background of the right pane, and the domain decomposition is indicated by the vertical and horizontal lines.
https://github.com/user-attachments/assets/173e2ff5-b0a5-4305-9b61-aba2dc4b282a

### 3D 1 rank vs. 8 rank
The following tests use the `examples/3D_moving_lag_particles` example case (with some modifications for bubble dynamics and such). The first plot shows the results using particles (`bubble_model=0`) with identical integral quantities.
image

The next plot shows integral quantities and their relative errors when radial dynamics are enabled. Relative differences are small, though they creep above machine precision.
image

## Checklist

<strong>GPU changes</strong> (expand if you modified <code>src/simulation/</code>)

### Must do

- [x] **Add regression tests** for `vel_model > 0` paths (tracer and Newton's 2nd Law) in `toolchain/mfc/test/cases.py`. The `2D_moving_lag_bubs` and `3D_moving_lag_particles` examples use `adap_dt` with parameters tuned for their full grid size and hang when reduced to 25x25 test grids. Dedicated test cases with appropriate parameters are needed.
- [x] **Add user-facing documentation** in `docs/documentation/case.md` for new parameters (`vel_model`, `drag_model`, `pressure_force`, `gravity_force`, `write_void_evol`, `input_path`, `charNz`)

### Should verify

- [x] **`n_el_bubs_glb` is rank-0 only** after `MPI_REDUCE` — confirmed intentional (only rank 0 prints stats)
- [x] **`nReal = 7 + 16*2 + 10*lag_num_ts` magic number** in `s_initialize_particles_mpi` — this buffer size must stay in sync with the pack/unpack loops. Consider computing it from named constants or adding a comment mapping each count to its field

### Already fixed (by review commits)

**Correctness fixes:**
- [x] 4th-order Lagrange interpolation: L(1) denominator `xi(2)->xi(1)`, L(4) numerator `xi(4)->xi(5)` (`m_bubbles_EL_kernels.fpp`)
- [x] `case default` in `f_advance_step` removed — dereferenced absent optional `fVel`/`fPos` args when called from EE path with `adap_dt` (`m_bubbles.fpp`)
- [x] Restart write out-of-bounds: write loop now filters with `particle_in_domain_physical` to match allocation size (`m_bubbles_EL.fpp`)
- [x] Vapor/gas property swap fixed in both example cases — all 10 `bub_pp` thermodynamic params were inverted (`2D_moving_lag_bubs`, `3D_moving_lag_particles`)
- [x] Missing `$:END_GPU_PARALLEL_LOOP()` in `s_enforce_EL_bubbles_boundary_conditions` (`m_bubbles_EL.fpp`)

**GPU fixes:**
- [x] `adap_dt_stop` added to GPU `private` clause in `m_bubbles_EE.fpp` (race condition)

**MPI fixes:**
- [x] `s_mpi_reduce_int_sum` `#else` branch for non-MPI builds (`m_mpi_common.fpp`)
- [x] `bubs_glb` initialized to 0 before conditional MPI reduce (`m_mpi_common.fpp`)
- [x] `@:DEALLOCATE` for `p_send_buff`, `p_recv_buff`, `p_send_ids` with `allocated()` guard (`m_mpi_proxy.fpp`)
- [x] `intent(in)` added on `lag_num_ts`, `nBub`, `pos`, `posPrev` in MPI subroutines (`m_mpi_proxy.fpp`)

**Code quality fixes:**
- [x] Duplicate `hyper_cleaning` removed from pre_process namelist (`m_start_up.fpp`)
- [x] Bare `3` -> `3._wp` in vel_model=3 force computation (`m_bubbles.fpp`)
- [x] `beta_vars` indices documented: `1=void fraction, 2=d(beta)/dt, 5=energy source` (`m_constants.fpp`)
- [x] Uninitialized `write_void_evol`, `pressure_force`, `gravity_force` in pre_process (`m_global_parameters.fpp`)
- [x] Stray `sum.png` removed from examples
- [x] Pylint directives removed (replaced by ruff per #1323)
- [x] `path` variable shadowing fixed in `cases.py`

**Test fixes:**
- [x] `2D_moving_lag_bubs` and `3D_moving_lag_particles` added to `casesToSkip` — these examples use `adap_dt` with parameters tuned for 200x200 grids. When reduced to 25x25, the CFL condition forces thousands of internal sub-steps, causing hangs. Stale golden files removed.

---

## Review commit changelog (sbryngelson/Claude Code)

The following commits were added by @sbryngelson via Claude Code review to fix bugs and bring the PR to CI-passing state. All changes are to wilfonba's code; no new features were added.

### `52adde37` — Fix review bugs: Lagrange interpolation, GPU race, MPI, namelist
- **L(1) denominator bug** in 4th-order Lagrange interpolation (`m_bubbles_EL_kernels.fpp:664`): `(xi(2) - xi(5))` corrected to `(xi(1) - xi(5))`
- **L(4) numerator bug**: `(pos - xi(4))` corrected to `(pos - xi(5))` — must exclude own index
- **GPU race condition**: `adap_dt_stop` added to `private` clause in `m_bubbles_EE.fpp` GPU parallel loop
- **Non-MPI build fix**: added `#else; sum = var_loc` to `s_mpi_reduce_int_sum` in `m_mpi_common.fpp`
- **Uninitialized output**: `bubs_glb = 0` added before conditional MPI reduce in both MPI and non-MPI paths
- **Duplicate namelist entry**: removed duplicate `hyper_cleaning` from pre_process `m_start_up.fpp`
- **Stray binary**: removed `examples/3D_moving_lag_particles/sum.png`

### `506dccd2` — Fix optional arg UB, missing deallocates, missing intent
- **Undefined behavior**: removed `case default` in `f_advance_step` (`m_bubbles.fpp`) that wrote to absent optional args `fVel`/`fPos` when called from the EE path with `adap_dt=T`
- **Memory leak**: added `@:DEALLOCATE` for `p_send_buff`, `p_recv_buff`, `p_send_ids` in `s_finalize_mpi_proxy_module`
- **Missing intent**: added `intent(in)` on `lag_num_ts` in `s_initialize_particles_mpi` and `nBub`, `pos`, `posPrev` in `s_add_particles_to_transfer_list`
- **Precision**: `3` -> `3._wp` in vel_model=3 force computation

### `f6adae3a` — Fix dealloc crash on single-rank MPI
- Changed dealloc guard from `if (bubbles_lagrange)` to `if (allocated(p_send_buff))` — the buffers are only allocated when `num_procs > 1` (inside `s_initialize_particles_mpi`), so single-rank MPI runs crashed with "Attempt to DEALLOCATE unallocated"

### `fa2bc4e4` — Fix GPU directive, example physics, documentation
- **Missing GPU directive**: added `$:END_GPU_PARALLEL_LOOP()` after the locate-cell loop in `s_enforce_EL_bubbles_boundary_conditions`
- **Physics bug**: fixed systematic vapor/gas property swap in both `2D_moving_lag_bubs/case.py` and `3D_moving_lag_particles/case.py` — all 10 `bub_pp` thermodynamic parameters (`gam_v`/`gam_g`, `M_v`/`M_g`, `k_v`/`k_g`, `cp_v`/`cp_g`, `R_v`/`R_g`) were inverted vs the correct convention in `3D_lagrange_shbubcollapse`
- **Documentation**: added inline comment documenting `beta_vars` indices in `m_constants.fpp`

### `99dbd4d0` — Fix restart write OOB and uninitialized lag_params
- **Out-of-bounds write**: `s_write_restart_lag_bubbles` allocated buffer with filtered count (`bub_id`) but wrote loop iterated over total count (`n_el_bubs_loc`). Added `particle_in_domain_physical` filter to the write loop using a separate counter `i`
- **Uninitialized variables**: added `write_void_evol = .false.`, `pressure_force = .false.`, `gravity_force = .false.` to pre_process `m_global_parameters.fpp` default initialization

### `d030b8a4` — Remove pylint directives
- Removed 2 `# pylint: disable` comments from `cases.py` (replaced by ruff per upstream #1323)

### `c0f63238` — Formatting and lint cleanup after rebase
- Auto-formatted files after rebase onto master
- Fixed `path` variable shadowing lint error in `cases.py` (renamed to `case_path`)
- Removed `===` separator comment that new source linter flags

### `35a6d5bd` — Skip hanging example tests
- Added `2D_moving_lag_bubs` and `3D_moving_lag_particles` to `casesToSkip` in `cases.py`
- **Root cause**: these examples use `adap_dt=T` with `dt=1` on domain `[-2000,2000]`. When the test framework reduces the grid from 200x200 to 25x25, cell size increases from 20 to 160, but sound speed (~1500 m/s) stays the same. The adaptive stepper gets stuck in thousands of sub-steps trying to satisfy CFL, causing an infinite hang.
- Removed stale golden files for both tests

## AI code reviews

- `/review` — Qodo review


sbryngelson:
## Claude Code Review

**PR #1290** — Moving EL Bubbles with MPI Decomposition

This PR implements moving Euler-Lagrange bubbles with MPI domain decomposition, Newton's 2nd law particle dynamics, and multiple drag/force models. 30+ simulation source files are touched. Below are findings from reviewing the new code against project standards.

---

### Critical Issues

**1. Array out-of-bounds in `s_mpi_sendrecv_particles` when bubbles converge on one rank**
`src/simulation/m_mpi_proxy.fpp`

Lagrangian arrays are allocated `(1:nBubs_glb, ...)`. During `s_mpi_sendrecv_particles`, received bubbles are unpacked with an unconditional increment: `nbubs = nbubs + 1; bub_id = nbubs; bub_R0(bub_id) = ...`. If bubbles from neighbouring ranks converge simultaneously, `n_el_bubs_loc` can exceed `nBubs_glb`, causing silent out-of-bounds writes — silent memory corruption in multi-rank MPI runs.

Fix: add a guard before unpacking:
```fortran
if (nbubs + p_recv_counts(i,j,k) > lag_params%nBubs_glb) &
    call s_mpi_abort("Received bubbles exceed nBubs_glb on rank "//...)
```

**2. Out-of-bounds restart when current rank count exceeds the saved rank count**
`src/simulation/m_bubbles_EL.fpp` ~line 399

```fortran
allocate(proc_bubble_counts(file_num_procs))
bub_id = proc_bubble_counts(proc_rank + 1)   ! proc_rank can be >= file_num_procs
```

Restarting a 4-rank run with 8 ranks accesses `proc_bubble_counts(5..8)` — guaranteed out-of-bounds on ranks 4–7.

Fix:
```fortran
if (proc_rank < file_num_procs) then
    bub_id = proc_bubble_counts(proc_rank + 1)
else
    bub_id = 0
end if
```

---

### Important Issues

**3. No multi-rank CI tests for the headline feature**

Every new test for `vel_model = 1` (tracer) and `vel_model = 2` (Newton's law) uses the default `ppn=1`. The core contribution — `s_enforce_EL_bubbles_boundary_conditions` and `s_mpi_sendrecv_particles` — is never exercised under CI. Issues 1 and 2 above would not be caught by the test suite as written. At minimum a `ppn=2` test for each `vel_model` should be added before merge.

**4. Memory leak risk for newly allocated arrays**

`s_initialize_bubbles_EL_module` now allocates `keep_bubble`, `wrap_bubble_loc`, `wrap_bubble_dir`, `kahan_comp`, `cell_list_start/count/idx`, and conditionally `grad_p_{x,y,z}`, `fd_coeff_{x,y,z}_pgrad`. Please confirm each new `@:ALLOCATE` has a matching `@:DEALLOCATE` in `s_finalize_lagrangian_solver` — missing deallocs leak both GPU device memory and host memory.

**5. `d`-literal `5.0d-11` fails the precheck linter**
`src/simulation/m_bubbles_EL.fpp` line 1345

```fortran
if ((1._wp - q_beta(1)%sf(i, j, k)) > 5.0d-11) then
```

The `d` exponent is forbidden by project standards (CLAUDE.md; `./mfc.sh precheck` will reject it). Replace with `5.0e-11_wp`. (Pre-existing line, but since the file is modified by this PR it should be corrected here.)

---

### Suggestions

- **`bub_id = 0._wp`** (`s_write_restart_lag_bubbles`): real literal assigned to an `integer`. Use `bub_id = 0`.
- **`sizeof()` for MPI file offsets**: `sizeof()` is a GCC extension; project standard is `storage_size()/8`. New call-sites are added by this PR.
- **`drag_model` without `vel_model == 2`**: No `case_validator.py` constraint prevents setting `drag_model > 0` with `vel_model != 2`, which would be silently ignored. Add a dependency check.

---

### Strengths

- `$:GPU_UPDATE(host=...)` before MPI and `$:GPU_UPDATE(device=...)` after receive in `s_enforce_EL_bubbles_boundary_conditions` — GPU coherence pattern is correct.
- `$:GPU_ATOMIC` used correctly for concurrent void fraction accumulation in `s_deltafunc` and `s_gaussian`.
- All new `lag_params` fields broadcast correctly in `s_mpi_bcast_user_inputs`.
- No raw `!` / `!` directives; all GPU regions use `GPU_*` Fypp macros.
- Kahan summation in `s_beta_periodic` / `s_beta_reflective` for void fraction accumulation across MPI boundaries is precision-conscious and correct.

---

The two array out-of-bounds bugs (issues 1 and 2) and the absence of multi-rank tests (issue 3) should be resolved before merge.

sbryngelson:
## Automated Code Review

**Summary:** 2 critical issues, 4 important issues. (Draft PR)

### Critical Issues

**1. Out-of-bounds array access: `mtn_vel(k, l, stage)` when `stage=3`** (confidence: 97%)

`src/simulation/m_bubbles_EL.fpp`, `s_compute_bubble_EL_dynamics`, tracer particle case:
```fortran
mtn_vel(k, l, stage) = mtn_dposdt(k, l, stage)   ! stage can be 3
```
`mtn_vel` is allocated as `(1:nBubs_glb, 1:3, 1:2)` — third dimension is only `1:2`. For `time_stepper==3` (3rd-order TVD RK), `stage` reaches 3, causing an out-of-bounds write. `mtn_dposdt` is correctly allocated to `lag_num_ts` slots.

**Fix:** `mtn_vel(k, l, 2) = mtn_dposdt(k, l, stage)` — write to the temp slot (slot 2), consistent with how the RK integrator stores intermediate values.

---

**2. Missing `@:DEALLOCATE` for `keep_bubble`, `wrap_bubble_loc`, `wrap_bubble_dir`** (confidence: 95%)

`s_initialize_bubbles_EL_module` allocates:
```fortran
@:ALLOCATE(keep_bubble(1:nBubs_glb))
@:ALLOCATE(wrap_bubble_loc(1:nBubs_glb, 1:num_dims), wrap_bubble_dir(1:nBubs_glb, 1:num_dims))
```
None of these have matching `@:DEALLOCATE` calls in `s_finalize_lagrangian_solver`. Per CLAUDE.md: every `@:ALLOCATE` must have a matching `@:DEALLOCATE`.

**Fix:** Add `@:DEALLOCATE(keep_bubble, wrap_bubble_loc, wrap_bubble_dir)` to finalization.

---

### Important Issues

**3. No bounds guard when unpacking received particles in `s_mpi_sendrecv_particles`** (confidence: 85%)

`src/simulation/m_mpi_proxy.fpp`: `n_el_bubs_loc` is incremented for each received particle without checking if it exceeds `nBubs_glb`. All bubble arrays are sized to `nBubs_glb`. During simultaneous boundary crossings (corner transfer), a rank could momentarily receive more particles than removed, silently corrupting memory past array bounds.

**Fix:**
```fortran
if (n_el_bubs_loc > lag_params%nBubs_glb) call s_mpi_abort("n_el_bubs_loc exceeded nBubs_glb during MPI receive")
```

**4. `q_prim_vf` device residency assumed but not verified in `s_enforce_EL_bubbles_boundary_conditions`** (confidence: 88%)

After the MPI exchange, a `GPU_PARALLEL_LOOP` accesses `q_prim_vf(eqn_idx%adv%beg)%sf(cell(1), cell(2), cell(3))` on device. No `GPU_UPDATE` or `ACC_SETUP_SFs` for `q_prim_vf` is called within this subroutine — it relies on the array being device-resident from the outer simulation loop. This should be documented with a comment to prevent future breakage.

**5. `5.0d-11` forbidden `d`-exponent literal** (confidence: 93%)

`src/simulation/m_bubbles_EL.fpp`:
```fortran
if ((1._wp - q_beta(1)%sf(i, j, k)) > 5.0d-11) then
```
CLAUDE.md prohibits `d` exponent literals. `./mfc.sh precheck` will catch this. Fix: `5.0e-11_wp`.

**6. `bub_id = 0._wp` — real literal assigned to integer variable** (confidence: 88%)

`s_write_restart_lag_bubbles`:
```fortran
integer :: bub_id
bub_id = 0._wp
```
Implicit truncation of a real to integer is incorrect Fortran style and will be caught by strict compiler flags. Fix: `bub_id = 0`.

### Suggestions

- `s_write_lag_bubble_evol` writes position (`mtn_pos`) but not translational velocity (`mtn_vel`) — writing velocity would aid post-processing of moving bubble dynamics
- No regression tests for `vel_model` or `drag_model` paths — at minimum a 2D `vel_model=1` test should be added before merge (PR description acknowledges this)
- New `lag_params` fields (`vel_model`, `drag_model`, `pressure_force`, `gravity_force`, `charNz`) are not broadcast in pre_process `m_mpi_proxy.fpp` — confirm these are genuinely not used by pre_process, or add defensive BCAST calls

### Strengths

- Cell-list-based Gaussian smearing avoids GPU atomics entirely — correct choice for Cray/AMD where atomic performance is poor
- GPU coherence around `s_enforce_EL_bubbles_boundary_conditions` carefully handled with explicit host↔device syncs bracketing CPU-side boundary logic
- MPI particle buffer size formula `nReal = 7 + 16*2 + 10*lag_num_ts` correctly accounts for all packed fields including new translational arrays
- `GPU_ROUTINE` added to `s_locate_cell` and `s_get_pinf` enabling their use inside GPU parallel loops
- New `lag_params` fields correctly broadcast in simulation `m_mpi_proxy.fpp`
- Kahan summation in smearing kernels prevents floating-point drift in void fraction accumulation

wilfonba:
@sbryngelson I have a fix for the failing NVHPC runners that I'm testing on the full test suite now. Should be pushed soon.

wilfonba (commented):
I've reviewed all files for unintentional comment changes and made the changes locally. I will push once I've tested.

sbryngelson:
Merged current master into this branch to resolve the conflicts from #1545 (Herschel-Bulkley viscosity). Three files conflicted, all additive: your CFL-tracking locals/reduction combined with HB's `fl` loop iterator in the GPU private lists (`m_data_output`, `m_time_steppers` — both your `dt_local` reduction and their iterator are kept), and both sets of CI skip entries in `cases.py`. No physics changes.

One pre-existing observation while in there: `ccfl_max_loc/ccfl_max_glb` are declared in `m_data_output.fpp` but appear unused — worth a look whether ccfl tracking is still intended.

sbryngelson:
## Merged current `master` into this branch (resolved 11 conflicts + boundary-split fallout)

`master` moved a lot since this branch was last synced — the named-constant/parameter-generation work (#1550–#1553), the global-parameters-core sharing (#1552), auto-generated MPI broadcasts (`generated_bcast.fpp`), and the boundary-module split (#1556) all touch the exact files this PR touches. I reconciled them and pushed the merge commit (`232ff8cd9`).

**Validation:** 3-target CPU build + GPU (nvfortran `acc`) build, EL-bubble golden tests **6/6 on CPU and 6/6 on an H100 GPU**, `./mfc.sh precheck` clean.

The guiding rule was: **keep master's refactored structure (auto-generated params, `_common` shared state, the split BC modules) and re-attach only this PR's genuinely-new state** (`glb_bounds`, `pcomm_coords`, `nidx`, `neighbor_ranks`, the moving-bubble flat vars, and the beta-buffer family).

### ⚠️ Judgment calls worth a look (@wilfonba)

These are the spots where the merge wasn't mechanical and your eyes would help:

1. **`glb_bounds` moved files.** #1556 relocated `s_populate_grid_variables_buffers` from `m_boundary_common` → `m_boundary_io`, so your `glb_bounds` allreduce went with it: `src/common/m_boundary_io.fpp:865` (added `use m_mpi_common` at `:14`). Same logic, new home — please confirm placement (after the offset setup, inside `#ifdef MFC_SIMULATION`).

2. **Beta-buffer family restored by hand.** Taking master's refactored `m_boundary_common` dropped your `s_populate_beta_buffers` + `s_beta_periodic/extrapolation/reflective`; I re-added them at `src/common/m_boundary_common.fpp:211–337`, with `beta_bc_bounds` at `:34` and `use m_mpi_common, only: s_mpi_reduce_beta_variables_buffers` at `:14`. I kept them in `m_boundary_common` (where `m_bubbles_EL` calls them) rather than `m_boundary_io` — flag if you'd prefer them with the other split-out BC routines.

3. **`pcomm_coords` grafted onto the new shared IO routine.** Master extracted parallel-IO setup into `s_initialize_parallel_io_common`/`s_finalize_global_parameters_common`; I call those and just append your `pcomm_coords` alloc/dealloc: `src/simulation/m_global_parameters.fpp:954–955` (+ the matching dealloc). Double-check the alloc ordering is equivalent to your inline version.

4. **pre_process now declares `bc`, `lag_params`, `fd_order/fd_number` locally.** Master's pre_process doesn't use these but your code does, so I declared them pre-process-local: `src/pre_process/m_global_parameters.fpp:53,56–58`. Note: your `lag_params` default-setting in pre_process may now be **redundant** with the shared `s_assign_common_defaults` — worth checking whether it can just be dropped.

5. **Manual broadcasts dropped (now auto-generated).** Your hand-written `lag_params`/`chem_params`/scalar `MPI_BCAST` loops in `m_mpi_proxy` were removed because master auto-generates them via `#:include 'generated_bcast.fpp'` (`src/simulation/m_mpi_proxy.fpp:112`). Your new lag params (`pressure_force`, `gravity_force`, `write_void_evol`, `vel_model`, `drag_model`, `charNz`, `input_path`) are registered in `definitions.py`, so they're auto-broadcast. The hardcoded-patch vars (`interface_file`, `normFac`, …) stay manual since they aren't registry-bound. Please sanity-check nothing that needs a manual broadcast got dropped.

6. **`x_domain` re-activated.** This branch had commented out `type(bounds_info) :: x_domain, y_domain, z_domain`, but master keeps it as a live namelist param (grid generation reads it). Restored at `src/simulation/m_global_parameters.fpp:57`. Note `glb_bounds` is a separate grid-derived global extent, not a rename of `x_domain` — both coexist.

### Mechanical fixes (FYI, lower risk)
- `s_populate_capillary_buffers` gained a `bc` arg in master; updated your call → `src/simulation/m_surface_tension.fpp:275`.
- Adopted master's named constants: `precision == 1` → `precision == precision_single` (`src/simulation/m_bubbles_EL.fpp:1682,1709`), `WENO_TYPE` → `recon_type_weno`.
- post_process: dropped the stale local `eqn_idx`/`qbmm` index computation in favor of master's shared `s_initialize_eqn_idx`; your `beta_idx`-for-lagrange is handled by master's version.

Happy to adjust any of the above if a different call fits your design better.

sbryngelson:
@wilfonba - a few fixes pushed to this branch, plus one left for you.

Pushed (f4acac05):
- Corrected the `lag_params%drag_model` validator message ("0, 1, or 2" -> "0, 1, 2, or 3"; the guard already allows 3).
- Removed a duplicated condition in the Lagrangian send guard in `m_mpi_proxy.fpp` (`abs(i) + abs(j) + abs(k) /= 0` was repeated).

Left for you - precision bug in the Kahan compensation:
`kahan_comp` is declared `type(scalar_field)` (`m_bubbles_EL.fpp`), so `%sf` is `real(stp)`. Under `--mixed` (`stp=half`), the compensation term - which holds the rounding error Kahan is meant to recover - is truncated to 16-bit every step, making it coarser than the error it corrects, so the Kahan pass is effectively a no-op. Same truncation in `s_beta_periodic`/`s_beta_reflective` (`m_boundary_common.fpp`) and `s_mpi_reduce_beta_variables_buffers` (`m_mpi_proxy.fpp`), where `y_kahan`/`t_kahan` are correctly `wp` but the write-back lands in the `stp` field. Fix is to store `kahan_comp` in `wp` rather than `scalar_field`. I held off because that is a data-layout change with `@:ACC_SETUP_SFs` GPU setup across three files and I cannot validate GPU here.

(The `pcomm_coords` early-return leak from the earlier review is already resolved on this branch.)

wilfonba:
> One pre-existing observation while in there: `ccfl_max_loc/ccfl_max_glb` are declared in `m_data_output.fpp` but appear unused — worth a look whether ccfl tracking is still intended.

The relevant computations for the capillary CFL condition were added back by a850436

wilfonba:
Leaving this coment here for myself to edit and comment on each item as I address it.

> 1. **`glb_bounds` moved files.** [Phase-2 completion: post-process context types, m_riemann_solvers split, cleanup #1556](https://github.com/MFlowCode/MFC/pull/1556) relocated `s_populate_grid_variables_buffers` from `m_boundary_common` → `m_boundary_io`, so your `glb_bounds` allreduce went with it: `src/common/m_boundary_io.fpp:865` (added `use m_mpi_common` at `:14`). Same logic, new home — please confirm placement (after the offset setup, inside `#ifdef MFC_SIMULATION`).

This is fine, besides this restructuring being weird.

> 2. **Beta-buffer family restored by hand.** Taking master's refactored `m_boundary_common` dropped your `s_populate_beta_buffers` + `s_beta_periodic/extrapolation/reflective`; I re-added them at `src/common/m_boundary_common.fpp:211–337`, with `beta_bc_bounds` at `:34` and `use m_mpi_common, only: s_mpi_reduce_beta_variables_buffers` at `:14`. I kept them in `m_boundary_common` (where `m_bubbles_EL` calls them) rather than `m_boundary_io` — flag if you'd prefer them with the other split-out BC routines.

These belong in m_boundary_common, so this is all good.

> 3. **`pcomm_coords` grafted onto the new shared IO routine.** Master extracted parallel-IO setup into `s_initialize_parallel_io_common`/`s_finalize_global_parameters_common`; I call those and just append your `pcomm_coords` alloc/dealloc: `src/simulation/m_global_parameters.fpp:954–955` (+ the matching dealloc). Double-check the alloc ordering is equivalent to your inline version.

`pcomm_coords` is only used in simulation, so it should stay outside of `s_initialize_parallel_io_common`. This looks fine.

> 4. **pre_process now declares `bc`, `lag_params`, `fd_order/fd_number` locally.** Master's pre_process doesn't use these but your code does, so I declared them pre-process-local: `src/pre_process/m_global_parameters.fpp:53,56–58`. Note: your `lag_params` default-setting in pre_process may now be **redundant** with the shared `s_assign_common_defaults` — worth checking whether it can just be dropped.

`lag_params` isn't used in post process, so these defaults can't go into the common subroutine. The rest of these pre_process declarations also look fine.

> 5. **Manual broadcasts dropped (now auto-generated).** Your hand-written `lag_params`/`chem_params`/scalar `MPI_BCAST` loops in `m_mpi_proxy` were removed because master auto-generates them via `#:include 'generated_bcast.fpp'` (`src/simulation/m_mpi_proxy.fpp:112`). Your new lag params (`pressure_force`, `gravity_force`, `write_void_evol`, `vel_model`, `drag_model`, `charNz`, `input_path`) are registered in `definitions.py`, so they're auto-broadcast. The hardcoded-patch vars (`interface_file`, `normFac`, …) stay manual since they aren't registry-bound. Please sanity-check nothing that needs a manual broadcast got dropped.

This looks fine

> 6. **`x_domain` re-activated.** This branch had commented out `type(bounds_info) :: x_domain, y_domain, z_domain`, but master keeps it as a live namelist param (grid generation reads it). Restored at `src/simulation/m_global_parameters.fpp:57`. Note `glb_bounds` is a separate grid-derived global extent, not a rename of `x_domain` — both coexist.

`x[y,z]_domain%beg[end]` removed from global parameters and MPI broadcast in simulation.

> * `s_populate_capillary_buffers` gained a `bc` arg in master; updated your call → `src/simulation/m_surface_tension.fpp:275`.
> * Adopted master's named constants: `precision == 1` → `precision == precision_single` (`src/simulation/m_bubbles_EL.fpp:1682,1709`), `WENO_TYPE` → `recon_type_weno`.
> * post_process: dropped the stale local `eqn_idx`/`qbmm` index computation in favor of master's shared `s_initialize_eqn_idx`; your `beta_idx`-for-lagrange is handled by master's version.

These all look fine

>kahan_comp is declared type(scalar_field) (m_bubbles_EL.fpp), so %sf is real(stp). Under --mixed (stp=half), the compensation term - which holds the rounding error Kahan is meant to recover - is truncated to 16-bit every step, making it coarser than the error it corrects, so the Kahan pass is effectively a no-op. Same truncation in s_beta_periodic/s_beta_reflective (m_boundary_common.fpp) and s_mpi_reduce_beta_variables_buffers (m_mpi_proxy.fpp), where y_kahan/t_kahan are correctly wp but the write-back lands in the stp field. Fix is to store kahan_comp in wp rather than scalar_field. I held off because that is a data-layout change with @:ACC_SETUP_SFs GPU setup across three files and I cannot validate GPU here

This is fixed. Kahan compensation is now an optional feature that defaults to true. If it's true, and simulation is run with `--mixed,` `case_validator.py` raises an error saying this is prohibited. Kahan compensation and EL bubbles with WENO/HLLC in half-precision aren't necessary features.

sbryngelson (dismissed):

src/simulation/m_time_steppers.fpp
```diff
@@ -648,7 +641,9 @@ contains
             call s_convert_conservative_to_primitive_variables(q_cons_ts(1)%vf, q_T_sf, q_prim_vf, idwint)
         end if
 
-        $:GPU_PARALLEL_LOOP(collapse=3, private='[vel, alpha, Re, rho, vel_sum, pres, gamma, pi_inf, c, H, qv, fl]')
+        dt_local = huge(1.0_wp)
+        $:GPU_PARALLEL_LOOP(collapse=3, private='[vel, alpha, Re, rho, vel_sum, pres, gamma, pi_inf, c, H, qv, fl]', &
```
anandrdbz: Need to add variable max_dt in private clause, this is why a few openmp cases with cfl_adap_dt = T are failing on frontier, everything else seems to pass

anandrdbz:
@sbryngelson I think this is ready for merge, passes all test cases and fixes the bug with Euler--Lagrange bubbles on GPUs on frontier

wilfonba:
Hmm, I have no idea what got the benchmarks to start completing in the alloted wall time, but I'm not mad about it

sbryngelson:
> Hmm, I have no idea what got the benchmarks to start completing in the allotted wall time, but I'm not mad about it

@wilfonba i sharded it in the upstream (I think).

PR #1298, opened 2026-03-09
Add Herschel-Bulkley non-Newtonian viscosity model

jasontruong2707:
## Summary

Implements Herschel-Bulkley (HB) non-Newtonian viscosity with Papanastasiou regularization, covering power-law, Bingham plastic, and general HB fluids. The effective viscosity at each cell is computed dynamically from the local strain-rate tensor rather than a pre-cached Reynolds number, so viscosity varies spatially in every time step.

The implementation is backward-compatible: all new parameters default to Newtonian behavior (`non_newtonian = F`), and the performance-critical `any_non_newtonian` flag gates the shear-rate computation so purely Newtonian runs are unaffected.

---

## Physics model

The Herschel-Bulkley model with Papanastasiou regularization:

```
μ_eff(γ̇) = τ₀ · (1 − exp(−m·γ̇)) / γ̇  +  K · γ̇^(n−1)
```

where `γ̇ = √(2 Dᵢⱼ Dᵢⱼ)` is the second invariant of the strain-rate tensor. The Papanastasiou term avoids the singularity at `γ̇ → 0` analytically, removing the need for any hard clamp on the shear rate.

Special cases:
- `τ₀ = 0, n = 1`: Newtonian with `μ = K`
- `τ₀ = 0, n ≠ 1`: Power-law fluid
- `τ₀ > 0, n = 1`: Bingham plastic
- General `τ₀ > 0, n ≠ 1`: Herschel-Bulkley

**Nondimensionalization:** All HB parameters (`K`, `τ₀`, `mu_min`, `mu_max`, `mu_bulk`) must be provided in MFC's nondimensional units, i.e., scaled by `ρ_ref · U_ref · L_ref`. At any shear rate, `1/μ_eff` in these units equals the local effective Reynolds number, consistent with how `fluid_pp(i)%Re(1)` is used for Newtonian fluids. The mixture formula `Re_mix = 1 / Σᵢ(αᵢ / Re_eff_i)` is the correct harmonic-mean viscosity mixing rule and applies uniformly to Newtonian, non-Newtonian, and mixed fluid systems.

---

## New parameters (`fluid_pp(i)%...`)

| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `non_newtonian` | logical | `F` | Enable HB model for this fluid |
| `K` | real | `dflt_real` | Consistency index (nondimensional) |
| `nn` | real | `dflt_real` | Flow behavior index (n < 1: shear-thinning, n > 1: shear-thickening) |
| `tau0` | real | `dflt_real` | Yield stress (nondimensional); 0 for power-law fluids |
| `hb_m` | real | `dflt_real` | Papanastasiou regularization parameter; larger values → sharper yield |
| `mu_min` | real | `dflt_real` | Optional lower viscosity clamp (nondimensional) |
| `mu_max` | real | `dflt_real` | Optional upper viscosity clamp (nondimensional); inactive when unset |
| `mu_bulk` | real | `dflt_real` | Bulk viscosity for non-Newtonian fluids (nondimensional); inactive when unset |

All parameters are registered in all four required locations: `params/definitions.py`, `m_global_parameters.fpp` (all three targets), `m_start_up.fpp` (all three targets), and `case_validator.py`.

---

### `src/simulation/m_hb_function.fpp`
Two pure GPU_ROUTINE functions:
- `f_compute_hb_viscosity(tau0, K, nn, mu_min, mu_max, shear_rate, hb_m)` — evaluates the Papanastasiou-regularized HB formula, clamps to `[mu_min, mu_max]` when those are explicitly set (checked against `dflt_real` sentinel, not `> 0`)
- `f_compute_shear_rate_from_components(D_xx, D_yy, D_zz, D_xy, D_xz, D_yz)` — computes `γ̇ = √(2 Dᵢⱼ Dᵢⱼ)` from strain-rate tensor components; no artificial clamp (Papanastasiou handles near-zero analytically)

### `src/simulation/m_re_visc.fpp`
Three pure GPU_ROUTINE subroutines:
- `s_compute_velocity_gradients_at_cell(q_prim_vf, j, k, l, D_xx, ...)` — finite-difference strain-rate tensor at a single cell. Uses 2nd-order central differences in interior; in boundary cells, each of the six tensor components (`D_xx`, `D_yy`, `D_zz`, `D_xy`, `D_xz`, `D_yz`) independently falls back to whatever terms have valid neighbors, accumulating partial contributions rather than zeroing the whole component. This matches the D_xy boundary treatment throughout.
- `s_compute_re_visc(q_prim_vf, alpha, j, k, l, Re_visc_per_phase, [grad_x_vf, grad_y_vf, grad_z_vf])` — fills a `(num_fluids, 2)` array with per-phase `1/μ_eff` (shear, bulk) for the local cell. Optional pre-computed gradient arrays are used when available (viscous flux path); otherwise falls back to `s_compute_velocity_gradients_at_cell`. The Newtonian path (when `any_non_newtonian = F`) early-exits and reads `fluid_pp(q)%Re(i)` directly, preserving pre-PR performance.
- `s_compute_mixture_re(alpha, Re_visc_per_phase, Re_mix)` — harmonic-mean mixture: `Re_mix(i) = 1 / Σ_q(α_q / Re_visc_per_phase(q,i))`, skipping entries at `dflt_real` (inactive phases)

### `examples/2D_lid_driven_cavity_nn/`
Shear-thinning (`n = 0.5`) and shear-thickening (`n = 1.5`) lid-driven cavity cases validated against Li et al. (2015), Re = 500.

### `examples/2D_poiseuille_nn/`
Power-law Poiseuille flow with analytical velocity profile comparison.

---

### `src/common/m_derived_types.fpp`
Adds eight new fields to `physical_parameters`: `non_newtonian` (logical), `tau0`, `K`, `nn`, `mu_min`, `mu_max`, `mu_bulk`, `hb_m` (all `real(wp)`).

### `src/simulation/m_global_parameters.fpp`
- Declares `any_non_newtonian` logical flag (set at startup, GPU_DECLARE'd)
- GPU_DECLARE's the entire `fluid_pp` array so HB parameters are resident on device
- GPU_UPDATE's `fluid_pp` after MPI broadcast so device copy is coherent

### `src/simulation/m_viscous.fpp`
Removes the pre-allocated `Res_viscous(2, Re_size_max)` array and its GPU data region. Each GPU parallel loop now calls `s_compute_re_visc` + `s_compute_mixture_re` per cell, using pre-computed `grad_{x,y,z}_vf` to avoid redundant finite differences. The `any_non_newtonian` flag gates the shear-rate computation so Newtonian runs are unaffected.

### `src/simulation/m_riemann_solvers.fpp`
Removes `Res_gs` cached array. All three Riemann solvers (HLL, HLLC, HLLD) now call `s_compute_re_visc` + `s_compute_mixture_re` per face to compute the local viscous Re. The optional gradient arguments are not passed here (gradients are recomputed from cell-centered primitives via finite differences), which is consistent with the existing first-order viscous flux reconstruction in the Riemann solver path.

### `src/simulation/m_time_steppers.fpp`
`s_compute_dt` converts to primitive variables when `any_non_newtonian` is true before the CFL loop, then calls `s_compute_re_visc` to get the local effective viscosity for the viscous CFL constraint.

### `src/simulation/m_pressure_relaxation.fpp`
Removes `Res_pr` cached array; init/finalize are now empty stubs.

### `src/simulation/m_ibm.fpp`
Previously approximated IBM viscous forces using `μ_eff(γ̇ = 1)` as a spatially uniform reference viscosity — physically wrong for non-Newtonian fluids where viscosity varies with local shear rate. Now:
- Pre-loop still computes `1/Re(1)` for Newtonian fluids (correct, since their viscosity is constant)
- Inside the GPU parallel loop, when `any_non_newtonian`, calls `s_compute_velocity_gradients_at_cell` at the current cell `(i, j, k)` to get the actual local strain-rate tensor, then `f_compute_shear_rate_from_components`, then `f_compute_hb_viscosity` per non-Newtonian fluid
- Volume-fraction-weighted `dynamic_viscosity` is accumulated with the correct per-cell, per-fluid viscosity before computing the viscous stress divergence

### `src/simulation/m_checker.fpp`
Adds `s_check_inputs_non_newtonian` with `@:PROHIBIT` guards: `K > 0`, `nn > 0`, `tau0 ≥ 0`, `mu_min ≥ 0`, `hb_m > 0`. The `mu_max` constraint uses `f_is_default(fluid_pp(i)%mu_max)` (from `m_helper_basic`) to correctly detect whether the user has set the parameter, rather than the incorrect `mu_max < dflt_real` sentinel comparison (which is never true since `dflt_real = -1e6`).

### `src/simulation/m_data_output.fpp`
Adds output of per-fluid HB viscosity fields when `any_non_newtonian` is true.

### `toolchain/mfc/params/definitions.py`
Registers all eight new `fluid_pp` parameters with types, defaults, tags, and documentation strings.

### `toolchain/mfc/case_validator.py`
- Adds `check_non_newtonian()`: Python-side validation of HB parameter ranges (K > 0, nn > 0, tau0 ≥ 0, mu_min ≥ 0, mu_max > mu_min when set, hb_m > 0)
- Restores the `weno_Re_flux requires viscous` constraint to `check_viscosity()` (had been inadvertently removed during refactoring)

### `src/{pre_process,post_process}/m_global_parameters.fpp` and `m_start_up.fpp`
New parameters declared and bound in namelists for all three executables.

---

## Regression tests

Four new test UUIDs in `toolchain/mfc/test/cases.py`, each with golden files:

| UUID | Case | Parameters |
|------|------|------------|
| `0CD345C7` | Shear-thinning | `non_newtonian=T, K=0.1, nn=0.5, tau0=0, hb_m=100` |
| `1E64A4D9` | Shear-thickening | `non_newtonian=T, K=0.1, nn=1.5, tau0=0, hb_m=100` |
| `30713C1C` | Bingham plastic | `non_newtonian=T, K=0.01, nn=1, tau0=0.1, hb_m=100` |
| `45DA4729` | HB with viscosity clamps | `non_newtonian=T, K=0.1, nn=0.7, tau0=0.05, mu_min=1e-4, mu_max=10` |

---

## Test plan

- [x] Builds with gfortran (CPU)
- [x] Precheck (6 CI lint checks) passes
- [x] 2D lid-driven cavity non-Newtonian cases run without errors
- [x] Results match Li et al. (2015) benchmarks (Re=500, n=0.5 and n=1.5)
- [x] Existing viscous Newtonian test cases unaffected (`non_newtonian=F` by default)
- [x] Four regression test golden files added and passing


sbryngelson:
## Code Review — Non-Newtonian Viscosity

Thorough investigation of all findings from the automated review. Each item was verified against the source code.

---

### Critical / Correctness

**1. Index mapping bug in Riemann solver for NORM_DIR=2,3** ⚠️

`src/simulation/m_riemann_solvers.fpp` passes rotated loop indices to `s_compute_re_visc`, but `src/simulation/m_re_visc.fpp` interprets its `j,k,l` arguments as unrotated physical `(x,y,z)` indices and uses `x_cc(j)`, `y_cc(k)`, `z_cc(l)` for gradient computation.

For **NORM_DIR=2** (y-normal faces), the Riemann solver calls:
```fortran
call s_compute_re_visc(q_prim_vf, alpha_L, k, j, l, Re_visc_per_phase_L)
```
Inside `s_compute_re_visc`, `D_xx` is computed using `x_cc(j')` where `j'=k` (which spans the y-direction) — **wrong grid spacing**. Similarly for NORM_DIR=3.

**Impact:** Wrong shear rate → wrong viscosity at y-normal and z-normal Riemann interfaces in 2D/3D non-Newtonian simulations. Most severe on anisotropic grids. Newtonian fluids are unaffected (early exit before gradient computation).

**Affected locations in `m_riemann_solvers.fpp`:** lines ~464-480 (HLL), ~1268-1284 (HLLC), ~2180-2194 (HLLD), ~2827-2841, ~3266-3280.

**Fix options:**
- (A) Pass unrotated physical indices to `s_compute_re_visc` (simplest)
- (B) Make `s_compute_re_visc` accept a `norm_dir` argument and rotate internally

---

**2. GPU coherence: `fluid_pp` not updated on device after MPI broadcast** ⚠️

`src/simulation/m_start_up.fpp` — `s_initialize_gpu_vars()` (lines ~1231-1280) updates hundreds of variables on the GPU device but **does not include `fluid_pp`**. After MPI broadcast populates `fluid_pp(i)%tau0`, `%K`, `%nn`, etc. on the host, these values are never copied to device memory.

GPU kernels in `s_compute_re_visc` (called from `GPU_PARALLEL_LOOP` regions in `m_riemann_solvers.fpp`) read `fluid_pp(q)%tau0`, `fluid_pp(q)%K`, etc. from device memory, which contains stale/uninitialized values.

**Impact:** Silent wrong results on all GPU builds when `any_non_newtonian=True`.

**Fix:** Add `$:GPU_UPDATE(device='[fluid_pp]')` to `s_initialize_gpu_vars()` in `m_start_up.fpp`.

---

### Significant

**3. Hard-coded shear-rate clamp is not parameterizable**

`src/simulation/m_hb_function.fpp`, line 73:
```fortran
shear_rate = min(max(shear_rate, 1.0e-2_wp), 1.0e5_wp)
```
The bounds `[10⁻², 10⁵]` are hard-coded with no corresponding user parameters. For low-Re Bingham flows, the lower bound of 0.01 s⁻¹ is orders of magnitude above realistic values and will artificially cap the yield-stress plateau viscosity. No other MFC code has similar hard-coded clamps.

**Recommendation:** Either make these user-configurable (`fluid_pp(i)%shear_rate_min/max`) or document as a known limitation.

---

**4. Missing pre-computed gradients in Riemann solver calls**

All calls to `s_compute_re_visc` from `m_riemann_solvers.fpp` omit the optional `grad_x_vf`/`grad_y_vf`/`grad_z_vf` arguments. This forces recomputation of velocity gradients from scratch via `s_compute_velocity_gradients_at_cell` (1st-order one-sided differences at boundaries), even though higher-order gradients are already available at the Riemann solver call sites.

**Impact:** Performance regression (redundant gradient computation) and slight inconsistency with the viscous flux gradients. Not a correctness bug.

---

**5. IBM non-Newtonian viscosity uses fixed `gdot=1` reference shear rate**

`src/simulation/m_ibm.fpp`, lines ~1018-1024:
```fortran
dynamic_viscosities(fluid_idx) = f_compute_hb_viscosity( &
    ..., 1._wp, ...)  ! Fixed shear rate = 1
```
The IBM force/torque calculation evaluates viscosity at a constant shear rate of 1 s⁻¹ rather than the local shear rate at each IB ghost cell. The comment "compute reference viscosity at gdot=1" acknowledges this but doesn't explain the physical reasoning.

**Impact:** Incorrect viscous forces on immersed boundaries in flows with spatially varying shear rate. Low severity if IBM + non-Newtonian is not a primary use case.

---

**6. Newtonian performance regression** ✅ No issue

Verified: `s_compute_re_visc` is guarded by `if (viscous)` at all call sites, and the `any_non_newtonian` flag inside the function gates the expensive gradient/HB computation. The Newtonian-only path is a simple `O(Re_size)` lookup. No measurable regression.

---

### Minor (fixed in latest push)

**7. File headers say `.f90` instead of `.fpp`** ✅ Fixed

`m_hb_function.fpp` line 2 and `m_re_visc.fpp` line 2 had `@file m_*.f90` — corrected to `.fpp`.

**8. Boundary shear rate asymmetry**

`s_compute_velocity_gradients_at_cell` uses one-sided differences at boundaries, so `D_xy` only includes one of the two symmetric terms when a neighbor is missing. This underestimates shear at grid boundaries. Low impact for interior-dominated flows.

**9. Missing regression test** ✅ Fixed

Added two non-Newtonian test cases to `toolchain/mfc/test/cases.py`:
- Shear-thinning (nn=0.5, tau0=0, K=0.1)
- Bingham plastic (nn=1, tau0=0.1, K=0.01)

Both run in the single-fluid viscous test block for each dimension.

---

### Summary

| # | Finding | Status | Severity |
|---|---------|--------|----------|
| 1 | Index mapping bug (NORM_DIR=2,3) | **Needs fix** | Critical |
| 2 | GPU coherence (`fluid_pp` not on device) | **Needs fix** | Critical |
| 3 | Hard-coded shear-rate clamp | Needs decision | Medium |
| 4 | Missing pre-computed gradients in Riemann solver | Enhancement | Low |
| 5 | IBM fixed gdot=1 | Document limitation | Low |
| 6 | Newtonian performance | No issue | — |
| 7 | File headers .f90→.fpp | Fixed | Trivial |
| 8 | Boundary shear asymmetry | Known | Low |
| 9 | Missing regression test | Fixed | High |

Items 1 and 2 should be addressed before merge — they produce silently wrong results in 2D/3D and GPU builds respectively.

sbryngelson:
## Claude Code Review

**PR #1298** — Add Herschel-Bulkley non-Newtonian viscosity model

---

### Critical Issues

**1. Hard-coded shear rate clamp overrides user viscosity bounds**
`src/simulation/m_hb_function.fpp:69`

```fortran
shear_rate = min(max(shear_rate, 1.0e-2_wp), 1.0e5_wp)
```

The lower bound of `1e-2` is physically significant and unconditional. For quiescent or near-yield regions where the true shear rate is `< 0.01`, this clamp elevates the shear rate before the HB formula runs, producing a lower viscosity than the physical solution — exactly wrong for yield-stress fluids. The Papanastasiou regularization already handles near-zero shear rates analytically; the clamp is redundant and harmful. The upper clamp similarly bypasses `mu_max`. The clamp should be removed here; viscosity limiting via `mu_min`/`mu_max` is already done correctly inside `f_compute_hb_viscosity`.

**2. IBM force computation uses a fixed non-physical shear rate**
`src/simulation/m_ibm.fpp:932–935`

```fortran
dynamic_viscosities(fluid_idx) = f_compute_hb_viscosity(..., 1._wp, ...)
```

Viscosity at immersed boundary cells is evaluated at `shear_rate = 1.0` regardless of the actual local flow. For shear-thinning fluids near a no-slip wall where the physical shear rate is O(100–10000), this overestimates viscosity by orders of magnitude, producing incorrect IB forces and torques. The local shear rate must be computed at each IB cell using surrounding velocity gradients.

---

### Important Issues

**3. `mu_max` constraint check is logically dead**
`src/simulation/m_checker.fpp:116`

```fortran
@:PROHIBIT(fluid_pp(i)%mu_max < dflt_real .and. fluid_pp(i)%mu_max <= fluid_pp(i)%mu_min, ...
```

`dflt_real = -1.e6_wp` is the sentinel for "not set". The condition `mu_max < dflt_real` (i.e., `mu_max < -1e6`) is essentially never true for any viscosity value. This means the `mu_max > mu_min` constraint is silently never enforced. The correct check is `fluid_pp(i)%mu_max /= dflt_real .and. fluid_pp(i)%mu_max <= fluid_pp(i)%mu_min`.

**4. Mixed Newtonian/non-Newtonian fluids produce dimensionally inconsistent Re**
`src/simulation/m_re_visc.fpp:215–248`

When `any_non_newtonian = .true.` and the simulation has both non-Newtonian and Newtonian fluids:
- Non-Newtonian path stores: `Re_visc_per_phase(q,1) = 1/mu` (units: 1/Pa·s, inverse dynamic viscosity)
- Newtonian path stores: `Re_visc_per_phase(q,i) = fluid_pp(q)%Re(i)` (dimensionless Reynolds number)

`s_compute_mixture_re` then sums `alpha(q)/Re_visc_per_phase(q,i)` across all fluids. Mixing values with different dimensions silently produces a nonsensical result. Consider either adding a `@:PROHIBIT` preventing mixed Newtonian/non-Newtonian fluids, or normalizing the Newtonian path to also store `1/mu`.

**5. `D_xz` corner fallback is inconsistent with `D_xy` treatment**
`src/simulation/m_re_visc.fpp:118–128`

`D_xy` correctly falls back to one-sided stencils for each partial derivative when a neighbor is out-of-bounds. `D_xz` instead returns zero the moment ANY required neighbor index is unavailable, even if one of the two partial derivatives could still be evaluated. At 3D domain edges this underestimates the shear rate and produces artificially high viscosity — inconsistent with the `D_xy` treatment.

---

### Suggestions

- **`K = 0` default will abort instead of warn**: `fluid_pp(i)%K` defaults to `0._wp`. A user who enables `non_newtonian` without setting `K` gets a runtime abort from the checker. Defaulting to `dflt_real` would allow earlier detection, consistent with all other optional parameters.
- **File header comment mismatch**: Both new `.fpp` files use `@file m_hb_function.f90` / `@file m_re_visc.f90` in Doxygen headers. The actual extension is `.fpp`.
- **`HINTS["fluid_pp"]` not updated**: The new non-Newtonian parameters are registered in `definitions.py` but have no hint strings in the `HINTS` dict.

---

### Strengths

- GPU usage is correct throughout: `$:GPU_PARALLEL_LOOP`/`$:END_GPU_PARALLEL_LOOP` for all spatial loops, `$:GPU_ROUTINE(parallelism='[seq]')` for device functions. No raw `!`/`!`.
- `fluid_pp` receives a correct `$:GPU_UPDATE(device='[fluid_pp]')` after the MPI broadcast — GPU coherence for the new fields is sound.
- All four required parameter-system locations are updated (definitions.py, m_derived_types.fpp, m_global_parameters.fpp, namelists via derived-type binding).
- Papanastasiou regularization is physically appropriate for yield-stress modeling near stagnation.
- Six regression test cases with golden files cover the new code paths.

sbryngelson (dismissed):

sbryngelson:
## Automated Code Review — Follow-up Fixes Applied

**Summary:** 3 issues addressed. Fix branch: [sbryngelson/MFC:pr-1298](https://github.com/sbryngelson/MFC/tree/pr-1298)

### Fixed

1. **IGR incompatibility guard** — Added `@:PROHIBIT(igr, ...)` inside `s_check_inputs_non_newtonian` in `m_checker.fpp` (Fortran runtime check) to match the existing Python-side guard in `case_validator.py`. Without this, enabling `igr` with a non-Newtonian fluid silently corrupts primitive variable storage.

2. **Shear-thinning requires `mu_max`** — Added `@:PROHIBIT(fluid_pp(i)%nn < 1._wp .and. f_is_default(fluid_pp(i)%mu_max), ...)` in `m_checker.fpp` and the matching check in `case_validator.py`. Without `mu_max`, a shear-thinning fluid (nn < 1) has unbounded viscosity as shear rate → 0, causing NaN/overflow in practice.

3. **`weno_Re_flux` test coverage** — Added a `shear-thinning+weno_Re_flux` test variant (1D/2D/3D) to the Non-Newtonian test block in `cases.py` with generated golden files.

### Remaining (not addressed)

- `m_hb_function.fpp` line 53: `if (mu_max_val > dflt_real)` should use `f_is_default(mu_max_val)` for consistency. Skipped because adding `use m_helper_basic` to a GPU pure function has portability risks and the behavior is functionally identical (`dflt_real = -1e6`).
- `mu_min` defaulting to 0 rather than requiring explicit set — no fix applied; the current behavior (mu_min=0 is a valid lower bound) is correct.
- Split commit suggestion — not applied; fixes are small enough to be a single commit.

sbryngelson:
@jasontruong2707 Please update this PR or comment on your progress. Closing otherwise.

sbryngelson:
Closing this for now — thanks @jasontruong2707 for the work, the physics and the HB model itself are sound.

The reason isn't the model, it's branch staleness: the branch is ~67 commits behind `master`, and in the meantime `master` refactored the two subsystems this PR integrates into. The Riemann solver dropped the per-direction reconstruction buffers entirely (`qL_prim_rsy_vf`/`rsz_vf` etc. — 74 references on this branch, 0 on `master`), and the immersed-boundary derived types were restructured. So merging `master` in isn't conflict resolution — it's re-porting the solver/viscous/IBM integration onto an interface that no longer exists, in the same file that still has the open rotated→physical index-mapping issue. Forcing the old code through wouldn't compile.

Rather than carry a half-merged branch, we'll re-open fresh on current `master`. The salvageable pieces (still good as-is):

- `m_hb_function.fpp` — self-contained, validated HB + Papanastasiou formula
- The 8 `fluid_pp(i)` parameters + validators + docs + the two example cases

The re-port should rewrite the Riemann/`m_viscous`/IBM integration against master's single-buffer interface and fix the index mapping (centralize the rotated→physical mapping) at the same time.

Anyone picking this up: start from `m_hb_function.fpp` here and re-integrate on top of current `master`.

PR #1318, opened 2026-03-17
Improve parameter checking

danieljvickers:
## Description

Updates the way that we handle IB parameters in case files. Currently, we create a large list of parameters for every IB value for every IB that can be generated, up to the max value of 1000. This generates a list of 40k entries. This refactor just uses a pattern matching replacement which means we check values on a case by case base, improving performance of the tool chain on startup and reducing the array of parameters back to around 3.5k

Fixes #(issue)

### Type of change

- [x] Refactor

## Testing

Ran test suite on IBM cases


sbryngelson:
thanks @danieljvickers i was going to add this but somehow i lost it. i improved it and am merging.

PR #1331, opened 2026-03-20
test: 2.3x faster test suite via direct MPI execution and reduced timesteps

sbryngelson:
## Summary

This PR speeds up the MFC test suite by **1.6-2.4x** across all CI runners through two main optimizations, plus fixes several pre-existing bugs surfaced by the new test execution path.

### Test suite speedup

- **Halve timesteps**: Reduce `Nt` from 50 to 25, cutting simulation time for all 560+ tests. Golden files regenerated.
- **Direct MPI execution**: Instead of spawning `./mfc.sh run` per test (which re-bootstraps Python, re-parses case files, generates shell scripts, runs syscheck), the test runner now generates `.inp` files in-memory and calls the MPI launcher directly. Per-test overhead drops from ~12s to ~2s.

### Self-hosted runner test execution times (excluding SLURM queue)

| Runner | Master | PR | Speedup |
|--------|--------|-----|---------|
| Phoenix CPU | 65m | 29m | **2.2x** |
| Phoenix GPU-ACC | 71m | 29m | **2.4x** |
| Phoenix GPU-OMP | 69m | 28m | **2.4x** |
| Frontier CPU | 35m | 19m | **1.8x** |
| Frontier GPU-ACC (avg shard) | 11m | 8m | **1.4x** |
| Frontier GPU-OMP (avg shard) | 11m | 7m | **1.6x** |
| Frontier-AMD GPU-OMP (avg shard) | 12m | 10m | **1.2x** |
| Frontier-AMD CPU | 30m | 18m | **1.6x** |

### Bug fixes

Bypassing the batch templates for direct MPI execution required replicating several system-specific behaviors. This surfaced (and fixed) pre-existing issues:

- **`s_axis` optional args** (`m_boundary_common.fpp`): `pb_in`/`mv_in` were non-optional in `s_axis()` but passed as absent optionals from `s_populate_variables_buffers` in post\_process. CCE dereferenced the null descriptor causing a segfault on 3D cylindrical post-process tests. Added `optional` attribute and `present()` guards.
- **Stack limit** (`case.py`): Set soft stack limit to hard limit via `resource.setrlimit` at module load, matching `ulimit -s unlimited` in batch templates. Fixes FPE in viscous cylindrical tests on CCE.
- **macOS `setrlimit`**: Wrapped in `try/except` for macOS where the kernel restricts stack limit changes.
- **GPU srun flags** (`mpi_config.gpu_flags`): Added per-template `gpu_flags` list for systems that need explicit `--gpus-per-task`/`--gpu-bind` in srun (tuo, santis). Frontier service partition provides GPUs without GRES, so no srun GPU flags needed there.
- **GPU environment variables** (`mpi_config.gpu_env`): Added `gpu_env` dict for env vars conditional on GPU builds (e.g., `MPICH_GPU_SUPPORT_ENABLED=1` for Frontier CCE). CPU builds explicitly set `MPICH_GPU_SUPPORT_ENABLED=0` to override inherited values. Frontier-AMD (flang) does not set this variable as it uses AFAR/ROCm GPU-aware MPI instead of Cray GTL.
- **System-specific env vars**: Moved `HSA_XNACK`, `FI_CXI_*`, `NV_ACC_*`, `NVCOMPILER_ACC_*` into `mpi_config.env` in respective templates (tuo, santis) so the direct test runner inherits them.
- **srun topology**: Use `--nodes 1 --ntasks-per-node N` instead of `--ntasks N` for srun, matching template dispatch and providing correct node topology for GPU binding.
- **RDMA golden files**: Regenerated RDMA MPI golden files that were missed during the Nt=50→25 regeneration (they require `--rdma-mpi` flag to generate).
- **GPU build detection**: Fixed `bool(ARG("gpu"))` returning `True` for `--no-gpu` (since `bool("no")` is `True` in Python).

## Test plan

- [x] All 560+ tests pass on Frontier CCE GPU (OpenACC) — verified locally on service partition
- [x] All 5 previously-failing 3D cylindrical tests pass (post-process segfault + viscous FPE)
- [x] 3D RDMA MPI 2-rank test passes on Frontier GPU
- [x] CPU tests pass on Frontier (no GTL error)
- [x] macOS CI passes (setrlimit try/except)
- [x] Full CI green: GitHub (ubuntu, macOS), Phoenix (CPU, GPU-ACC, GPU-OMP), Frontier (CPU, GPU-ACC, GPU-OMP), Frontier-AMD (CPU, GPU-OMP)


sbryngelson:
@danieljvickers @wilfonba @anandrdbz @ChrisZYJ 

Curious about your opinion on this PR. All new goldenfiles. Tests are much faster now. 3D tests made somewhat smaller, 1/2 the time steps, and direct MPI execution rather than overhead from hundreds of calls to `./mfc.sh run`.

Cuts the test suite cost by ~3x.

wilfonba:
> @danieljvickers @wilfonba @anandrdbz @ChrisZYJ
> 
> Curious about your opinion on this PR. All new goldenfiles. Tests are much faster now. 3D tests made somewhat smaller, 1/2 the time steps, and direct MPI execution rather than overhead from hundreds of calls to `./mfc.sh run`.
> 
> Cuts the test suite cost by ~3x.

How much impact does doing only the direct MPI execution, without changing all the golden files (so keeping the time steps the same), have? If the time-steps are reduced some of the tolerances may also need to be reduced to keep the same sensitivity to bugs.

sbryngelson:
> > @danieljvickers @wilfonba @anandrdbz @ChrisZYJ
> > Curious about your opinion on this PR. All new goldenfiles. Tests are much faster now. 3D tests made somewhat smaller, 1/2 the time steps, and direct MPI execution rather than overhead from hundreds of calls to `./mfc.sh run`.
> > Cuts the test suite cost by ~3x.
> 
> How much impact does doing only the direct MPI execution, without changing all the golden files (so keeping the time steps the same), have? If the time-steps are reduced some of the tolerances may also need to be reduced to keep the same sensitivity to bugs.

Yeah i noticed this too, but just reducing the overhead is good for 1D cases, but the 3D ones are dominated by runtime not startup.

sbryngelson:
## Claude Code Review

**PR #1331**

This PR delivers a genuine 1.6–2.4x test suite speedup through direct MPI execution and halved timesteps, and correctly fixes the `s_axis` segfault on 3D cylindrical post-process. The architecture of the new test runner is clean. Below are issues that need attention before merge.

---

### Critical Issues

**1. `restart_check` / `run_restart` silently dropped — regression in test coverage**

`toolchain/mfc/test/case.py` — The PR removes `restart_check: bool = False` from `TestCase` and drops `run_restart()`. Simultaneously, `cases.py` removes `restart_roundtrip_tests()`, which contained two tests with `restart_check=True` (1D and 3D). Master's `test.py` calls `case.restart_check` and `case.run_restart()`; the PR removes that entire block.

This silently drops restart-roundtrip verification from the suite. The `run_restart()` method could be updated to call the new `subprocess.run` path instead of being removed. If the omission is intentional, it should be called out explicitly — restart correctness is a meaningful regression risk for any numerical solver.

---

### Important Issues

**2. `None`-value filtering removed from `TestCase.__init__`**

Master:
```python
merge = {key: val for key, val in merge.items() if val is not None}
```

PR: `None` values pass through to `super().__init__`. Currently no test in `cases.py` passes `None` explicitly, but path-based tests loaded via `input.load()` can produce `None` values. A `None` reaching the namelist writer would produce a silent failure or bad `.inp` content.

**3. `frontier.mako` batch path now passes `--cpus-per-task 7` on the `srun` command line**

`toolchain/templates/frontier.mako`: `mpi_config['flags']` now includes `["--unbuffered", "--cpus-per-task", "7"]`, which is expanded into both interactive and batch `srun` invocations. The original batch path expressed this only via `#SBATCH --cpus-per-task=7`. Explicitly passing it to `srun` in batch is usually harmless but worth confirming with Frontier's scheduler since some sites reject the redundant flag.

**4. `_get_mpi_config` auto-detection omits `jsrun`**

The fallback launcher detection tries `["mpirun", "srun", "mpiexec"]` but not `jsrun` (IBM Spectrum MPI, used on Summit and some OLCF nodes). A user on such a system without a matching `-c computer` template would silently get an incorrect launcher.

---

### Suggestions

- **Confirm all ~560 golden files were regenerated for Nt 50→25**: The diff contains ~120 regenerated golden files. A stale golden file after halving `Nt` would either false-pass or false-fail. Worth confirming the full set of tests using the default `Nt` is covered.
- **`helpers.mako` vs `case.py` srun inconsistency**: Template uses `srun --ntasks ${nprocs}` while `_mpi_cmd` generates `srun --nodes 1 --ntasks-per-node ${ppn}`. Functionally equivalent for single-node tests but a comment clarifying why they differ would prevent future confusion.

---

### Strengths

- The `s_axis` fix is correct: making `pb_in`/`mv_in` `optional` with `present()` guards added consistently across all six directional loops in `s_populate_variables_buffers`, plus inside `s_symmetry` and `s_periodic`. Fixes the CCE segfault cleanly.
- `_extract_mpi_config` is robust: brace-depth tracking + `ast.literal_eval` handles nested dicts correctly and returns `None` safely on parse failure.
- Stack limit pre-setting at module import with a `try/except` for macOS is the correct place — inherited by all spawned MPI processes.
- The `bool(ARG("gpu")) == True` fix for `--no-gpu` is a real correctness bug, correctly fixed.
- Unifying `mpi_config` dicts across all templates is a clean pattern.

sbryngelson:
## Automated Code Review

**Summary:** 1 critical issue, 2 important issues. (Draft PR)

### Critical Issue

**1. Restart roundtrip tests silently removed — only automated coverage of restart path gone** (confidence: 90%)

The PR removes `restart_check` from `TestCase`, `run_restart()`, the restart roundtrip verification in `_handle_case`, and `restart_roundtrip_tests()` with its two test cases from `cases.py`. These were the **only** automated tests verifying that `m_start_up.fpp` restart reading and `m_time_advance.fpp` restart writing produce bit-identical results versus a straight-through run. Regressions in the restart data format, IC reconstruction from restart, or mid-simulation state serialization would now go undetected. The PR provides no stated rationale for the removal and no replacement test strategy.

---

### Important Issues

**2. Thread-unsafe `_get_mpi_config` global cache initialization** (confidence: 80%)

`toolchain/mfc/test/case.py`: `_resolved_mpi_config: Optional[MPIConfig] = None` is a module-level global populated lazily. Multiple worker threads from the scheduler will call `_get_mpi_config()` concurrently. The check-then-set pattern is not protected by a lock. The GIL is released during `os.path.isfile`, `file_read`, and `shutil.which` inside the initialization, creating a window where two threads both see `None` and both parse the template. In CPython the outcome is correct (both produce the same result), but this is an unguarded shared mutable global in multi-threaded code — a `threading.Lock()` double-checked locking pattern is the correct fix.

```python
_mpi_config_lock = threading.Lock()

def _get_mpi_config() -> MPIConfig:
    global _resolved_mpi_config
    if _resolved_mpi_config is not None:
        return _resolved_mpi_config
    with _mpi_config_lock:
        if _resolved_mpi_config is not None:
            return _resolved_mpi_config
        # ... resolution logic
```

**3. `None` value filtering removed from `TestCase.__init__`** (confidence: 80%)

The old code filtered `None` values: `{k: v for k, v in merge.items() if v is not None}`. The PR removes this. Without the filter, a test author writing `{"some_key": None}` to remove a parameter would instead generate `key = None` in the Fortran namelist — a parse error. Not triggered by any current test, but a latent hazard.

### Suggestions

- `if ARG("gpu"):` in `test.py` evaluates `bool("no") == True` when running without `--gpu`. The PR correctly fixes this in `case.py` (`ARG("gpu") not in (None, "", "no")`) but not in `test.py`, meaning Gauss-Seidel tests are always skipped regardless of build mode.

### Strengths

- The `MPIConfig` dataclass and `_mpi_cmd` dispatch are clean and well-structured
- `resource.setrlimit` at module load time with `except (ValueError, OSError)` for macOS is the right pattern — fixes viscous cylindrical stack overflow on CCE
- `s_axis` optional parameter fix is correct: `optional` + `present()` guards follow the existing precedent in `s_symmetry`/`s_periodic`
- `gpu_build = ARG("gpu") not in (None, "", "no")` correctly fixes the `bool("no") == True` bug in `case.py`
- `--nodes 1 --ntasks-per-node N` srun topology matches batch template behavior and provides correct GPU binding

sbryngelson:
this is a bad idea. closing. main blocker is that 3d cases dominate test time and this doesn't fix that

PR #1378, opened 2026-04-23
Local Aware IBM

danieljvickers:
## Description

The current code has an issue scaling much past 10k particles due to limitations in the MPIAllReduceSum in the IB force computation. This PR attempts to alleviate this by limiting the number of IBs any given rank can be aware of to its neighbors. This turns the AllReduce compute to a MPI neighbor computation, removing the communication bottlneck. To support this, a massive overhaul of IB ownership between ranks was required.

### Type of change

- [x] Refactor

## Testing

TBD

## Checklist

## AI code reviews

- `/review` — Qodo review


sbryngelson:
## Claude Code Review

**PR #1378 — Local Aware IBM**

This is a draft PR. The concept (replacing global allreduce with neighborhood-local communication for IBM forces) is sound and the scalability motivation is clear. Several correctness issues need to be addressed before this is merge-ready.

---

### Critical Issues

**1. `f_local_rank_owns_location` returns uninitialized result for single-process MPI builds**
`src/simulation/m_collisions.fpp`

The new function wraps its entire ownership logic in `#ifdef MFC_MPI` / `if (num_procs > 1)` but provides no `else` branch. If `num_procs == 1` inside an MPI build, `owns_collision` is never set, producing an uninitialized logical. The old `f_local_rank_owns_collision` handled this correctly with `if (num_procs == 1) then; owns_collision = .true.`. Fix: add `else; owns_collision = .true.` inside the `if (num_procs > 1)` block.

**2. `f_neighborhood_ranks_own_location` silently disables locality check for 1–2 rank runs**
`src/simulation/m_collisions.fpp`

The guard `if (num_procs > 2)` means that for `num_procs == 2` the function unconditionally returns `.true.`, so every IB is treated as neighborhood-owned with no filtering. Two-rank runs are the canonical small test configuration. The threshold should be `> 1`, not `> 2`.

**3. Out-of-bounds potential in `s_reduce_ib_patch_array`**
`src/simulation/m_start_up.fpp`

In 3D with 27 neighbor cells, `num_aware_ibs` can reach `27 * (local IB count)`, which for any non-trivial simulation exceeds `num_local_ibs_max = 2000`. There is no `@:PROHIBIT` or bounds guard before indexing `local_ib_patch_ids(1:num_local_ibs_max)`. Silent memory corruption or segfault will occur once `num_aware_ibs > 2000`.

**4. Analytic MIBM code generation uses local loop index, not global patch ID**
`toolchain/mfc/case.py — __get_analytic_mib_fpp`

The template still emits `if (i == {pid})` where `pid` is the global patch number from the Python case file and `i` is the local loop counter over `num_ibs`. After this PR, `num_ibs` counts locally-owned IBs, so local index `i` no longer maps to the global patch number. All simulations with analytically-defined IB motion will silently apply velocity to the wrong patches in multi-rank runs.

---

### Important Issues

**5. `error stop` in `s_compute_image_points`**
`src/simulation/m_ibm.fpp:485`

```fortran
if (bounds_error) error stop "Ghost Point and Image Point on Different Processors. Exiting"
```

`error stop` is forbidden by project standards (CLAUDE.md). Must be replaced with `call s_mpi_abort(...)`.

**6. `models` array allocated without matching `@:DEALLOCATE`**
`src/simulation/m_ibm.fpp:56`

`s_initialize_ibm_module` has `@:ALLOCATE(models(num_ibs))` but `s_finalize_ibm_module` has no matching `@:DEALLOCATE(models)`. This violates the project rule requiring every `@:ALLOCATE` to have a paired `@:DEALLOCATE`, and leaks GPU device memory.

**7. Post-process `s_write_ib_variable` called from all ranks with orphaned mesh**
`src/post_process/m_data_output.fpp`

The PR removes the `if (proc_rank == 0)` guard, moving `DBPUTPM` (mesh definition) to rank 0 only but leaving `DBPUTPV1` (variable write, inside `s_write_ib_variable`) called on all ranks. Non-zero ranks will write a point variable onto a mesh that was never defined in their per-process SILO file, producing corrupt output or a library error.

**8. `normal_axis` normalization bug in `s_compute_moment_of_inertia`**
`src/simulation/m_ibm.fpp:1154`

```fortran
normal_axis = axis/sqrt(sum(axis))   ! BUG: missing **2
```

Should be `sqrt(sum(axis**2))`. As written this is `sqrt(sum of components)` rather than Euclidean norm, corrupting moment-of-inertia calculation for any 3D rotating IB. Pre-existing issue but touched by this PR.

---

### Suggestions

- `num_local_ibs_max = 2000`: add `@:PROHIBIT(num_local_ibs > num_local_ibs_max, "num_local_ibs exceeds num_local_ibs_max")` at the point of use.
- Mixed `#:if defined('MFC_MPI')` vs `#ifdef MFC_MPI` within the same module — pick one for consistency.
- Testing is marked "TBD". At minimum the existing 2D IBM/MIBM regression tests should be shown passing and a multi-rank (≥4) run with >10 IBs demonstrated before leaving draft.

---

### Strengths

- The architectural direction (local-only IB awareness replacing global allreduce) is the correct approach for scaling to O(10⁴) particles.
- The `s_encode_patch_periodicity`/`s_decode_patch_periodicity` interface cleanly maintains global identity through the periodic marker field.
- The 26-neighbor rank lookup table in `s_compute_ib_neighbor_ranks` avoids repeated topology queries at runtime.
- The post-process SILO simplification (single mesh instead of per-rank mesh) is a positive cleanup.

sbryngelson:
## Automated Code Review

**Summary:** 6 critical issues, 3 important issues. (Draft PR)

### Critical Issues

**1. `error stop` prohibited — use `s_mpi_abort()` instead** (confidence: 100%)

`src/simulation/m_ibm.fpp`, `s_compute_image_points`:
```fortran
if (bounds_error) error stop "Ghost Point and Image Point on Different Processors. Exiting"
```
`error stop` bypasses MPI finalization, hanging other ranks. Replace with `call s_mpi_abort()` or `@:PROHIBIT(bounds_error, "...")`.

---

**2. Missing `@:DEALLOCATE` for `models` array** (confidence: 100%)

`s_initialize_ibm_module` allocates `models` via `@:ALLOCATE(models(num_ibs))` but `s_finalize_ibm_module` has no matching `@:DEALLOCATE(models)`. Since `@:ALLOCATE` also emits `GPU_ENTER_DATA(create=...)`, the GPU memory is leaked. Per CLAUDE.md: every `@:ALLOCATE` must have a matching `@:DEALLOCATE`.

---

**3. Missing `@:DEALLOCATE` for `ghost_points` array** (confidence: 100%)

`s_ibm_setup` allocates `ghost_points` via `@:ALLOCATE(ghost_points(1:max_num_gps))` and separately emits `GPU_ENTER_DATA(copyin=...)`. Neither `s_finalize_ibm_module` nor `s_update_mib` frees it. Add `@:DEALLOCATE(ghost_points)` to finalization.

---

**4. Dead expression in pressure correction loop — result silently discarded** (confidence: 100%)

`m_ibm.fpp`, `s_ibm_correct_state`:
```fortran
q_prim_vf(eqn_idx%E)%sf(j, k, l) + pres_IP/(1._wp - ...)
```
This is a bare expression — not an assignment. The pressure field is never updated. For `moving_ibm > 1`, pressure is always left at `0._wp` from the zero-initialization line above. The commented-out code directly below confirms the intended form is an assignment:
```fortran
q_prim_vf(eqn_idx%E)%sf(j, k, l) = q_prim_vf(eqn_idx%E)%sf(j, k, l) + pres_IP/(...)
```

---

**5. MPI tag accumulation across phases in `s_communicate_ib_forces`** (confidence: 95%)

The back-propagation Fypp loop does not reset `tag = 300` — it continues from wherever the accumulation phase left off. For `num_dims=2`, accumulation x uses tag 300, y uses 302; back-propagation x starts at tag 304 (not 300). Since both sender and receiver apply the same increment, tags happen to match for symmetric topologies — but the protocol is fragile and non-obvious. Reset `tag` to a distinct base at the start of each phase (e.g., 300 for accumulation, 400 for back-propagation).

---

**6. `s_reduce_ib_patch_array` never sets `gbl_patch_id` on single-rank / no-MPI path** (confidence: 90%)

The `#else` (no-MPI) and `num_procs == 1` branches assign `patch_ib(:) = patch_ib_gbl(1:num_aware_ibs)` but never set `patch_ib(i)%gbl_patch_id = i`. The multi-rank MPI path sets this explicitly. Although `s_communicate_ib_forces` returns early for `num_procs == 1`, `s_get_neighborhood_idx` with `gbl_patch_id = 0` would never find a match, silently corrupting any path that reaches it. Fix: add `patch_ib(i)%gbl_patch_id = i` in the single-rank path.

---

### Important Issues

**7. `requests(52)` fixed array may overflow when `ax >= 3`** (confidence: 85%)

`s_compute_ib_neighbor_ranks`: `nreqs` is never reset to 0 between iterations of `do k = 2, ax`. On the second round, `nreqs` carries over from the first, incrementing past index 52 — out-of-bounds write. Fix: reset `nreqs = 0` at the top of each `k` iteration.

**8. De-duplication scheme in force accumulation untested for `ib_neighborhood_radius > 1`** (confidence: 80%)

The snapshot-based force de-duplication (subtract previous snapshot, add new) is sound for radius=1 but relies on each IB appearing at most once per received message. For radius > 1 with non-linear neighborhood topology, duplicate accumulation is possible. Needs explicit testing.

**9. `recv_forces_snap` reset is per-dimension but inner loop overwrites unconditionally** (confidence: 80%)

Minor design concern for the radius > 1 case — the accumulation loop structure may not generalize cleanly beyond the tested radius=1 case.

### Suggestions

- Debug `print *` left in `s_handoff_ib_ownership`: `print *, proc_rank, " New Owner ", ...  ! TODO :: REMOVE THIS DEBUG PRINT` — remove before merge
- Commented-out `! print *, proc_rank, num_local_ibs, ...` in `s_ibm_setup` — remove before merge

### Strengths

- Neighborhood rank discovery protocol is well-structured (face→edge→corner seeding with distinct tag range)
- `gbl_patch_id` matching across ranks is architecturally sound for the moving IB migration case
- `storage_size(0)/8` used correctly for byte sizing (no hardcoded widths)
- GPU_* Fypp macros used consistently throughout — no raw `!$acc`/`!$omp` directives found

sbryngelson:
You had an issue that I think I fixed. 


>  Root cause: num_ib_patches_max = 2,050,000 caused a ~2.25 GB unconditional allocation + full init loop on every startup (even for non-IBM cases), crashing CI debug builds.
>
>  Fix — 3 files, 2 commits:
>
>  1. m_constants.fpp: Added num_ib_patches_max_namelist = 50000 — the small initial budget (same as before particle beds were added).
>  2. m_global_parameters.fpp + m_start_up.fpp: Initial patch_ib allocation uses num_ib_patches_max_namelist on all ranks (~55 MB instead of ~2.25 GB).
>  3. m_particle_bed.fpp: Grows patch_ib to num_ib_patches_max via MOVE_ALLOC only when particle beds are actually being generated (rank 0 only, before the MPI broadcast).
>  4. m_mpi_proxy.fpp: Non-rank-0 processes grow patch_ib to match if num_ibs > size(patch_ib) — handles the MPI > 1 case where rank 0 grew for particle beds but other ranks still
>  have the small allocation.
>  5. m_start_up.fpp s_reduce_ib_patch_array: Changed the 2.25 GB stack-allocated patch_ib_gbl to a heap-allocated array sized to exactly num_ibs — eliminates the SIGSEGV for IBM
>  cases.

sbryngelson:
I generated and pushed the missing golden file for `BA186DDF` (`2D -> Example -> mibm_particle_bed`) directly to your branch — that should fix the CI failure.

While reviewing the diff I found several other issues worth addressing:

---

## Critical (wrong results)

**`f_neighborhood_ranks_own_location` guard is `num_procs > 2` instead of `num_procs > 1`** (`m_collisions.fpp`)
On a 2-rank run the condition is false, so both ranks think they own every collision and forces get double-applied. Should match `f_local_rank_owns_location`'s guard.

**`recv_forces_snap` not zeroed before y/z accumulation passes** (`m_ibm.fpp`, `s_communicate_ib_forces`)
`recv_forces_snap = 0._wp` lives only inside the x-dimension block. For y and z it accumulates stale values from the previous dimension, making the double-count subtraction remove too much force.

**`s_handoff_ib_ownership` can silently drop newly received patches** (`m_ibm.fpp`)
After the compact-and-delete loop, `ib_gbl_idx_lookup` still reflects the old layout (it is rebuilt *after* unpacking). `s_get_neighborhood_idx` is therefore testing the pre-handoff table — a just-evicted patch may still appear as resident, causing the newly received authoritative state to be discarded.

---

## High (MPI / GPU correctness)

**`recv_forces_snap`, `recv_torques_snap`, `recv_ids`, `recv_ft` allocated with plain `allocate`, not `@:ALLOCATE`** (`m_ibm.fpp`)
These arrays are used inside `GPU_PARALLEL_LOOP` via `copyin` clauses, but plain `allocate` skips `GPU_ENTER_DATA`. On OpenACC builds the GPU sees uninitialized device memory.

**`patch_ib` re-allocated with plain `allocate`/`deallocate` while `GPU_DECLARE` is active** (`m_global_parameters.fpp` + `m_start_up.fpp`)
`patch_ib` was changed from a static array to allocatable, but the device pointer registered by `GPU_DECLARE` at startup goes stale every time `s_reduce_ib_patch_array` or the MPI proxy resizes it. Each reallocation should use `@:ALLOCATE`/`@:DEALLOCATE`.

**`neighbor_domain_x/y/z` computed on host but never `GPU_UPDATE`'d to device** (`m_start_up.fpp`)
These are `GPU_DECLARE`'d and used inside `f_neighborhood_ranks_own_location` (a `GPU_ROUTINE`). After `get_neighbor_bounds` fills them on the CPU, a `GPU_UPDATE(device=[neighbor_domain_x, neighbor_domain_y, neighbor_domain_z])` is needed.

**`s_compute_ib_neighbor_ranks`: `nreqs` not reset between outer loop iterations** (`m_start_up.fpp`)
After `MPI_WAITALL`, `nreqs` retains its end value going into the next `k` iteration. With `ib_neighborhood_radius > 2` this overflows the `dimension(52)` requests array.

**Post-process SILO writes called from all ranks** (`post_process/m_data_output.fpp`)
`DBPUTMVAR` to `dbroot` is now called from all ranks, not just rank 0. In `file_per_process=F` mode this produces duplicate or corrupted multi-mesh entries.

---

## Medium (robustness / performance)

**`num_local_ibs_max = 2000` hard cap with no overflow check** — `local_ib_patch_ids(num_local_ibs) = num_ibs` writes out of bounds silently if more than 2000 particles land on one rank. Needs an `@:PROHIBIT`.

**`ib_force_send_buf`/`recv_buf` heap-allocated on every RK stage call** (`m_ibm.fpp`) — these should be module-level buffers sized once in `s_ibm_setup` rather than allocated and freed on every timestep.

**`moving_immersed_boundary_flag` not recomputed after `s_handoff_ib_ownership`** — set once at startup, then stale after each ownership handoff cycle when moving IBs migrate between ranks.

---

## Low (code quality)

- `s_get_neighborhood_idx` has an unused local variable `i` (compiler warning)
- `f_xorshift` can return exactly `1.0` (violates the `[0, 1)` contract in the docstring — particle placed at `xmax`)
- `particle_bed_parameters` type defined in simulation-only `m_global_parameters.fpp`; convention is to put derived types in `m_derived_types.fpp`

---

The two highest-priority fixes are the `num_procs > 2` off-by-one (one-character fix, guaranteed wrong on 2-rank jobs) and `recv_forces_snap` not being zeroed per-dimension (wrong forces in any 2D/3D multi-rank run).

sbryngelson (dismissed):

danieljvickers:
Addressing AI comments:

All of the critical issues are incorrect for one reason or another. I have gone through and verified that the relevant chunks of code are in place to address the concerns it has.

As for the neighbor boundaries, they are never called in a GPU region, so this comment is mute. I have removed the GPU routine call to help with compilation in response.

The file per process F claim is just false, and I tested that the code runs find without it. But it is true that we do not need other ranks to be doing any of this calculation regardless. I extended the rank 0 check in response.

nreqs is in fact reset between loop iterations.

Snapshot arrays are in fact reset between loop iterations

It is confused about the need to GPU allocate arrays for patch_ib. Because the array is used with a declare, but never actually allocated on the GPU, we just have the GPU pointer. When we perform the first copy to the GPU to allocate the array size, we have already done the reshape. This means that the GPU array is properly sized. This is important because the size of the patch_ib array is significant in the new code and could causes issues with limiting problem size because the GPU must be able to hold the global patch_ib array, even though it would then immediately be deallocated and reallocated to a smaller array. If the number of ib patches grows to the extent that it costs many for GB of memory, this could cause memory limitations. The current implementation should be the correct way to go.

the moving IBM flag should not be recomputed. You want it to be the same on all ranks. There is argument that it should be moved to the case level for case optimizaiton later, but that is a future PR.

Force arrays getting moved is going to be handled in a future PR that already has a github issue I created yesterday.

Prohibit for number of local IBs resolved.

In summary, two of these I made minor adjustments for. One I reject for reasons that I believe are valid and better for long-term code support. All others are incorrect.

danieljvickers:
General Note: Mac errors are on sections of the code that were not touched by this PR (bubbles), and it is a random dyl linker error. I get the same error on reldebug on a chemsitry case. I get the same error locally, but on a different set of tests. This seems like a MacOS instability issue, and not related to the particular PR. Going to attempt to rerun. If ti fails again, but on different tests, then it looks like the MacOS test suite checks may be broken.

Edit: Rerunning the job changed the jobs which failed. This appears to be a spurious failure for MacOS specifically. It appears to be specific to the branch, as other branch are not experiencing it. Will investigate.

danieljvickers:
@sbryngelson review?

PR #1379, opened 2026-04-23
WENO/MUSCL case guard w/ recon_type number + Inlining typo causing segmentation faults on compilation

Cowsreal:
## Description

1. I ran a whole batch of cases with WENO parameters set while recon_type = 2 was also set. There is no case guard, so this resulted in a MUSCL order 0 simulation being ran, where the toolchain should have warned the user.

2. I kept running into segmentation faults with phase_change.fpp while trying to compile using OpenMP + CCE 19 on Frontier. The phase change module extensively uses the $:GPU_ROUTINE macro, specifically, the cray_noinline option. It's currently doing nothing when it evaluates the _CRAYFTN branch. I was able to compile after fixing this typo inside the parallel_macros.fpp. 

### Type of change

- [ ] Bug fix
</details>

## AI code reviews

- `/review` — Qodo review


Cowsreal:
@claude full review

Cowsreal:
Issue: MUSCL guard add certain real valued WENO parameters, that have default values set in BASE_CFG which don't fail the test cases. 

Solution: In the corresponding test cases, push None for those parameters, then pop them out inside the TestCase constructor before they trip schema. This does not interfere wiht any already existing test cases since no one else uses None.

Cowsreal:
@coderabbitai review

sbryngelson:
why do you get seg faults but no one else?

Cowsreal:
The information we have is it was only the non case optimized OpenMP runs on Frontier, where it was crashing trying to expand one of the gpu collapse 3 loop. I tried to compile it myself on Frontier, with same setup, no crash.

sbryngelson:
It's hard to parse what you are saying, but if I understand correctly (non-case optimized, either CCE or AMD compilers, Frontier), then those cases run in CI.

Cowsreal:
I missed a "when" in my last comment. I'll be more precise here.

Last time, when the test suite ran, most of the machines successfully built. However, they all failed because of an edge case I missed.

ONLY 2 segfaulted on compilation, which were both Frontier, CCE OpenMP, NON case-optimized builds. (All other Frontier runs built, but failed the tests) 

The call stack in the segfault outputted 

`namespace)::TransposeArrays::get_transposed_symbol(int) v_openmp.c:0:0      
#4 0x0000000000b07a0f transpose_rank_expanded_private_arrays()`
`...`
` "/opt/cray/pe/cce/19.0.0/cce/x86_64/bin/../../../cce/x86_64/bin/optcg" was 
terminated due to receipt of signal 013:  Segmentation fault (core dumped).  `
`gmake[3]: *** [CMakeFiles/simulation.dir/build.make:1477: CMakeFiles/simulation.dir/fypp/simulation/m_variables_conversion.fpp.f90.o]  
Error 2  `

which is what leads me to think it crashed inside the variable_conversion module trying to expand the private array of one of the $:GPU_PARALLEL_LOOP macros.

sbryngelson:
> why do you get seg faults but no one else?

can you just answer this q

Cowsreal:
I don't know

PR #1383, opened 2026-04-28
muscl_eps

ChrisZYJ:
## Description

  - Add muscl_eps parameter to control MUSCL limiter slope-product threshold. Default preserves existing behavior;
  muscl_eps = 0 gives textbook limiter behavior.

### Type of change

- [x] New feature

## Testing

All 30 existing MUSCL tests tests passed. Bit-identical default behavior as the old MUSCL code. 
15 new muscl_eps=0 tests added (5 limiters x 3 dims)

## Checklist

## AI code reviews

- `/review` — Qodo review


toolchain/dependencies/CMakeLists.txt
sbryngelson, line 1: you're mixing contributions in one pr

src/simulation/m_global_parameters.fpp
```diff
@@ -856,6 +859,15 @@ contains
             $:GPU_UPDATE(device='[igr, igr_order, igr_iter_solver]')
         #:endif
 
+        ! muscl_eps: use per-limiter defaults when user did not set it
+        if (muscl_eps == dflt_real) then
```
sbryngelson: using == to compare real to real. i think we have f_approx_equal for this

src/simulation/m_start_up.fpp
```diff
@@ -83,7 +83,7 @@ contains
 
         namelist /user_inputs/ case_dir, run_time_info, m, n, p, dt, &
             t_step_start, t_step_stop, t_step_save, t_step_print, &
-            model_eqns, mpp_lim, time_stepper, weno_eps, &
+            model_eqns, mpp_lim, time_stepper, weno_eps, muscl_eps, &
```
sbryngelson, line 86: 2 vars -> 1 via `reconst_eps` which is either weno_eps or muscl_eps, depending on which method is used?

toolchain/dependencies/CMakeLists.txt
ChrisZYJ: I'll open a new PR for that

src/simulation/m_global_parameters.fpp
ChrisZYJ: Yeah sorry I introduced f_approx_equal a long time ago but forgot about it

src/simulation/m_start_up.fpp
ChrisZYJ: I feel that we should keep "weno_eps" for backwards compatibility to not break existing case files, so maybe we can use weno_eps for both, or have both weno_eps and muscl_eps. Personally I feel that muscl_eps for MSUCL is cleaner?

sbryngelson: I think going to one, then updating all the case files at once via a sed replace (AI-friendly!) is fine.

ChrisZYJ: I thought about this a bit more, but I'm hesitant to remove weno_eps because it would break all existing case files, including users' offline case files that have weno_eps. But you're right that using a single variable for both would be cleaner. Could we just keep weno_eps and use it for both WENO and MUSCL? Alternatively, I could replace all weno_eps and muscl_eps with recon_eps in the simulation code, but keep weno_eps as a legacy option in the case file (i.e., if recon_eps isn't set, it defaults to weno_eps). This way, we maintain backward compatibility.

ChrisZYJ: On second thought, even using weno_eps as legacy could be messy for CI as weno_eps=1e-16 was set in BASE_CFG, but we have different default behaviors when _eps is unset for WENO, and MUSCL. Since weno_eps and muscl_eps serve different mathematical purposes in the code and have different default behaviors, so I'm inclined to just keep both weno_eps and muscl_eps. Would that be all right?

PR #1397, opened 2026-05-05
Fix 1D flow perturbation applying random factor twice (index aliasing)

### LINKED ISSUE #1369: bug: 1D flow perturbation applies random factor twice due to index aliasing
## Summary

In `src/pre_process/m_perturbation.fpp`, `s_perturb_surrounding_flow` applies a random perturbation to momentum:

```fortran
q_prim_vf(eqn_idx%mom%end)%sf(i, j, k) = rand_real*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k) = (1._wp + rand_real)*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
```

In 1D (`n == 0, p == 0`), `num_vels == 1`, so `eqn_idx%mom%end == eqn_idx%mom%beg`. Line 1 writes `rand * v0` into the only velocity slot, then line 2 reads that same (now-modified) slot and writes `(1 + rand) * rand * v0`. The perturbation is applied twice and compounded, giving a result of `rand*(1+rand)*v0` instead of the intended `(1+rand)*v0`.

## Fix

Swap the two lines so `mom%beg` is scaled first, then `mom%end` is set from the original value:

```fortran
q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k) = (1._wp + rand_real)*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
if (num_vels > 1) q_prim_vf(eqn_idx%mom%end)%sf(i, j, k) = rand_real*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
```

## Discovered

Found during review of PR #1365. Pre-existing in `master`.

sbryngelson:
## Summary

Fixes #1369.

In `s_perturb_surrounding_flow` (`src/pre_process/m_perturbation.fpp`), the original code was:

```fortran
q_prim_vf(eqn_idx%mom%end)%sf(i, j, k) = rand_real*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k) = (1._wp + rand_real)*q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
```

In 1D, `eqn_idx%mom%end == eqn_idx%mom%beg` (same slot), so:
- Line 1 writes `rand * v0` into the only velocity slot
- Line 2 reads the now-modified slot and writes `(1 + rand) * rand * v0`

The perturbation is applied twice, giving `rand*(1+rand)*v0` instead of `(1+rand)*v0`.

## Fix

Save the original `mom%beg` value before any writes, scale `mom%beg` first, then only assign `mom%end` when `num_vels > 1`:

```fortran
v_beg = q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k)
q_prim_vf(eqn_idx%mom%beg)%sf(i, j, k) = (1._wp + rand_real)*v_beg
if (num_vels > 1) q_prim_vf(eqn_idx%mom%end)%sf(i, j, k) = rand_real*v_beg
```

This also correctly preserves multi-D behaviour: `mom%end` gets `rand * v_beg` (the original x-velocity), not `rand*(1+rand)*v_beg`.

## Verification

Ran a 1D case (`m=9, v0=1.0, perturb_flow_mag=0.1`) and read the post-pre_process velocity field. All 10 cell values fell in `[1.0, 1.1]` — consistent with `(1 + rand) * v0` where `rand ∈ [0, 0.1]`. Under the old bug the values would have been near zero (`rand*(1+rand)*v0 ≪ 1`). Not in the test suite, so verified manually.


sbryngelson:
## Automated Code Review

**Summary:** Fix is correct. One suggestion.

### Strengths

The fix is mechanically correct:
- `num_vels > 1` is the right guard — it uses the same quantity that determines whether `mom%beg` and `mom%end` are aliased, and also correctly handles the MHD case where `num_vels = 3` even in 1D
- `v_beg` correctly typed as `real(wp)` — promotes `stp` field data to `wp` for computation, consistent with project conventions
- 2D/3D behaviour exactly preserved: `mom%beg` gets `(1+rand)*v_beg`, `mom%end` gets `rand*v_beg` — identical to the original code's output in non-aliased cases
- `num_vels` is guaranteed initialized before `s_perturb_surrounding_flow` is ever called

### Suggestion

No automated test covers `perturb_flow`. Since the bug went undetected, a programmatic test case in `toolchain/mfc/test/cases.py` would prevent regression — a simple 1D case with `perturb_flow: T, perturb_flow_fluid: 1, perturb_flow_mag: 0.1` whose golden output should fall in `[1.0, 1.1]` for a unit-velocity IC.

PR #1398, opened 2026-05-05
fix: free tau_Re_vf in all viscous runs, not only cyl_coord

### LINKED ISSUE #1366: bug: tau_Re_vf memory leak in Cartesian viscous runs (dealloc only under cyl_coord)
## Summary

In `src/simulation/m_rhs.fpp`, `tau_Re_vf` is allocated inside `if (viscous)` but deallocated only when `cyl_coord` is also true:

```fortran
! Allocation (inside if (viscous)):
@:ALLOCATE(tau_Re_vf(1:sys_size))
do i = 1, Re_size(1)
    @:ALLOCATE(tau_Re_vf(eqn_idx%cont%end + i)%sf(...))
end do
@:ALLOCATE(tau_Re_vf(eqn_idx%E)%sf(...))

! Deallocation (inside if (cyl_coord)):
if (cyl_coord) then
    do i = 1, num_dims
        @:DEALLOCATE(tau_re_vf(eqn_idx%cont%end + i)%sf)
    end do
    @:DEALLOCATE(tau_re_vf(eqn_idx%E)%sf)
    @:DEALLOCATE(tau_re_vf)
end if
```

Any run with `viscous = .true.` and `cyl_coord = .false.` (i.e., every Cartesian viscous simulation) leaks all `tau_Re_vf` GPU and host allocations. Violates the `@:ALLOCATE`/`@:DEALLOCATE` pairing rule.

## Fix

The deallocation guard should be `if (viscous)`, not `if (cyl_coord)`. The number of `sf` sub-arrays to free mirrors the allocation loop (one per viscous fluid, plus the energy slot).

## Discovered

Found during review of PR #1365. Pre-existing in `master`.

sbryngelson:
## Summary

Fixes #1366.

In `s_finalize_rhs_module`, `tau_Re_vf` is allocated inside `if (viscous)` but was only deallocated inside `if (cyl_coord)`. Any Cartesian viscous run (`viscous = .true.`, `cyl_coord = .false.`) leaked all `tau_Re_vf` host and GPU allocations.

The fix removes the erroneous `if (cyl_coord)` guard — the deallocation is already inside the `if (viscous)` block, so no additional condition is needed.

## Change

`src/simulation/m_rhs.fpp`: drop the `if (cyl_coord) then ... end if` wrapper around the three `tau_re_vf` `@:DEALLOCATE` calls.

## Test plan

- [ ] Existing viscous regression tests pass
- [ ] Existing cylindrical-coordinate viscous tests pass (deallocation still fires via the outer `if (viscous)` guard)


sbryngelson:
The fix is correct. The allocation of `tau_Re_vf` at line 333 of `m_rhs.fpp` is guarded only by `if (viscous)` — there is no `cyl_coord` guard anywhere near the allocation:

```fortran
if (viscous) then
    @:ALLOCATE(tau_Re_vf(1:sys_size))
    do i = 1, num_dims
        @:ALLOCATE(tau_Re_vf(eqn_idx%cont%end + i)%sf(...))
    end do
    @:ALLOCATE(tau_Re_vf(eqn_idx%E)%sf(...))
    ...
end if
```

The old code had the deallocation incorrectly nested inside `if (cyl_coord)`, which had no corresponding guard on the allocation — that asymmetry was the original bug causing a memory leak in Cartesian (`cyl_coord = .false.`) viscous runs. The fix removes the spurious `cyl_coord` guard from the deallocation so both allocation and deallocation are symmetric: they both execute whenever `viscous = .true.`, regardless of `cyl_coord`.

sbryngelson:
## Automated Code Review

**Summary:** No issues found. ✅

Verified:
- All three `tau_Re_vf` allocation levels are now symmetrically paired with their deallocations after removing the `if (cyl_coord)` guard
- GPU deallocation is fully covered — `@:DEALLOCATE` macro always emits `GPU_EXIT_DATA(delete=...)` before the Fortran `deallocate`
- No double-free risk: `tau_Re_vf` is allocated in exactly one place (inside `if (viscous)`) and deallocated in one matching block
- The `igr` path is consistent: both allocation (`if (.not. igr)`) and deallocation (`if (.not. igr)`) carry the same guard — `igr=true` runs unaffected

Minimal, correct fix. Safe to merge.

PR #1400, opened 2026-05-05
fix: correct single-precision CI testing

### LINKED ISSUE #759: Single/modular precision fix ups
Some fix-ups relevant to PR #632 by @arciyer123 

* These cases are skipped in testing. I'm not sure why.
	* 1D -> 2 MPI Ranks, 2D -> 2 MPI Ranks, 3D -> 2 MPI Ranks
* Add benchmark check that ensures we see the expected speedup of single over double precision (and throw a non-zero stop code if we don't)

These are the two most pressing, I think. Everything else I can think of is more of a feature than something that needs to be fixed.

sbryngelson:
Closes #759

## Summary
- Add `\$PRECISION` to the **Test** step in \`.github/workflows/test.yml\` so single-precision binaries are actually tested with \`--single\` tolerances (root bug: \`\$PRECISION\` was only in the Build step env, so the Test step ran double-precision test selection/tolerance against single-precision output)
- Change single-precision CI job from \`mpi: no-mpi\` to \`mpi: mpi\` so \`--test-all\` is passed and a broader test set is exercised

## Test plan
- [ ] CI single-precision job builds with \`--mpi --single\`
- [ ] CI single-precision job runs tests with \`--single\` flag (1e-1 tolerances)
- [ ] No regressions in regular (double-precision) CI jobs


sbryngelson:
## Automated Code Review

**Summary:** 0 critical issues, 0 issues.

### Retraction

~~**`--only-changes` and `--test-all` are now both active simultaneously on PRs for the single-precision job**~~ — **this was a false positive.**

The two flags are orthogonal: `--only-changes` controls *which tests* are run (tests affected by changed files), while `--test-all` controls *which stages* are run per test (pre_process + simulation + post_process vs. pre_process only). They compose without conflict. All other `mpi: mpi` jobs already get both flags on PRs, so this job's behavior is consistent with the rest of the matrix. No issue here.

### Suggestions

- The primary fix (missing `PRECISION` in the Test step `env:` block) is correct. GitHub Actions step-level `env:` blocks are scoped to that step only — `$PRECISION` in the Test step was silently expanding to empty string, running single-precision outputs against double-precision tolerances.
- The `mpi: no-mpi` → `mpi: mpi` change is not strictly required to fix the tolerance bug. Splitting into two commits (tolerance fix; MPI/coverage expansion) would aid bisection if the broader test suite reveals failures.
- `PRECISION` expression is now duplicated identically in Build and Test step `env:` blocks — not a bug, but a comment noting why the duplication exists would help future maintainers.

### Strengths

The root cause diagnosis is correct. The fix is minimal and correct. `format('--{0}', matrix.precision)` handles empty string and `single` exactly right. No secrets involved.

PR #1401, opened 2026-05-05
fix: correct MTHINC normals on non-uniform grids

### LINKED ISSUE #1395: MTHINC: interface normals inaccurate on non-uniform (stretched) grids
## Summary

`s_compute_mthinc_normals` in `src/simulation/m_thinc.fpp` computes interface normals using raw index-space differences:

```fortran
nr_x = (alpha(j+1) - alpha(j-1)) * 0.5
nr_y = (alpha(k+1) - alpha(k-1)) * 0.5
nr_z = (alpha(l+1) - alpha(l-1)) * 0.5
```

No physical grid spacing (dx, dy, dz) is used.

## Why this is correct on uniform grids

The MTHINC formulation works entirely in **reference space** — the unit cell ξ ∈ [−½, ½]^ndim. `f_mthinc_volume_integral` and `f_mthinc_face_average` integrate over reference coordinates with face positions fixed at ±½. The reference-space gradient of α is:

```
∂α/∂ξ ≈ (α(j+1) − α(j−1)) / 2
```

which is exactly what the code computes (the ×0.5 cancels on normalization). On a **uniform grid**, reference-space and physical-space normals coincide (up to a global scale factor), so the result is correct.

## Why this is inaccurate on non-uniform grids

On a non-uniform Cartesian grid, the correct reference-space normal for cell (j,k,l) is proportional to:

```
( ∂α/∂x · Δx_j,  ∂α/∂y · Δy_k,  ∂α/∂z · Δz_l )
```

where `∂α/∂x ≈ (α(j+1) − α(j−1)) / (x_{j+1} − x_{j−1})`. The code uses `(α(j+1) − α(j−1)) / 2` instead, which is only correct when the stencil width equals the cell width (i.e., uniform spacing). On a stretched grid the normal direction can be wrong, proportional to the local grid non-uniformity.

Dividing by physical dx/dy/dz (as sometimes suggested) is **not** the correct fix — that gives the physical-space gradient, which is inconsistent with the reference-space face-average integrals. The correct fix is to properly account for the mapping between physical and reference coordinates when computing the gradient.

## Impact

- **Uniform grids**: no impact.
- **Mildly stretched grids**: small accuracy loss in MTHINC normal direction.
- **Strongly stretched grids** (e.g., near-wall refinement): potentially significant error in interface orientation, degrading compression quality.

## References

- B. Xie & F. Xiao, JCP 349 (2017) — THINC-QQ (physical-space formulation for unstructured grids)
- F. Xiao et al. (2011), Li & Xiao (2014) — structured-grid MTHINC reference-space formulations

### LINKED ISSUE #1396: THINC/MTHINC: int_comp incompatible with viscosity and surface tension (reconstruction split silently disables compression)
## Summary

When `int_comp > 0` is combined with `viscous = T` or `surface_tension = T`, interface compression is **silently disabled** — no error, no warning, no sharpening.

## Root cause

In `m_rhs.fpp`, the reconstruction path splits into multiple `s_reconstruct_cell_boundary_values` calls with disjoint variable index ranges when viscosity or surface tension is enabled:

```
# Viscous case (Re_size > 0):
call s_reconstruct(vf(1 : cont%end), ...)       ! partial densities only
call s_reconstruct(vf(E   : sys_size), ...)      ! energy + adv (volume fractions)

# Surface tension case:
call s_reconstruct(vf(1   : E-1), ...)           ! cont + mom
call s_reconstruct_first_order(vf(E), ...)       ! energy
call s_reconstruct(vf(E+1 : sys_size), ...)      ! adv (volume fractions)
```

Inside `s_weno` and `s_muscl`, compression is guarded by:

```fortran
if (int_comp > 0 .and. v_size >= eqn_idx%adv%end) then
    call s_thinc_compression(...)
end if
```

`v_size = ubound(v_vf, 1)` for the current slice. In the split paths, no single slice contains both `cont` (needed for density ratios) and `adv` (volume fractions), so `v_size >= eqn_idx%adv%end` is never true and compression never fires.

## Current mitigation (merged in PR #1303)

`check_interface_compression` in `toolchain/mfc/case_validator.py` now prohibits these combinations explicitly:

```python
self.prohibit(int_comp != 0 and viscous,
    "int_comp > 0 is not supported with viscosity (reconstruction is split; compression would be silently skipped)")
self.prohibit(int_comp != 0 and surface_tension,
    "int_comp > 0 is not supported with surface tension (reconstruction is split; compression would be silently skipped)")
```

This turns a silent wrong result into a clear error at validation time.

## What full support would require

To actually support `int_comp` with viscosity or surface tension, the reconstruction in `m_rhs` would need to be restructured so that `cont` and `adv` variables are always reconstructed together in a single WENO/MUSCL call when `int_comp > 0`, allowing `s_thinc_compression` to access both sets of variables at once. This is non-trivial and warrants its own PR.

sbryngelson:
## Summary

Builds on #1303 (MTHINC implementation). Fixes two bugs found during review:

- Closes #1395 — MTHINC interface normals inaccurate on non-uniform grids
- Closes #1396 — `int_comp` incompatible with viscous / surface-tension flows

## Fix 1: MTHINC normals on non-uniform grids (#1395)

The reference-space normal in `s_compute_mthinc_normals` used a uniform-grid assumption (`* 0.5`) that gives the wrong direction when cell widths vary. The correct weighting is:

```
n_ξ ∝ (∂α/∂x) · Δx_j  =  [α(j+1) - α(j-1)] · (x_cb(j) - x_cb(j-1)) / (x_cc(j+1) - x_cc(j-1))
```

On uniform grids this ratio equals `0.5` exactly, so all existing tests pass unchanged.

## Fix 2: int_comp with viscous / surface-tension flows (#1396)

`int_comp > 0` requires full (non-split) reconstruction, but the viscous and surface-tension RHS paths were calling split reconstruction, silently disabling interface compression. Fixed by routing those paths through the full reconstruction branch and adding a runtime `@:PROHIBIT` guard.

## New regression test (`7A1719C6`)

A new golden-file test exercises Fix 1 on a non-uniform grid:

- 2D domain with cosh-based x-stretching, creating non-uniform cell widths
- Circular fluid-2 bubble creates a curved interface with **diagonal normals** (both `nr_x ≠ 0` and `nr_y ≠ 0`) — only diagonal cells are sensitive to the `Δx_j` weighting correction; axis-aligned interfaces always normalize to `±1` regardless of scaling
- Uniform `vel(1)=0.5` advects the bubble so MTHINC reconstruction produces non-trivial fluxes; 129+ distinct values at t=50 confirm genuine interface evolution

## Test plan

- [x] All 4 existing uniform-grid MTHINC tests pass unchanged (`5126B21F`, `4E4FECA9`, `4F3722DB`, `4C4F339C`)
- [x] New non-uniform grid MTHINC test passes (`7A1719C6`)
- [x] `./mfc.sh precheck -j 8` passes
- [x] `./mfc.sh build -t simulation -j 8` compiles cleanly


sbryngelson:
## Automated Code Review

**Summary:** 1 critical issue, 2 important issues.

### Critical Issue

**1. Missing `use m_nvtx` in `m_muscl.fpp` — compile error on all 4 CI compilers** (confidence: 97%)

`src/simulation/m_muscl.fpp` calls `nvtxStartRange("WENO-INTCOMP")` and `nvtxEndRange()` without `use m_nvtx` in its module-level use list. None of the modules it does use (`m_derived_types`, `m_global_parameters`, `m_variables_conversion`, `m_mpi_proxy`, `m_helper`, `m_thinc`) re-export `m_nvtx`. Every other NVTX caller in the simulation source explicitly adds `use m_nvtx`. This will produce an "undefined symbol" compile error.

**Fix:** Add `use m_nvtx` to the module-level use block in `src/simulation/m_muscl.fpp`.

Minor: the NVTX label `"WENO-INTCOMP"` should be `"MUSCL-INTCOMP"` — this is the MUSCL module, not WENO.

---

### Important Issues

**2. `f_log_cosh_diff` can produce `+Inf` for large `ic_beta`** (confidence: 82%)

`src/simulation/m_thinc.fpp`:
```fortran
res = 2._wp*atanh(tanh(a)*tanh(h))
```
`tanh(x) = 1.0_wp` exactly for `|x| >= ~19.1` in IEEE 754 double. When `beta * |n1| >= 38.2` (e.g., `ic_beta = 40`), `atanh(±1.0)` returns `±Inf`, leaking into the Newton solver and solution fields. The default `ic_beta = 1.6` is safe, but no upper-bound validation exists in `case_validator.py` or `m_checker.fpp`.

**Fix (option A):** Clamp the `atanh` argument:
```fortran
t = tanh(a)*tanh(h)
res = 2._wp*atanh(sign(min(abs(t), 1._wp - epsilon(1._wp)), t))
```
**Fix (option B):** Add `@:PROHIBIT(ic_beta > 30._wp, ...)` in `m_checker.fpp` and `case_validator.py`.

**3. No Fortran-side `@:PROHIBIT` for `muscl_order==1` + `int_comp>0`** (confidence: 80%)

`case_validator.py` correctly blocks this combination, but `m_checker.fpp` has no matching runtime guard. A user who writes a `.inp` namelist directly (bypassing Python validation) hits undefined behavior: `v_rs_ws_*_muscl` contains stale/uninitialized data when `muscl_order==1` skips `s_initialize_muscl`, but `s_thinc_compression` still runs.

**Fix:** Add to `m_checker.fpp`:
```fortran
@:PROHIBIT(recon_type == MUSCL_TYPE .and. muscl_order == 1 .and. int_comp /= 0, &
           & "int_comp must be 0 when muscl_order = 1")
```

### Suggestions

- No upper-bound validation for `ic_eps` — `ic_eps >= 0.5` would disable compression entirely for all physical volume fractions; a simple `prohibit(ic_eps >= 0.5, ...)` guard would prevent silent misconfiguration
- Newton solver in `f_mthinc_solve_d` iterates up to 30 steps with no convergence diagnostic — a `@:LOG` on the residual after the loop would aid debugging near-degenerate cases

### Strengths

- Non-uniform grid normal formula `nr_x = (α(j+1) - α(j-1)) * Δx_j / (x_cc(j+1) - x_cc(j-1))` correctly reduces to `0.5*(α(j+1) - α(j-1))` on uniform grids — all existing tests pass unchanged ✓
- `@:ALLOCATE`/`@:DEALLOCATE` pairing for `mthinc_nhat` and `mthinc_d` correct, both gated on `int_comp == 2`
- `case_validator.py` correctly enforces all interdependencies: `int_comp > 0 => num_fluids == 2`, `int_comp == 2 => n > 0`, `int_comp > 0 => muscl/weno_order > 1`
- `GPU_DECLARE(copyin=...)` for `gq3_pts`/`gq3_wts` is the correct pattern for constant quadrature weights
- Normal snapping with `mthinc_align_tol` avoids near-degenerate multi-dimensional quadrature

Issue #1402, opened 2026-05-06
HLL axisymmetric: volume fraction solved in conservative form (missing Kapila source) after PR #800


sbryngelson:
## Summary

PR #800 changed one line in the HLL solver's `cyl_coord` block for the radial (y) direction:

```fortran
! m_riemann_solvers.fpp, HLL cyl_coord block, NORM_DIR == 2
- flux_gsrc_rs${XYZ}$_vf(j, k, l, i) = 0._wp
+ flux_gsrc_rs${XYZ}$_vf(j, k, l, i) = flux_rs${XYZ}$_vf(j, k, l, i)
```

This fixed a real issue (HLL was not applying any cylindrical correction to `α`), but the fix introduced a **different error**: HLL now solves the volume fraction equation in **pure conservative form** rather than the quasi-conservative form required by the Kapila/Allaire 5-equation model.

## The math

The Kapila volume fraction equation is:

```
∂α/∂t + ∇·(αu) = α∇·u
```

In cylindrical coordinates `(r, z)`, the geometric terms cancel:

```
∇·(αu) - α∇·u  =  u_r ∂α/∂r
```

So the physically correct equation is the **material-derivative form**, identical to Cartesian:

```
∂α/∂t + ∂(αu_r)/∂r = α ∂u_r/∂r
```

After PR #800, HLL sets `flux_gsrc(adv) = flux_rs(adv)` and `flux_src(adv) = 0`, giving:

| Term | Value |
|------|-------|
| Flux divergence | `-∂(αu_r)/∂r` |
| Geometric correction | `-(αu_r)/r` |
| Non-conservative source | `0` |
| **Net** | **`∂α/∂t + (1/r)∂(rαu_r)/∂r = 0`** ✗ |

This is the **purely conservative form with no Kapila source**, which will generate spurious pressure oscillations at material interfaces (the original motivation for the quasi-conservative approach).

HLL may appear to work despite this because numerical diffusion masks the interface error, but it is solving the wrong equation.

## Correct HLL fix

The right approach mirrors what HLLC already does:

- Keep `flux_gsrc(adv) = 0` (geometric terms cancel analytically)
- Supply `flux_src(adv)` with a velocity estimate so the non-conservative source `α ∂u_r/∂r` is applied

For HLL, there is no contact wave, but a natural estimate is the HLL-averaged normal velocity:

```
u_HLL = (s_L * u_R - s_R * u_L) / (s_L - s_R)
```

[...]

PR #1411, opened 2026-05-08
Add 2D axisymmetric WENO5 convergence test; fix single-fluid alpha drift; clarify HLL gsrc comment

sbryngelson:
## Summary

While developing a 2D axisymmetric (`cyl_coord=T`, `num_fluids=1`) convergence test, one bug in the simulation source was found and fixed, and a misleading comment in the HLL/LF solver was corrected. The convergence test (WENO5, fitted rate 5.08) is included as a new CI job.

### Bug — Single-fluid `cyl_coord`: alpha drifts each RK stage

For `num_fluids=1` + `cyl_coord=T`, α is trivially 1 everywhere but the HLLC contact-wave speed varies slightly across radial faces, producing a tiny net flux in α over RK sub-stages. This eventually causes a pressure NaN when computing `p = (γ−1)(E − ρ|u|²/2) / α`. Fix: reset `q_cons_ts(1)%vf(eqn_idx%adv%beg)%sf = 1._wp` after the `adv_n` call in each RK3 stage (guarded by `num_fluids == 1 .and. cyl_coord .and. .not. bubbles_euler`).

### Comment correction — HLL/LF void-fraction geometric source

A previous version of this PR incorrectly changed `flux_gsrc(adv) = flux_rs` → `= 0` for HLL and LF, which would have reverted PR #800. This has been corrected. Per the analysis in issue #1402 (comment by ChrisZYJ): `flux_rs(adv)` in HLL/LF is the **numerical diffusion** component (not the transport flux — that lives in `flux_src_rs`), and in cylindrical coordinates any face-flux requires the full geometric divergence `(1/r)∂(rF)/∂r = ∂F/∂r + F/r`. The `flux_gsrc = flux_rs` assignment provides the `F/r` correction for the diffusion. PR #800 was correct; only the comment is updated to document why.

### New convergence case and CI job

- `examples/2D_axisym_convergence/case.py`: density sine wave `ρ=1+0.2·sin(2πx)`, `u_x=1`, `u_r=0`, `p=1` in a `[0,5]×[0,0.5]` cylindrical domain. The domain contains exactly 5 full sine periods, so periodic x-BCs make `ρ(x,T)=1+0.2·sin(2π(x−T))` an exact solution everywhere. This is an exact nonlinear solution to the cylindrical Euler equations: it is a pure entropy wave — pressure stays uniform at `p=1` for all amplitudes because `p=(γ−1)(E−ρ/2)=1` identically, and the r-momentum geometric source cancels exactly (`(1/r)∂(rp)/∂r − p/r = 0` for uniform p).
- `toolchain/mfc/test/run_convergence_axisym.py`: refines `Nz∈{64,128,256}` with `Nr=32` fixed, reads `q_cons_vf1` from `p_all`, averages over r, and asserts fitted WENO5 rate ≥ 4.8. Starting at Nz=64 (12.8 cells/wavelength) avoids the pre-asymptotic regime where WENO-JS underperforms near smooth extrema.
- `.github/workflows/convergence.yml`: new `convergence-2d-axisym` job.

**Observed rates:**

| Nz | L2 error | rate |
|----|----------|------|
| 64 | 8.06e-4 | — |
| 128 | 2.46e-5 | 5.04 |
| 256 | 7.07e-7 | 5.12 |

Fitted rate: **5.08** (need ≥ 4.8) ✓

## Test plan

- [x] `./mfc.sh precheck -j 8` passes (all 6 checks)
- [x] `./mfc.sh build -j 8` compiles clean
- [x] `python toolchain/mfc/test/run_convergence_axisym.py` → PASS, rate 5.08
- [ ] Existing regression suite: `./mfc.sh test -j 8` (no golden files changed)


sbryngelson:
@ChrisZYJ give this a look. I found a bug in axisymm after seeing your comment on my other (now closed) PR. I think.

ChrisZYJ:
@sbryngelson Regarding the axisymmetric single-component alpha drift issue (I believe your scope is axisymmetric HLL and HLLC without bubbles_euler?), my experience is that tiny drifts are expected but should remain at machine precision and shouldn't cause crashes unless amplified by a less stable numerical method or an extreme physical setup, where the alpha=1 clamp can be used (just like mpp_lim for 2-comp). Would you mind sending me your case file so I can check for hidden bugs in the code that could cause alpha to drift more than it should?

sbryngelson:
Confirmed: current master (post #1403's HLLC xi-factor fix) shows **zero** alpha drift for this case — ran the Nz=64 axisym case on master binaries and got max|α−1|=0.0 at all cells at t=T. The root cause was the `ξ_L−1` cancellation that #1403 already fixed. Closing since the fix is in master.

PR #1414, opened 2026-05-09
Anchored Dual-pass HLLD for Hypoelasticity (+ HLLC and interface-consistent HLL)

ChrisZYJ:
## Description

Adds:
1. Hypoelasticity: Anchored Dual-pass HLLD
2. Hypoelasticity: HLLC
3. Hypoelasticity: HLL option (interface-consistent)
4. HLL option (alpha div U) so non-conservative treatment aligns with HLLC

#### Separate HLLD Riemann Solvers
At a glance it might be tempting to combine HLLD MHD with dual-pass hypoelasticity HLLD, but keeping them separate makes the code cleaner and much easier to maintain because:
1. Unlike HLL or HLLC, HLLD is a class of HLLD-type solvers, with formulas and states dependent on the eigenstructure of the governing equations, so the inner states' equations are completely different for MHD vs Hypoelasticity. 
2. HLLD hypoelasticity has a newly developed dual-pass anchored form, making it different from any convenional HLLD Riemann solver. The anchored forms are necessary for the non-conservative hypoelasticity terms, which MHD does not have.
3. MHD and Hypoelasticity deal with completely different physical regimes with different governing equations, and any changes or new physical models added in the future will not apply to both modules at once. 

#### Riemann Source Terms
For the non-conservative terms, unlike the usual governing equations that only need div U i.e. du/dx, dv/dy, dw/dz (alpha div U, K div U, etc.), Hypoelasticity has cross terms like du/dy, so we must also pass those Riemann-consistent traces from Riemann solver to the rhs. (The old Hypoelasticity code with the HLL Riemann solver uses finite difference for non-conservative rhs, which provides enough stability given that HLL smears the interface immediately, so there wasn't a need to pass the du/dy traces before this PR. But that does not work for HLLC/HLLD for Hypoelasticity.)

Also grouped/named the condition branches (with lots of comments within the code):
| Branch | Face quantity read | RHS formula per $\alpha_k$ | K*div(u) velocity source |
|--------|-------------------|---------------------------|--------------------------|
| `adv_src_alpha_iface` | `flux_src_n(dir)%vf(j_adv)` = per-fluid $\Psi_{\alpha_k}$ | $u_\text{cell} \cdot \Delta\Psi_\alpha / \Delta x$ | `nc_iface_vel_n(dir)%vf(1)` |
| `adv_src_vel_iface` | `flux_src_n(dir)%vf(adv\%beg)` = shared $\Psi_u$ | $\alpha_k \cdot \Delta\Psi_u / \Delta x$ | Same `flux_src_n` slot (already $\Psi_u$) |
| `adv_src_none` | — | Skipped (HLLD handles internally) | — |

The derivations, meanings, and usage of the Riemann source variables are not straightforward. I've added some hopefully very helpful notes in `misc/dev_notes` for future developers (or AI agents; directing them to my notes should help them make fewer mistakes with the source terms) in terms of the understanding and derivations for the HLL/HLLC non-conservative fluxes, and their variable mapping for Riemann solvers and RHS. 

#### Backwards Compatibiilty
* All default behaviors preserved exactly (newly added features as options)
  * The only exception is the removal of an incorrect ad-hoc fluids-limit guard that affects only Hypoelasticity HLL
* All existing usage of Riemann and rhs source terms are preserved. No refactor is done to keep the scope of this PR limited (any refactoring would touch most of the existing HLLC functionalities)

### Type of change

- [x] New feature

## Testing

- [x] All tests passed locally on CPU and Nvidia GPU, and on Frontier. 

- [x] Smooth Eigenmode Convergence
image

- [x] Weak Solution Comparison (Rodriguez & Johnsen (2019) §5.3(b))
image

- [x] Weak Scaling on Frontier
image

## Checklist

## AI code reviews

- `/review` — Qodo review


sbryngelson:
very impressive! but we're definitely going to need a Riemann solve refactor before merging this. it's all just too cumbersome.

sbryngelson:
Heads-up on what changed in master since this branch was cut, since the conflicts here are structural rather than mechanical:

1. **#1546 unified the flux arrays to x-only**: the per-direction `flux_rs{x,y,z}_vf` arrays are gone; Riemann code now writes `flux_rsx_vf(${SF('')}$, ...)` with permuted indices. The HLLD hypoelastic anchor block here still emits `flux_rs${XYZ}$_vf` inside its Fypp direction loop, so the y/z iterations reference arrays that no longer exist — that block needs reworking against the new architecture before a merge with master can compile.
2. **#1550 introduced named parameter values + generated namelists**: `riemann_solver` comparisons in master are now `riemann_solver == riemann_solver_hll` etc., and the namelist statement in `m_start_up` comes from a generated include — new parameters get registered in `toolchain/mfc/params/definitions.py` and appear automatically.

We attempted a maintainer-side conflict resolution and got everything mechanical done, but the HLLD flux block (and a couple of smaller physics choices, e.g. the elastic flux adjustment and the `hll_u_interface` advection-source semantics) need your judgment. Happy to share the partial resolution as a reference branch if useful — just say the word.

ChrisZYJ:
Thanks for the update. I'm working on the merge together with a few other changes - will commit soon

sbryngelson:
Following up on my earlier note about #1546 (x-only flux arrays) — the Riemann refactor I mentioned has now landed (#1550–#1556), and since you said you're mid-merge, here is a map of where your touched regions now live on `master` so you can relocate rather than re-resolve:

**`src/simulation/m_riemann_solvers.fpp` is now a ~116-line dispatcher.** The solver bodies were split out, and your hunks map as follows:
- `s_hll_riemann_solver` (your interface-consistent HLL changes) → `src/simulation/m_riemann_solver_hll.fpp`
- `s_hllc_riemann_solver` (your 12 hunks incl. low-Mach/ADC work) → `src/simulation/m_riemann_solver_hllc.fpp`
- `s_hlld_riemann_solver` and your new `s_hypo_hlld_riemann_solver` (anchored dual-pass) → `src/simulation/m_riemann_solver_hlld.fpp`
- `s_populate_riemann_states_variables_buffers`, `s_initialize_riemann_solver`, `s_finalize_riemann_solver`, and the viscous source-flux routines → `src/simulation/m_riemann_state.fpp`; your new `s_finalize_nc_iface_vel` belongs there next to `s_finalize_riemann_solver`
- `s_riemann_solver" (dispatch), `s_initialize/finalize_riemann_solvers_module` → stay in `m_riemann_solvers.fpp` (wire the `riemann_hypo_ADC`/dual-pass dispatch there)

**New parameters must go through the registry now.** Your hand-edited declarations and `m_mpi_proxy` broadcasts for `riemann_hypo_ADC`, `ADC_kappa`, `hll_u_interface`, `hypo_hll_interface_rhs` will conflict: declarations, namelist bindings, and MPI broadcasts are auto-generated from `toolchain/mfc/params/definitions.py` (`_r()` + `_nv()`; physics constraints in `case_validator.py`). Default assignments stay manual in `s_assign_default_values_to_user_inputs`; `GPU_DECLARE` lines for registry scalars live in `m_global_parameters_common.fpp` (declare directives must be in the declaring module). See `docs/documentation/contributing.md`. Derived internal flags (`hypo_nc_*`, `adv_src_*`, `use_nc_iface_vel`) can stay as manual declarations in `m_global_parameters.fpp`.

**Other mappings:** your `m_hypoelastic` split (`*_finite_diff_per_sweep`/`*_iface`/`*_axisym_geom_iface`) still targets `m_hypoelastic.fpp` (unchanged home, but diff against current master — the energy-coupling code there was recently consolidated); `m_checker.fpp`, `m_rhs.fpp`, `m_time_steppers.fpp`, `m_variables_conversion.fpp` keep their homes. Your `SPLIT_LONG_DIRECTIVE` macro patch still applies — `acc_macros.fpp`/`omp_macros.fpp` retain the `$:acc_directive`/`$:omp_directive` emission points.

**Two cautions that still stand:** (1) the x-only flux architecture concern from my earlier comment — everything writes `flux_rsx_vf` with permuted indices now, so the per-direction `flux_rs${XYZ}$_vf` writes in the hypo anchor blocks need rethinking, and that is a design decision I don't want to guess for you; (2) please regenerate your 38 golden files against current master rather than carrying the old ones over — an unexplained golden diff after the refactor would mask exactly the kind of silent regression we're trying to avoid.

Happy to help with the mechanical relocation once you push your merge — ping me here.

sbryngelson:
No movement/response from author.

ChrisZYJ:
@sbryngelson almost done with the merge

ChrisZYJ:
@sbryngelson Hi Spencer, I’ve completed the merge and the PR is ready for review. Would you mind reopening it when you have a chance? Thanks!

ChrisZYJ:
The merge took a while because I first merged #1546, then did a performance refactor of the HLLD path, and had to re-merge onto #1550–#1556. The refactored code is a lot cleaner now though. Thanks for the improvements!

The relocation map was helpful, and I followed it where I could. One thing I wanted to discuss is keeping m_riemann_solver_hypo_hlld.fpp separate from m_riemann_solver_hlld.fpp:

- The infrastructure they share is already factored out into s_populate_riemann_states_variables_buffers, s_initialize/finalize_riemann_solver, and s_compute_speed_of_sound.
- MHD and hypoelasticity are mutually exclusive, and the two solvers are mathematically different apart from the 5-wave structure. They share only like 40 lines of preamble (#:for scaffold, primitive unpack, mixture-property sum) that are common to all solvers, not just HLLD. Almost everything else is different, including different conservative variable vectors, magnetosonic vs elastic wave speeds, conservative vs non-conservative anchored dual-pass, etc.
- That being said, I'm happy to fold both subroutines into m_riemann_solver_hlld.fpp if you'd prefer all the HLLD-type subroutines in one file.

On a side note, I think the two developer's notes in misc/ are pretty useful. Feel free to take a look, and let me know if you'd like them moved somewhere else.

ChrisZYJ:
For Claude Code Review: 1 is false positive. 2 I've added a PROHIBIT to be extra safe

src/simulation/m_riemann_solvers.fpp
```diff
@@ -40,6 +42,16 @@ contains
         integer, intent(in)                                    :: norm_dir
         type(int_bounds_info), intent(in)                      :: ix, iy, iz
 
+        if (hypo_nc_dual_pass) then
```
sbryngelson: This is a second solver-selection path. The other solvers are picked by riemann_solver == NUM in the loop below; the dual-pass HLLD is picked by this separate boolean and returns early. Since hypo_nc_dual_pass is derived from riemann_solver and the elasticity flags, the two paths stay consistent, so this is fine to ship. But it is the point where hypoelasticity splits into three different shapes across the Riemann solvers: inline branches in HLLC and HLL, and this separate module for HLLD. Worth a short comment here, and in dev_notes, that names all three and says why HLLD needs its own path.

src/simulation/m_global_parameters.fpp
```diff
@@ -830,6 +851,35 @@ contains
...
+            else if (riemann_solver == 4) then
+                hypo_nc_dual_pass = .true.
+            end if
+        end if
+
+        ! flux_src: choose exactly one export mode for the NC volume fraction advection term. Developer note: keep exactly one of
+        ! adv_src_alpha_iface, adv_src_vel_iface, and adv_src_none true whenever this logic is modified.
+        !
+        ! HLL Method 1 (alpha-interface): flux_src(adv_idx%beg:adv_idx%end) carries interface alpha_k per fluid.
+        adv_src_alpha_iface = (riemann_solver == 1 .and. .not. hll_u_interface)
```
sbryngelson: These three booleans are mutually exclusive by construction, and the comment just above asks maintainers to keep them that way by hand. A single enum (for example adv_src_mode with values alpha_iface, vel_iface, none) would make the two-true-at-once state impossible and drop the manual rule. Same readability win at every if (adv_src_*) site downstream.

src/simulation/m_rhs.fpp
```diff
@@ -82,6 +82,14 @@ module m_rhs
 #endif
     !> @}
 
+    type(vector_field), allocatable, dimension(:) :: nc_iface_vel_n
+    $:GPU_DECLARE(create='[nc_iface_vel_n]')
+
+    !> hat_R-pass interface velocities for fused dual-pass HLLD: the axisymmetric geometric source consumes both passes' face
+    !! velocities after the direction loop, so the hat_R values need their own field set (hat_L uses nc_iface_vel_n).
+    type(vector_field), allocatable, dimension(:) :: nc_iface_vel_hatR_n
```
sbryngelson, line 92: This _hatR field set is a near copy of nc_iface_vel_n, and the dual pass also adds _hatR copies of the finalize routines in the hypo HLLD module. The left-anchored vs right-anchored distinction reads like it should be a parameter, not a duplicated field set plus routine. Passing the pass identity in would remove the copy and stop the two from drifting. Separate, larger point: the dual-pass orchestration (allocate hatR, solve, finalize hat_L, assemble, finalize hat_R, assemble again) sits inside an already ~2200-line m_rhs. Good candidate to move into its own routine later.

src/simulation/m_riemann_solver_hypo_hlld.fpp
```diff
@@ -0,0 +1,1123 @@
...
+    use m_global_parameters
+    use m_variables_conversion
+    use m_riemann_state
+
+    implicit none
+
+contains
+
+    !> HLLD Riemann solver resolves all 5 waves for the hypoelastic equations: 1 entropy wave, 2 shear stress waves, 2 fast waves.
+    subroutine s_hypo_hlld_riemann_solver(qL_prim_rsx_vf, dqL_prim_dx_vf, dqL_prim_dy_vf, dqL_prim_dz_vf, qL_prim_vf, &
```
sbryngelson, line 58: This subroutine is about 934 lines. That is actually better than the existing HLLC solver at ~1970, and you already split out the finalize helpers below, which is the right instinct. Not a blocker, and the giant-subroutine pattern is pre-existing here. If you want to set a better example, the left/right state setup and the per-wave-region blocks are natural helper boundaries.

src/simulation/m_riemann_solver_hypo_hlld.fpp
```diff
@@ -0,0 +1,1123 @@
...
+            $:END_GPU_PARALLEL_LOOP()
+        end if
+
+    end subroutine s_finalize_riemann_solver_hatR
+
+    !> Copy the hat_R-pass hypo interface velocities of the fused dual-pass HLLD solve from Riemann-space buffers to physical-space
+    !! output arrays. Mirrors s_finalize_nc_iface_vel for the nc_iface_vel_hatR_rs* buffers.
+    !! @param nc_iface_vel_vf Output: physical hat_R velocity components at interfaces
+    !! @param norm_dir Sweep direction (1=x, 2=y, 3=z)
+    subroutine s_finalize_nc_iface_vel_hatR(nc_iface_vel_vf, norm_dir)
```
sbryngelson: This is a copy of s_finalize_nc_iface_vel above with a _hatR suffix, and s_finalize_riemann_solver_hatR mirrors the regular finalize the same way. Two copies that have to stay in sync by hand. Consider one routine that takes the pass (left vs right) or the target field set as an argument, instead of the _hatR twin. Same point as the field duplication noted in m_rhs.

sbryngelson (commented):
High-level review of structure and organization, not a line-by-line bug pass. The change is smaller than the 110-file count suggests: most of those files are golden tests and examples. The real surface is about 6 files in src/simulation plus the global params and the dev notes.

What works, keep it:
- The new HLLD solver gets its own module, which matches the existing one-module-per-solver layout. Your reasons for not fusing it with MHD HLLD are correct. Leave them separate.
- The dev_notes derivations are good. This math is hard to recover later. Keep writing these.
- The adv_src_* and hypo_nc_dual_pass flags are derived once in m_global_parameters, not raw user inputs. That is the right place for them.

Main point: hypoelasticity now enters the Riemann solvers in three different shapes.
- HLLC: inline "if (hypoelasticity)" branches inside the existing ~1970-line s_hllc_riemann_solver.
- HLL: similar inline branches.
- HLLD: a separate module plus a separate hypo_nc_dual_pass selection path.
Each choice on its own is fine. Together, anyone asking "how is hypoelasticity handled at interfaces" has to read three different shapes of code. I would not refactor HLLC in this PR (you scoped that out, correctly), but add a short comment at each site, and in dev_notes, that names all three and says why HLLD needs its own path.

Two cleanups that are cheap and in scope:
1. Collapse the _hatR duplication in the dual pass. The second pass duplicates whole field sets (nc_iface_vel_n and nc_iface_vel_hatR_n) and finalize routines (the regular ones and their _hatR twins). The pass identity, left-anchored vs right-anchored, should be an argument, not a copied routine. This roughly halves the dual-pass code and stops the two copies from drifting apart.
2. Replace the three adv_src_* booleans with one enum. They are mutually exclusive by construction, and the code already asks maintainers to keep them that way by hand. An enum makes the "two true at once" state impossible and drops the manual rule.

Two notes for later, mostly out of scope:
3. m_rhs.fpp is already ~2200 lines. The dual pass deepens it: allocate hatR fields, solve, finalize hat_L, assemble RHS, finalize hat_R, assemble RHS again. That orchestration would be better in its own routine. Not required here, just flagging it so it does not harden.
4. The 900 to 2000 line solver subroutines are a pre-existing pattern, not something you introduced. Your new solver at 934 lines is actually better than HLLC because you split out the finalize helpers. Splitting the left/right state setup and the per-wave blocks into helpers would set a good example, but it is not a blocker.

Net: the hard parts, the module boundary and the documentation, are right. Before this leaves draft I would do (1) parameterize the dual pass instead of copying _hatR, and (2) make adv_src_* an enum. The rest is comments or future debt. Inline notes below.

src/simulation/m_riemann_solvers.fpp
ChrisZYJ: Added comments and explained clearly. Also added a corresponding note in misc/dev_notes and back references at HLL/HLLC sites

src/simulation/m_global_parameters.fpp
ChrisZYJ: Added adv_src_mode enum as you suggested and dropped the manual rule.

src/simulation/m_riemann_solver_hypo_hlld.fpp
ChrisZYJ: Agreed that it's not a blocker here

src/simulation/m_riemann_solver_hypo_hlld.fpp
ChrisZYJ: Used a template to factor out s_finalize_nc_iface_vel and _hatR. Kept s_finalize_riemann_solver_hatR as it's a subset of the shared finalize, without flux_src, gg3 gsrc.

src/simulation/m_rhs.fpp
ChrisZYJ: Used a template to factor out the _hatR routines and removed the duplicated parts. I kept nc_iface_vel_hatR_n as it's not a copy but a separate storage per pass. The axisymmetric geometric source needs to run after the direction loop and needs both passes' velocities, so we can't overwrite the hat_L values before the source uses it.

ChrisZYJ:
Made all the requested changes other than a few minor caveats. I've explained all the changes in the comment replies

sbryngelson:
@ChrisZYJ - pushed two correctness fixes to this branch (bb0090e8), both flagged by the Claude review:

1. `m_riemann_solver_hllc.fpp`, the `NORM_DIR == 2` cyl + hypoelastic geometric-source block: the `eqn_idx%cont%end + dir_idx(1)` flux read used literal `flux_rsx_vf(j, k, l, ...)` while the LHS and every sibling assignment use `${SF('')}$`. For the y-sweep `SF('')` expands to `(k, j, l)`, so the literal indices read a transposed element - wrong geometric-source momentum flux on 2D/axisymmetric hypoelastic HLLC runs. Switched it to `flux_rsx_vf(${SF('')}$, ...)`.

2. `m_checker.fpp`: `G1_eff`/`G2_eff` use `fluid_pp(1:2)` for all hypoelastic Riemann solvers, but only HLLD (`riemann_solver == 4`) had the `num_fluids == 2` guard. Added the matching guard for HLL/HLLC (`riemann_solver == 1 .or. == 2`) so `num_fluids > 2` is rejected rather than silently dropping fluids 3+ from the bulk modulus.

Both compile.

ChrisZYJ:
@sbryngelson Many thanks for the changes!

Speaking of point 2, there's an important, subtle bug outside of hypoelasticity: "num_fluids > 2 .and. alt_soundspeed" is not prohibited in the checkers, which silently computes the wrong "K-term". We should add a checker to enforce that alt_soundspeed is only used for 2 components (alt_soundspeed is only for the 5-equation model). I can open a separate PR for that.

sbryngelson (dismissed):

sbryngelson (commented):
## High-level code-quality pass

This PR is large (~12k additions), but most of that is auto-generated `tests/*/golden-metadata.txt` fixtures (~40 new test dirs) — the hand-written `src/` diff is closer to ~2,850 lines. Overall the new code is well-documented for its complexity (good WHY-comments, sensible naming, the dev-notes markdown files are legitimate cross-file architecture docs, not compensating for opaque code). The points below are meant to help trim the remaining LOC/complexity without touching the numerics, plus one real correctness gap.

### 1. Biggest lever: `s_hypo_hlld_riemann_solver` doesn't use the codebase's own decomposition tool

`src/simulation/m_riemann_solver_hypo_hlld.fpp:21-965` is a single ~945-line subroutine with an ~800-line GPU-parallel loop body (147 privatized scalars) — over `contributing.md`'s own soft guideline (≤500 lines/subroutine, ≤1000/file; this file is 1090). MFC already has a working, GPU-offload-safe pattern for factoring per-cell math out of parallel loops: `$:GPU_ROUTINE(parallelism='[seq]')` helper subroutines (used for `s_compute_speed_of_sound`, `s_compute_pressure`, etc.), and this file already *calls* `s_compute_speed_of_sound` that way (lines 301-304) — but never applies the same pattern to its own star-state algebra (~lines 556-630) or output-permutation logic (~lines 822-880). Extracting those into `[seq]`-parallel helpers would roughly halve the loop body with no offload-capability cost.

(Context: this isn't a new anti-pattern — `s_hllc_riemann_solver` is already ~1970 lines pre-existing — so this PR extends existing tech debt rather than introducing a new category of it. But it's the single biggest opportunity to cut LOC here.)

### 2. ~150 lines of duplicated small physics formulas — safe, mechanical extraction

Confirmed identical across 3-4 files:
- Hypoelastic energy-correction floor/guard formula: `m_riemann_solver_hll.fpp:347-360`, `m_riemann_solver_hllc.fpp:308-320` and `:1416-1428`, `m_riemann_solver_hypo_hlld.fpp:284-294`
- Wave-speed estimate (Rodriguez et al. 2019 formula): `m_riemann_solver_hll.fpp:396-403`, `m_riemann_solver_hllc.fpp:378-389` and `:1489-1500`, `m_riemann_solver_hypo_hlld.fpp:306-309`
- `s_finalize_riemann_solver_hatR` (`m_riemann_solver_hypo_hlld.fpp:1030-1088`) duplicates the existing generic finalizer in `m_riemann_state.fpp:1034-1183` almost verbatim, instead of using the fypp-templating approach this same file already uses for the analogous `s_finalize_nc_iface_vel${SUFFIX}$` (lines 974-1020).

Pulling these into shared helper functions/a template would save ~100-150 lines with no design risk.

### 3. Real gap, not just style: `mhd .and. hypoelasticity .and. riemann_solver==4` is never rejected

`m_checker.fpp` only prohibits HLLD (`riemann_solver==4`) when *neither* `mhd` nor `hypoelasticity` is set — never when *both* are set simultaneously. In that combination, `hypo_nc_dual_pass` becomes true and `m_riemann_solvers.fpp` unconditionally routes to `s_hypo_hlld_riemann_solver`, which has no reference to `mhd`/`Bx`/`By`/`Bz` anywhere — magnetic-field physics is silently dropped with no error, and the existing MHD HLLD path is simply never reached. Suggest adding `@:PROHIBIT(riemann_solver == 4 .and. mhd .and. hypoelasticity, "HLLD does not support combined MHD and hypoelasticity")` alongside the existing checks in `s_check_inputs_hypo_branch`.

### Lower priority / follow-up material

- `m_rhs.fpp` threads the new `adv_src_mode` through 6 separate subroutines (`s_initialize_rhs_module`, `s_compute_rhs`, `s_compute_directional_rhs`, `s_compute_advection_source_term`, `s_add_directional_advection_source_terms`, `s_finalize_rhs_module`) rather than resolving the mode once and passing it down — some shotgun surgery, though consistent with that file's existing (non-templated) style.
- `m_riemann_solvers.fpp`'s own comment (lines 42-49) states hypoelasticity now enters the Riemann layer in "three distinct code shapes" (HLL/HLLC inline branches vs. HLLD's separate fused-dual-pass module). A future 4th hypoelastic solver variant would face the same scattered integration choice — worth unifying behind one seam before that happens, not a blocker for this PR.
- The three mutually-exclusive hypo-mode booleans derived in `m_global_parameters.fpp` (`hypo_nc_finite_diff`/`hypo_nc_interface`/`hypo_nc_dual_pass`) could be one enum, matching the `adv_src_mode` pattern already used elsewhere in the same file — inconsistent style for an equivalent problem, not a bug.

None of the above is a blocker — the code is correct and well-tested per the PR description. These are opportunities to reduce the diff's LOC/maintenance footprint using patterns already established elsewhere in this codebase, plus one validation gap worth closing before merge.

---
*Generated with the help of Claude Code.*

ChrisZYJ:
And thanks @sbryngelson for the suggestions! I addressed the following cleanup items:
  - Added checker + Python validator guard for `mhd + hypoelasticity + HLLD`.
  - Deduped repeated hypo elastic-energy and HLL/HLLC wave-speed formulas via Fypp inline macros.
  - Fixed an HLLC 2D-axisym elastic-energy bug: HLLC now doubles only `shear_indices`, matching `m_variables_conversion`.
  - Kept the intrusive `s_hypo_hlld_riemann_solver` helper extraction as follow-up scope. The solve kernel has been optimized  for Frontier, and prior profiling showed it is register/occupancy-sensitive, so moving large blocks behind helpers could be done separately with performance profiling.

  Also fixed the spelling CI failure: `docs/documentation/case.md` now uses `patch(es)` instead of `patche(s)`. I think
  this was exposed by a newer `typos` version on CI.

ChrisZYJ:
@sbryngelson If this looks good after the CI rerun, would it be possible to merge this PR soon? I’m happy to continue improving the code in follow-up PRs after this lands. The branch has already gone through several upstream re-merges and CI/review cleanup rounds, so merging the core implementation now would make future cleanup work smaller and easier to review. Thank you!

sbryngelson:
I'm no longer merging messy/bloated/WIP code on the promise that it will be fixed later. I've been burned on this far too many times (perhaps not by you), and it only ratchets up, not down.

Please clean your code of smells, refactor into nice helpers, and minimize SLOC while maintaining speed if you want the PR merged. Thanks.

ChrisZYJ:
Okay. I'll refactor the solver into helpers according to your reviews, and minimize SLOC as you requested. Will post the changes here soon.

ChrisZYJ:
I've done the refactor. Here is what changed:

- Factored the 25 repeated per-component wave-fan flux folds into small Fypp component tables driving one shared template.
- Factored the 12 outer-wave stress star-state updates into one formula family with a table over component, sign, and elastic coefficient.
- Used a single definition for the per-component HLL flux formula in both the degenerate-fan fallback and the ADC blend.
- Factored `s_finalize_riemann_solver_hatR` and the generic finalizer into one template in `m_riemann_state.fpp`, so they will not diverge in the future.
- Used a single `[seq]` helper `f_hlld_wave_zone` for the five-wave fan-zone selection, now shared by the flux fold, the NC face-velocity export, and the axisymmetric face-state pick.
- Changed the three mutually exclusive `hypo_nc_*` booleans into one `hypo_nc_mode` enum, following the same pattern as `adv_src_mode`.

This removes 95 lines across the touched files. `m_riemann_solver_hypo_hlld.fpp` drops from 1084 to 968 lines, the main subroutine from 938 to 859, and the GPU loop body from ~820 to ~690.

What is left in the main subroutine is dominated by the per-cell star-state algebra, which appears only once, so there is no duplication left to remove. Extracting it behind an interface would take ~50 scalar arguments or regrouping the privatized locals into derived types, and either would change register allocation on the GPU kernel. Since the goal was fewer lines at unchanged speed, that block stays inline, and the helper extractions went where the interfaces are narrow. 

I've taken care to ensure that most changes do not affect the generated code, and that they have minimal impact on performance. Also merged in the current master.

ChrisZYJ:
One note on CI. The last two runs each failed a single Frontier job at the Fetch Dependencies step, before anything gets built or tested (gpu-omp 1/2, then gpu-acc 2/2). Both failures show the same error on the same runner, frontier-5:

```
error: Failed to install: pandas-3.0.3-cp313-cp313-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl
  Caused by: failed to open file `/tmp/uv-cache-sbryngelson/archive-v0/h07_8qgFw9kMCAsSlH4jZ/pandas-3.0.3.dist-info/METADATA`: No such file or directory
```

Maybe a corrupted uv cache entry on that runner's node? Could you help me re-run the one failed job? All the tests should pass.

sbryngelson:
> One note on CI. The last two runs each failed a single Frontier job at the Fetch Dependencies step, before anything gets built or tested (gpu-omp 1/2, then gpu-acc 2/2). Both failures show the same error on the same runner, frontier-5:
> 
> ```
> error: Failed to install: pandas-3.0.3-cp313-cp313-manylinux_2_24_x86_64.manylinux_2_28_x86_64.whl
>   Caused by: failed to open file ``/tmp/uv-cache-sbryngelson/archive-v0/h07_8qgFw9kMCAsSlH4jZ/pandas-3.0.3.dist-info/METADATA``: No such file or directory
> ```
> 
> Maybe a corrupted uv cache entry on that runner's node? Could you help me re-run the one failed job? All the tests should pass.

https://github.com/MFlowCode/MFC/pull/1630

sbryngelson:
Hi @ChrisZYJ — #1572 (Phase-3 Riemann hot-path decomposition) restructured/removed the exact solver code this PR builds on, which is what put `m_riemann_solver_hll/hllc/lf.fpp` and `m_riemann_state.fpp` into conflict.

I resolved the merge of current `master` into this branch for you. It's pushed to:

**`sbryngelson:pr1414-hypo-hlld-master-merge`**

Its first parent is your current head (`815f1e2f4`), so you can take it as a clean fast-forward if you want — no obligation:

```bash
git fetch https://github.com/sbryngelson/MFC.git pr1414-hypo-hlld-master-merge
git checkout hypo_hlld
git merge --ff-only FETCH_HEAD
git push
```

(Or just `git pull https://github.com/sbryngelson/MFC.git pr1414-hypo-hlld-master-merge`.) It's your PR, so please review the resolution before adopting it — the choices below involve your physics, and I have **not** compiled or run it (fypp preprocesses cleanly and the generated Fortran is block-balanced, but goldens are untested).

### The core tension

#1572 deleted the HLLC/LF hypoelastic-energy blocks as *dead code*, justified by `case_validator.py` restricting hypoelasticity to HLL (`riemann_solver == 1`). This PR lifts that restriction to `[1, 2, 4]` (HLL/HLLC/HLLD), so those blocks are **live again**. The resolution keeps your feature and layers master's refactors underneath.

### Choices made (10 conflict hunks)

| File / region | Resolution | Rationale |
|---|---|---|
| **LF** hypoelastic block | took **master** (deleted) | LF isn't in `[1,2,4]`, so the block is dead on both sides; merged LF is identical to master |
| **HLL** inline energy macro | took **master**'s helper | `s_compute_hypoelastic_interface_energy` (called just after the loop) already reproduces your inline physics *including* the `shear_indices` doubling — keeping the inline macro too would double-count |
| **HLLC** declarations | took **yours** (superset) | your ADC/star-state locals are needed downstream |
| **HLLC** 6-eq & 5-eq private lists | took **yours**, and **added master's `alpha_lim_L/R`** to the 5-eq list | GPU-safe superset; the merged 5-eq body uses both your ADC vars and master's `alpha_lim` arrays |
| **HLLC** 6-eq & 5-eq hypoelastic energy | took **yours** (live feature) | validator now permits HLLC hypoelasticity |
| **HLLC** 5-eq mixture loads | took **master** (`alpha_lim_L/R = ...`) | the auto-merged helper call accumulates mixture props from `alpha_lim`; your inline accumulation would double-count against it |
| **HLLC** flux `else` (not-hypoelasticity) | took **yours** | your `else` holds the full base momentum+energy+elastic flux; master's conflict-side tail alone would leave momentum/energy flux uncomputed |
| **`m_riemann_state`** | **union** | kept master's three shared helpers **and** your templated `s_finalize_riemann_solver` (+`_hatR`); dropped master's now-redundant plain `s_finalize_riemann_solver` decl since your template supersedes it |

### Two things worth your eyes

1. **`hypo_energy_guard` gating.** Your HLLC path uses `@:compute_hypo_elastic_energy` (which honors `hypo_energy_guard`), while HLL now goes through master's helper (which hard-gates on `G > verysmall`). Both are live, and I preserved each side's own choice — but HLL and HLLC now treat the soft-G case differently. You may want them consistent.
2. **Cosmetic:** an unused `Re_max` local survives in HLLC (#1572 had removed it). Harmless; drop at your leisure.

Happy to adjust any of these if you'd prefer a different call. cc @sbryngelson

ChrisZYJ:
@sbryngelson Thanks again for preparing the master merge. I reviewed and adopted the resolution, retained the requested refactor and cleanup, and addressed the remaining energy and validation consistency items, including the permanent per-side near-zero-G guard and thermal/kinetic enthalpy for the base acoustic sound speed.

All CI tests have passed. Could you please take a final look and approve/merge if satisfied? Thanks!

sbryngelson:
bumping this so I get to it sooner than later.

sbryngelson:
Three small robustness items in the anchored HLLD path, from working through the
solver algebra. None of them affects the results in this PR as it stands — the
first two are latent and the third is a guard against future regressions.

**1. The inner wave speeds are not clipped into the outer fan.**

In `m_riemann_solver_hypo_hlld.fpp` the inner speeds are set as

```fortran
sqrtC_NC = sqrt(C_NC)
S_Lstar = S_M - sqrtC_NC/rhoL_star
S_Rstar = S_M + sqrtC_NC/rhoR_star
```

and the only degeneracy test beforehand is

```fortran
degenerate = (abs(denomA) < verysmall .or. &
              .not. (S_L - verysmall <= S_M .and. S_M <= S_R + verysmall))
```

which checks the contact against the outer fan but never the inner waves against
it. `C_NC` is built from the *anchor* impedance while `rhoL_star`/`rhoR_star` come
from the reconstructed face state, so where the anchor cell and the face state
hold different materials — i.e. at every one-cell interface — the ratio can
exceed `c_f`. Sweeping the material table, a water face state anchored on a
stiffer solid gives `sqrtC_NC/rho_star` ≈ 2670 m/s against `c_f` ≈ 1625 m/s, so
`S_Lstar < S_L`.

The failure is quiet rather than loud. `f_hlld_wave_zone` classifies the
interface with an ordered `if`-chain (`0 <= S_L`, then `0 <= S_Lstar`, …), so an
inverted ordering just makes the outer-star wedge unreachable and the flux is
taken from the next wedge. No NaN, no crash, just a flux from the wrong region of
the fan. Clamping `S_Lstar` below by `S_L` and `S_Rstar` above by `S_R` fixes it
in two lines, and it subsumes the `C_NC -> 0` repair, since collapsing onto `S_M`
is the limiting case of the clamp.

None of the material pairs in the current example cases trigger this, so it is
latent, not active.

**2. The degenerate-shear guard is dimensional.**

```fortran
C_NC = rho_hat*(G_hat + tau_nn_hat)
if (C_NC < verysmall) then
```

`C_NC` has units of kg·Pa·m⁻³ and is ~1.8e12 for epoxy, so comparing it against
an absolute floating-point tolerance carries no fixed relation to the material.
The risk is that the guard is too *permissive* rather than too strict: a
near-fluid cell from volume-fraction undershoot can sit at `C_NC` ~ 1e-6 in SI,
far above `verysmall`, so the guard stays off while `1/sqrtC_NC` ~ 1e3 amplifies
the double-star states, which go as `dtau/(2*sqrtC_NC)`. Comparing against the
acoustic modulus instead, e.g. `C_NC < eps*rho_hat*gamma_hat*(p_hat + pi_hat)`,
or gating on `c_s/c_f`, makes it scale-free. One line.

**3. A regression test with a sharp criterion.**

The reason HLLD beats HLLC at a shear-bearing contact is that its momentum and
energy fluxes carry the *same* tangential star state — `F[rho v] = -tau**` and
`F[rho E] = -v** tau**` — so the kinetic term cancels the velocity-linear part of
the energy update identically. HLLC's carry different weights, `(1-beta)` in the
momentum flux against `beta` in the energy flux, and the mismatch is
first order in the shear jump.

That gives a test with a much sharper pass criterion than an eyeball comparison:
set up a static contact (`u = 0`, `tau_xx = 0`, `p_T` uniform, `tau_xy` jumping),
where the exact solution has `u = 0` and `p = p_0` so any pressure is spurious,
then sweep the shear jump over a few doublings and fit the order. HLLD should
come out at 2, HLLC at 1. In a 1D reference implementation the fitted orders are
1.99–2.00 and 1.00–1.02 respectively, clean enough that asserting 2.0 ± 0.1 for
HLLD would catch any future break of that flux pairing — a refactor, a GPU port,
a new reconstruction — that a smoke test would miss.

Related, for the docs: since the advantage is this pairing, `riemann_hypo_ADC`
blending toward HLL trades it away. Probably worth a sentence at the option so
that is explicit rather than implicit.

ChrisZYJ:
I've made the changes as requested:

### 1. Inner-fan ordering

The raw `S_Lstar` and `S_Rstar` values are now checked against `[S_L, S_R]` for each anchor. I initially tried clamping them, but a wave speed and its star-state jump are coupled through the Rankine–Hugoniot relation. Changing only the speed would make that wave internally inconsistent.

If the five-wave ordering is invalid, the complete anchored contribution now uses the existing HLL fallback, or the corresponding one-sided physical flux when the outer fan does not straddle zero. The exported face state and NC interface velocity use the same fallback-safe wave zone, so they cannot sample the invalid five-wave fan. The valid HLLD path is unchanged, while the exact or cancellation-level zero-shear case remains a separate finite HLLC-like collapse.

Added regression test `C531DC93` - creates an anchor/face mismatch that distinguishes the two policies. HLL fallback gives a measured one-step transverse-momentum $L_\infty$ change of `0.001`; evaluating the rejected speed-clipped five-wave fold gives approximately `0.0084515`.

### 2. Dimensionless shear-degeneracy guard

I replaced the former absolute `C_NC` threshold with a cancellation-relative criterion:

```text
G_eff   = G_hat + tau_nn_hat
delta_G = 4*epsilon(1._wp)*(abs(G_hat) + abs(tau_nn_hat))
```

When the scale is nonzero, this is equivalent to testing the dimensionless signed cancellation ratio `G_eff/(abs(G_hat) + abs(tau_nn_hat))` against `±4*epsilon(1._wp)`. The result is independent of the units used for stress.

The solver classifies the state as follows:

- Resolved positive `G_eff`, with positive representable `rho_hat*G_eff`: retain the HLLD shear subfan, subject to the independent fan-ordering check.
- `-delta_G <= G_eff <= delta_G`: use the finite HLLC-like zero-shear limit.
- Materially negative or NaN `G_eff`: reject the non-real shear subfan and use HLL fallback.
- Positive `G_eff` for which `rho_hat*G_eff` underflows to zero: use the numerical zero-shear limit; nonpositive or NaN density falls back.

The factor four is a small conservative allowance for the two mixture products, their accumulation into `G_hat`, and the final signed sum with `tau_nn_hat`.

I did not use an acoustic-relative cutoff. This guard is intended to detect cancellation and loss of floating-point representability, not to classify the material. Scaling by the acoustic modulus would eventually classify every sufficiently soft but physically valid solid as fluid-like and suppress a resolved shear subfan, even though a small positive shear stiffness is not singular.

Volume-fraction admissibility also remains separate. MFC already handles reconstructed-state admissibility through `mpp_lim`; solver-local clipping or an additional ad hoc alpha threshold would modify diffuse-interface states and could compromise consistency and accuracy.

Added 2 regression tests to cover the changed outcomes:

- `10CCE589` verifies that a small but resolved positive stiffness retains HLLD.
- `A96DA754` verifies HLL fallback for materially negative effective stiffness.

### 3. Shear-contact regression

I added `Convergence -> HypoShearContact -> {HLLD, HLLC}`. It holds the grid fixed, sets the initial normal velocity to zero, retains a nonzero uniform tangential velocity, and sweeps the `tau_xy` jump over five amplitudes spanning four doublings. The diagnostic measures the baseline-subtracted pressure response after one step. The exact tangential evolution preserves $u=0$ and $p=p_0$, so this pressure response is purely numerical.

The fitted orders are:

- HLLD: `2.000` (required `2.0 ± 0.1`)
- HLLC: `0.998` (required `1.0 ± 0.2`)

The HLLC leg is intentionally asserted as a control. If the uniform tangential velocity were accidentally removed, both solvers would appear quadratic and the HLLD-only test would lose its discriminating power.

### 4. ADC documentation

The documentation now states explicitly that ADC trades contact and shear resolution for HLL robustness and is default-off/recommended-off.

I also corrected the documented `ADC_kappa` direction. It scales the reference jump in the sensor denominator, so smaller values produce more HLL blending, while larger values keep the flux closer to HLLC/HLLD.

ChrisZYJ:
The one failed documentations CI is unrelated to this PR - it's Lychee link checking for a file unchanged during this PR

ChrisZYJ:
PR should be fully ready after just a few more minor bug fixes / improvements. Will update very soon.

sbryngelson (dismissed):

ChrisZYJ:
Thanks for approving the PR! All CI checks have passed, and I’ve just double-checked everything again and think it’s ready to merge. Could you help merge it once you’re happy with it? Thanks!

sbryngelson:
benchmark was suspiciously slow so i reran it to so if it's noise or code issues.

sbryngelson:
still slow

ChrisZYJ:
I'm looking into it right now. Will update soon.

ChrisZYJ:
The one failed test looks like a filesystem copy failure when installing syscheck. Could you help to rerun it? Also, the HLLC GPU performance regression should now be resolved. Could you rerun the Benchmark workflow on the latest commits too? Thank you!

ChrisZYJ:
@sbryngelson Could you please rerun the benchmark when you have a moment? Thanks!

sbryngelson:
when i have a moment i will
________________________________

sbryngelson (approved):

sbryngelson (dismissed):

PR #1421, opened 2026-05-11
Add runtime call tracing

sbryngelson:
## Summary

Adds an opt-in runtime call tracing mode for MFC Fortran targets.

- instruments generated Fortran sources during trace builds
- adds a lightweight trace runtime that resolves symbols and call sites
- traces the local middle grid point by default during `run --trace`
- writes trace output to `trace.log` in the case directory and mirrors to stdout only for terminal runs
- keeps build-time `--trace` in the build lock while making run-time `--trace` a separate non-persisted run option
- supports GNU, NVHPC, and IntelLLVM CPU trace builds

## Validation

- `./mfc.sh precheck -j 12`
- parser sanity check: `build --trace` sets build config `trace=True`; `run --trace` sets `runtime_trace=True` without changing build config `trace`
- GNU/NVHPC CPU trace runs previously verified `trace.log` generation and no duplicated trace lines in redirected stdout
- IntelLLVM 2024.2.1 CPU/no-MPI trace build with `ifx/icx`: `pre_process` and `simulation`
- IntelLLVM short Sod run: exit 0, `trace.log` generated with 1928 trace lines, redirected stdout had 0 trace lines

## Notes

This is draft because the runtime tracing implementation is intentionally broad and may need review around compiler support, trace volume, and the generated-source instrumentation heuristics.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ option(MFC_OpenMP        "Build with OpenMP"                                 OFF
 option(MFC_GCov          "Build with GCov"                                   OFF)
 option(MFC_Unified       "Build with unified CPU & GPU memory (GH-200 only)" OFF)
 option(MFC_Fastmath      "Build with -gpu=fastmath on NV GPUs"               OFF)
+option(MFC_TRACE_CALLS   "Build with generated runtime call tracing"          OFF)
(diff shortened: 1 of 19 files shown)


sbryngelson:
maybe i'll come back to this at some point

PR #1447, opened 2026-05-19
ci: add Spack install + smoke test workflow

sbryngelson:
## Summary

- Adds `.github/workflows/spack.yml` that runs on every push, PR, and weekly schedule
- Installs MFC via `spack install mfc` (all default variants: +mpi +pre_process +simulation +post_process)
- Verifies binaries (`pre_process`, `simulation`, `post_process`, `mfc` shim) are present and executable
- Runs a real dry-run case (`1D_sodshocktube`) through the full toolchain via the installed `mfc` shim

Uses the spack buildcache and `actions/cache` on the spack store for fast subsequent runs. Pinned to `sbryngelson/spack-packages:add-mfc` until the mfc recipe is merged into upstream `spack/spack-packages`.

## Test plan

- [ ] Confirm CI passes on this PR
- [ ] After `spack/spack-packages` merges the mfc recipe, update `packages_repository`/`packages_ref` to point at upstream


sbryngelson:
Parking this for now. The recipe is ready but the upstream spack/spack-packages PR has uncertain ROI at the moment. Will revisit if a concrete use case arises.

PR #1461, opened 2026-05-30
Coverage-based test selection (done right) — shadow mode

sbryngelson:
## What

Re-introduces execution-coverage-based test selection (`./mfc.sh test --only-changes`), replacing the gcov coverage-cache removed in #1460. It ships in **shadow mode** (PR jobs print what they *would* select but still run the full suite); enforcement is a separate later change, gated on shadow evidence.

Design spec: `docs/superpowers/specs/2026-05-29-coverage-test-selection-design.md`. Plan: `docs/superpowers/plans/2026-05-29-coverage-test-selection.md`.

## Why this one is sound where the old one wasn't

The old cache rotted invisibly (UUID-keyed, never auto-refreshed, silent fallback). The post-mortem established that **only execution coverage is sound** — static labels miss transitive side effects. This rebuild fixes the root causes:

- **Keyed by `param_hash`** (SHA-256 of the test's full resolved params), not `CRC32(trace)` — stable to cosmetic edits, invalidates on real behavior changes, decoupled from the golden-file UUID.
- **Single authoritative committed map** (`tests/coverage_map.json.gz`), refreshed by a bot on master — freshness is a git-visible fact, not an evictable cache.
- **Loud anti-rot:** a scheduled `coverage-health` workflow fails red if the map is stale/under-covers (kept off the PR path so it can't wedge PRs).
- **Two invariants enforce everything:** (1) the selector may only *over*-include, never under-include; (2) staleness is always loud.

## The conservative ladder (soundness)

A test runs if any rung holds: changed-file detection failed → all · macro/codegen/build/`src/**/include/` input → all · changed `.f90`/`.f` → all · changed `.fpp` covered by zero tests → all · test param_hash absent → run it · coverage overlap → run it · else skip. Every uncertainty resolves to *run*.

## Validated end-to-end against a real gcov map (574 tests)

| change | selected |
|---|---|
| bubbles feature | 101/610 |
| riemann (core) | 589/610 |
| macro / include | 610/610 (run all) |
| docs only | 0/610 |

## Known gcov caveat (handled)

gcov rolls `#:include`'d `.fpp` into the parent compilation unit, so include files aren't reliably attributed. **Closed by rule:** any `src/**/include/` change forces a full run, so the attribution gap can never cause under-inclusion. (Map built on macOS gfortran 15.2; CI's Linux build may attribute differently — the map rebuilds there, and the include-rule keeps it sound regardless.)

## Test plan
- [x] 33 unit tests for the selection logic (param_hash, the ladder, changed-file parsing, health)
- [x] Full toolchain suite green
- [x] Real `--gcov` build + `--build-coverage-map` over all tests (collector validated at runtime)
- [x] End-to-end selection verified against the committed map
- [ ] **Shadow-mode CI evidence** that selection never under-selects (this PR's purpose)
- [ ] Re-verify gcov attribution on Linux gfortran
- [ ] Follow-up PR: self-hosted job wiring + `--select-enforce`


sbryngelson:
Good catch — fixed in f7a692fc. An empty/whitespace `--changed-files` now falls through to git detection (→ `None` → rung 1, run all) instead of returning an empty set (which read as rung 7, skip all). You're right that it's benign in shadow but a silent full-skip under `--select-enforce`; treating it as uncertainty→run matches the soundness invariant. Added a regression test.

PR #1468, opened 2026-05-30
ci(test): enforce coverage-based test selection on PRs

sbryngelson:
## What

Flips the coverage-based test selection from **shadow** to **enforced** on PR runs:
- `test.yml` github (GNU/intel) Test step → `--only-changes --select-enforce`
- `common/test.sh` self-hosted Phoenix/Frontier → `--only-changes --select-enforce`

PRs now run **only** the tests whose recorded coverage overlaps the changed files. The conservative ladder in `coverage.py` falls back to **run-all** for anything risky: unavailable file lists (rung 1), macro/codegen/build inputs (rung 2), hand-written `.f90` (rung 3), and — critically — a changed `.fpp` that **no test covers** (rung 4). The only "run nothing" path is docs/irrelevant-only changes (rung 7).

## Safety / backstops (why this is safe to enforce)
- **Pre-merge full check:** the NVHPC jobs still run `--test-all` on every PR, so the full suite runs on at least one config *before* merge — a selector under-inclusion is caught pre-merge, not just post-merge.
- **Master backstop:** pushes to master run the full suite (`SELECT` empty), never cancelled per the concurrency config.
- **Selection is visible** in each PR's test log (`Coverage selection: ran N/M …`), so pruning can be eyeballed.

## Note
This PR itself touches `test.yml`/`test.sh` (non-`.fpp`) → rung 2 → it runs the **full** suite, so pruning isn't demonstrated here. A follow-up Fortran-only PR will show the enforced prune. Pruning the NVHPC jobs too (for more savings) is a later step once this proves out.


sbryngelson:
Added a second commit simplifying the flags: `--select-enforce` now **implies** selection, so it no longer needs to be paired with `--only-changes` (which was redundant — `--select-enforce` did nothing on its own). 

- `--select-enforce` (alone) → select **and** prune
- `--only-changes` (alone) → shadow (print selection, run full suite)

CI enforce paths now pass just `--select-enforce --changed-files …`. Verified locally: `--select-enforce` alone prunes (81/578 for a bubbles change, no shadow message); `--only-changes` alone still shows the shadow message. 273 unit tests pass.

PR #1477, opened 2026-06-01
Fix numerical Schlieren for single fluid IGR + add case validation guard for schlieren_alpha

Cowsreal:
## Description

For N fluids, IGR tracks the volume fraction of the first N-1 fluids. The part of the code in m_derived_variables.fpp in post-processing did not account for this and writes out a field of 1.0's instead. 

A guard for schlieren_alpha is also necessary since schlieren_alpha(i) is set to dflt_real and the same result is seen if that's the case.

### Type of change

- [x] Bug fix

## Testing

After fix: (IGR, single fluid)

Screenshot 2026-06-01 at 09 49 51

## Checklist

- [X] I updated documentation if user-facing behavior changed

## AI code reviews

- `/review` — Qodo review


src/post_process/m_derived_variables.fpp
```diff
@@ -494,11 +495,23 @@ contains
...
                         do i = 1, eqn_idx%adv%end - eqn_idx%E
                             q_sf(j, k, l) = q_sf(j, k, l) - schlieren_alpha(i)*q_cons_vf(i + eqn_idx%E)%sf(j, k, l)*gm_rho_sf(j, &
                                  & k, l)/gm_rho_max(1)
+                            if (igr) then
+                                alpha_last = alpha_last - q_cons_vf(i + eqn_idx%E)%sf(j, k, l)
+                            end if
                         end do
+
+                        ! IGR stores only num_fluids-1 volume fractions; the last fluid's volume fraction is untracked.
+                        ! For a single fluid this is simply 1.0 everywhere. Without this term, the entire single-fluid
```
sbryngelson: Nit: stray closing paren in "the entire single-fluid Schlieren field)" — should be "the entire single-fluid Schlieren field".

src/post_process/m_derived_variables.fpp
```diff
@@ -494,11 +495,23 @@ contains
...
+                        ! Schlieren field) is dropped, leaving exp(0) = 1 everywhere. Compute that term below.
+                        if (igr) then
```
sbryngelson, line 500: Optional (efficiency/clarity): both `if (igr)` checks (here and the one inside the `i` loop at line 504) test a loop-invariant flag evaluated on every `(j,k,l)` cell. Since `igr` is constant for the whole run, you could hoist the IGR/non-IGR split outside the triple loop — e.g. keep the inner `alpha_last` accumulation only when `igr`, or branch once above the loops. Negligible cost in post-processing, so not blocking, just a readability note.

Also note this new term divides by `gm_rho_max(1)`, which is 0 for a perfectly uniform density field — the same pre-existing risk as line 502, so no regression, just flagging.

toolchain/mfc/case_validator.py
```diff
@@ -1553,14 +1553,24 @@ def check_schlieren(self):
...
+        # value stays at the -1e6 sentinel and produces nonsensical output. Under IGR the last fluid's volume fraction is
+        # reconstructed (not stored), so schlieren_alpha must still be set for all num_fluids components, including the single
+        # fluid of a single-fluid IGR case. The gamma/pi_inf model (model_eqns == 1) does not use schlieren_alpha.
         if num_fluids is not None:
             for i in range(1, num_fluids + 1):
                 schlieren_alpha = self.get(f"schlieren_alpha({i})")
-                if schlieren_alpha is not None:
+                if schlieren_alpha is None:
+                    self.prohibit(
+                        schlieren_wrt and model_eqns is not None and model_eqns != 1,
```
sbryngelson, line 1575: The guard correctly excludes `model_eqns == 1`, which matches the Fortran (the `model_eqns == 1` branch never uses `schlieren_alpha`). One edge case to consider: IGR's adv-index layout is only set up under `model_eqns == 2` (`m_global_parameters.fpp`), but nothing forbids `igr = T` with `model_eqns ∈ {3,4}`. In that (unsupported) combo the new `if (igr)` term in `m_derived_variables.fpp` would double-count the last fluid. IGR is fundamentally a `model_eqns == 2` feature and would likely fail elsewhere, so this is theoretical — but a belt-and-suspenders `self.prohibit(igr and model_eqns is not None and model_eqns != 2, ...)` in `check_igr` would make the assumption explicit.

Minor: if `model_eqns` is unset (`None`), the guard won't fire even with `schlieren_wrt` on; fine in practice since `model_eqns` is virtually always specified.

sbryngelson (commented):
Careful review — this is a clean, correct bug fix. I verified the core logic:

- **Index math** (`m_derived_variables.fpp:501–513`): for IGR, `eqn_idx%adv%end − eqn_idx%E = num_fluids − 1`, so the loop covers `schlieren_alpha(1..N−1)` and the new term correctly applies `schlieren_alpha(num_fluids)` to the reconstructed `alpha_last = 1 − Σαᵢ`. This reproduces exactly what the non-IGR branch computes.
- The `if (igr)` term lives inside the `model_eqns /= 1` branch and IGR only exists under `model_eqns == 2`, so non-IGR paths are untouched.
- Validator default `schlieren_alpha = dflt_real` (−1e6) confirmed, so the new "must be set" guard is well justified, and the `if/else` restructure preserves both original prohibits.
- No regressions: all example cases touching schlieren (`2D_IGR_triple_point`, `2D_triple_point`) set every `schlieren_alpha`; no golden/test cases generate schlieren.

Only minor, non-blocking notes inline (a comment typo, a loop-invariant branch, and an edge-config suggestion). LGTM once the typo is fixed.

sbryngelson (changes requested):
Requesting changes to track the follow-ups from the inline review. The core fix is correct (IGR index math reproduces the non-IGR result exactly, and the validator default `dflt_real` justifies the new guard), so these are small:

**Please address before merge:**
1. Fix the comment typo at `m_derived_variables.fpp:510` — stray closing paren in "Schlieren field)".

**Optional (fine to defer or wave off):**
2. Hoist the loop-invariant `if (igr)` branch out of the `(j,k,l)` triple loop in `m_derived_variables.fpp` (readability only).
3. Consider an explicit `igr ⇒ model_eqns == 2` guard in `check_igr` so the new `if (igr)` Schlieren term can't double-count under an unsupported `model_eqns ∈ {3,4}` config.

See inline comments for details. Happy to approve once (1) is in.

sbryngelson:
Reviewed this carefully — nice catch on the single-fluid IGR Schlieren bug. The fix is correct: under IGR the loop covers `schlieren_alpha(1..N−1)` and the added term correctly applies `schlieren_alpha(num_fluids)` to the reconstructed last fluid (`alpha_last = 1 − Σαᵢ`), which reproduces exactly what the non-IGR branch computes. The case-validator guard is well justified since `schlieren_alpha` defaults to `dflt_real` (−1e6), and no example/golden cases regress.

I've left a few small inline notes (one comment typo to fix, plus a couple of optional hardening/readability suggestions) and marked the review as Request Changes just to track the typo. Everything else is non-blocking — happy to approve once that's addressed.

src/post_process/m_derived_variables.fpp
Cowsreal: Will fix.

src/post_process/m_derived_variables.fpp
Cowsreal: Will fix.

toolchain/mfc/case_validator.py
Cowsreal: Irrelevant, IGR is blocked to run for model_eqn /= 2

sbryngelson:
did you fix issues

Cowsreal:
Yes, I did (1) by removed the stray parenthesis. I also did (2) by separating the IGR and non IGR branch into separate if statement branches.

PR #1524, opened 2026-06-01
Remove airfoil and stl paramters

### LINKED ISSUE #1451: STL and Airfoil Memory Structs
The IB patch data type caries with it a lot of useless information that is not required by most simulation. As such, we are wasting much time moving data around and our page reads/writes deal with data that is not very coalesced.

A refactor is to be introduced that eliminates model and airfoil parameters for IB patches. They will instead be their own data types. Documentation of this refactor is quite important, as it will fundamentally modify how case files are read in with regards to IB patches.

danieljvickers:
## Description

The patch_ib array has become bloated as many variable have been added to this struct, growing each element to 962 bytes. Several of these variables are unused except in niche cases. Two such examples are the STL models and airfoils. Particularly the STL model is responsible for 492 bytes of this usage (over half).

Since we are not MPI communicating the patch_ib elements regularly, it is important to try to optimize their size. This will improve communication time as well as likely providing a noticeable performance benefit. This was done by creating separate smaller data structures that the patches will call out to that actually contain airfoil and STL information. This will make minor modifications to the case file, which other users should be made aware of.

Fixes #1451 

### Type of change

- [x] Refactor

## Testing

Ran test suite.

## Checklist


danieljvickers:
@sbryngelson approval for benchmark?

danieljvickers:
AI commanets have been addressed, btw

sbryngelson (approved):

wilfonba:
@danieljvickers Benchmark results are not looking good, even with NVHPC. A quick look through the diff didn't give me any idea why either.

danieljvickers:
@wilfonba the only test with slow down was Phoenix openACC. It showed 20% slowdowns in pre-process and post-process, and 40% in simulation. Slow down is even for cases without IBM enabled. But Phoenix openMP did not show slow down in any execution times. Because pre-process and post-process are CPU only, that must mean that the binaries being run in the ACC test are identical to the MP test. I think this indicates that the slowdown is just some acute/random failure on the node and not a real slowdown.

I will rerun this for safety, but nothing about the results lead me to think this PR introduces an NVHPC openACC slowdown yet.

danieljvickers:
Reran the case and got back good execution/grind times. It looks like the previous run was just randomly slow.

sbryngelson:
Case                     Pre Process               Simulation   Post Process  
 ────────────────────────────────────────────────────────────────────────────── 
  igr                       Exec: 0.80      Exec: 0.79 & Grind:     Exec: 0.78  
                                                           0.52                 
  ibm                       Exec: 0.79      Exec: 0.74 & Grind:     Exec: 0.87  
                                                           0.49                 
  viscous_weno5_sgb_aco…    Exec: 0.76      Exec: 0.76 & Grind:     Exec: 0.85  
                                                           0.63                 
  hypo_hll                  Exec: 0.79      Exec: 0.78 & Grind:     Exec: 0.81  
                                                           0.58                 
  5eq_rk3_weno3_hllc        Exec: 0.75      Exec: 0.78 & Grind:     Exec: 0.84  
                                                           0.65

sbryngelson (approved):

danieljvickers:
@sbryngelson I thought I checked all cases last night and the only one out of place was Phoenix OpenACC, which reran fine. What is that from?

Edit, looking at the old run: https://github.com/MFlowCode/MFC/actions/runs/26841117410/job/79221540422
I got OpenACC and OpenMP mixed up in my message. my previous comment still holds true that if we see no regression in the CPU case and/or we see no regression in the other GPU code, then regression in pre-process and post-process in other cases is likely just noise in the measurement. I expect it to rerun fine still.

danieljvickers:
@sbryngelson The new benchmark times are all good. Ready to merge.

Issue #1454, opened 2026-05-26
Projection Optimization


danieljvickers:
The current code projects particles one at a time, with a different kernel for each. This is horribly inefficient. We need more parallelism. Explore a refactor that chooses the parallelism based on the particle count and have them both call the same interior checking code.

PR #1531, opened 2026-06-02
patch_ib Memory Copy Reduction

### LINKED ISSUE #1453: Patch Mem Cpy Reduction
Currently, we are copying the ib_patch array back and forth often in the code. This is required because some compute is still performed on the CPU. Move this to the GPU to eliminate the memory copies.

danieljvickers:
## Description

The patch_ib array was being updated on the CPU. This had the downstream effect of more host-device memory copies that strictly required. This PR modified the code to use the GPU copy of patch_ib everywhere and only copy out during file IO. This should remove memory copies and enable parallelism that was not present before.

Fixes #1453 

### Type of change

- [ ] Refactor


danieljvickers:
There was one `$:GPU_UPDATE(host=...)` that I simply could not remove unless I wanted to break another more-important optimization. I have a solution for this, but it requires the changes that will be made from the projection optimization first. Choosing to merge this in its current state, as I removed several copies that should improve performance. Will address the final memory copy in the m_ib_patch.fpp refactor coming from issue #1454

danieljvickers:
All of the AI comments were correct this time and have not been merged.

sbryngelson (approved):

PR #1540, opened 2026-06-05
1538 ib state file per process io

### LINKED ISSUE #1538: IB State File-Per-Process IO
For large cases, the ib state file is very slow to right. There should be a file-per-process option added to the code to support this.

This involves modification o simulation and post-process. Simulation needs an alternative file-per-process path for IB state data. And post process needs an alternate way to import that data.

danieljvickers:
## Description

Integrated IB code with the `file_per_process` flag, which allows IB state and IB marker data to both be written as file_per_process output. This should optimize file IO for large-scale HPC runs with immersed boundaries.

Closes #1538 

- New feature

## Testing

I ran several iterations of the 2D_mibm_shock_cylinder case on GNU and NVHPC compilers using file_per_process enabled and ib_state_wrt. A restart was performed half-way through the simulation (t_step = 1000) to verify that case restarts work as well.

### DIFF
--- a/src/post_process/m_data_output.fpp
+++ b/src/post_process/m_data_output.fpp
@@ -1356,6 +1356,8 @@ contains
         integer, dimension(num_procs)                   :: meshtypes
         integer                                         :: i, ios, file_unit
         integer                                         :: ierr, nBodies
+        integer                                         :: r, nlocal, gbl_id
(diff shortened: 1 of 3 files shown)


danieljvickers:
Addressed AI review comment.

sbryngelson (approved):

PR #1546, opened 2026-06-09
Unify ICPP STL onto the shared IB model path

### LINKED ISSUE #1478: Unify ICPP STL Implimentation
The original version of STL patches was identical for both IB and ICPP patches. The IB STL code in simulation recently was refactored significantly to be compatible with GPU compute and efficient interior edge checking of the IB.  Because of this, there are many duplicate subroutines in the m_models file that do similar and/or near-identical things. These should be unified so that the ICPP patch follows the implementation of the IB patch code.

This involves refactoring the stl code in m_icpp_patches file to follow the new model data structure, populating the models structure in pre_process, and deleting the unneeded lines of code from the m_models file. The result should remove a significant number of lines of code.

kokomoor:
## Description

Unifies the ICPP STL/OBJ patch path onto the shared immersed-boundary (IB) model machinery.
ICPP model patches (geometry 21) now resolve their model via `patch_icpp(i)%model_id` into the
shared `stl_models` table and use the GPU-offloadable winding-number test `f_model_is_inside_flat`, 
identical to IB STL patches. The model is loaded once in pre_process by `s_instantiate_STL_models`
(guard widened to `#ifndef MFC_POST_PROCESS`), with a per-cell bounding-box check that skips the
winding test outside the model.

Removes the duplicated STL machinery the issue targets:
- the old ray-casting inside-test (`f_model_is_inside`, `f_intersects_triangle`, `f_model_random_number`, `t_ray`);
- the now-dead inline `patch_icpp%model_*` params (`model_filepath/scale/translate/rotate/spc/threshold`);
- the write-only `gpu_total_vertices` array (GPU-uploaded every run but never read; its source field was never initialized).

**Behavior/API change:** ICPP STL cases are now configured like IB STL cases, with a set `num_stl_models`,
`stl_models(k)%`, and `patch_icpp(i)%model_id` instead of the old inline `patch_icpp%model_*`
fields. Per-patch `model_rotate` is no longer applied to ICPP STL patches. No example or regression
case used the old inline style.

Closes #1478.

### Type of change
- Refactor

## Testing

- New `3D -> ICPP -> STL` golden test (`examples/3D_icpp_stl_cube`: a `Cube_IBM.stl` imported as a
  geometry-21 ICPP patch); also covered by `foreach_example` as `3D -> Example -> icpp_stl_cube`
  (serial-IO path, reduced mesh). Both goldens were independently verified to contain a centered
  solid cube (fill ratio 1.0, density contrast 1.19 air / 5.95 model).
- Existing IB STL golden tests (`2D/3D -> IBM -> STL`) pass unchanged.
- Built all three targets (nvfortran, CPU) at each step; `./mfc.sh precheck` clean. Rebased onto current `master`.


sbryngelson:
@kokomoor this is good and will get my more full attention today sometime.

sbryngelson:
AI Review:

 1. 2D ICPP STL patches may silently apply nothing (m_model.fpp / m_icpp_patches.fpp)
  The unified path routes geometry-21 patches through f_model_is_inside_flat. The 2D branch (p==0) depends on
  gpu_boundary_edge_count/gpu_boundary_v, populated by s_check_boundary. If that boundary data is zero, f_model_is_inside_flat returns
  fraction=0 everywhere — the patch does nothing, with no warning. The new example (3D_icpp_stl_cube) is 3D only, so the 2D ICPP path is
  untested.
  ⚠️ Caveat: my agents disagreed on the evidence — the git-history and comment agents read s_check_boundary as still being called for p==0,
  while the scoring agent read the diff as removing it. This contradiction is exactly why it landed at 75, not higher. Worth verifying by hand:
  is s_check_boundary still invoked for p==0 in s_instantiate_STL_models? If so, this is likely fine; if not, it's a real silent-failure bug.
  Adding a 2D ICPP STL test would settle it.

  2. Garbled error message (m_check_patches.fpp, s_check_model_geometry)
  "patch_icpp(" // trim(iStr) // ")%model_id=" // trim(midStr) // "must be in [1, num_stl_models]"
  Missing space → prints model_id=0must be in [1, num_stl_models]. Cosmetic, trivial one-character fix.

  Lower-confidence findings (filtered out)

  - gpu_ntrs unallocated for a 0-triangle STL → crash (65): real but requires an empty/corrupt STL; degenerate input. A defensive if
  (allocated(gpu_ntrs)) guard would close it.
  - iStr not set locally in s_check_model_geometry (25): currently correct (the caller sets it), but breaks the pattern every sibling
  s_check_*_geometry follows — fragile if the caller is ever refactored.
  - num_stl_models still tagged {"ib"} in definitions.py (25): tags are metadata-only today, so harmless, but now misleading since ICPP STL uses
  it without IBM.
  - model_rotate dropped for ICPP STL (0): confirmed intentional — the PR description explicitly states per-patch rotation is no longer applied
  and no example relied on it. Not a regression to flag.
  - Duplicate "Transforming model." print, stale doc comments, params%spc dead assignment: pre-existing or cosmetic.

  My take

  This is a clean, well-motivated refactor (dedup ICPP onto the shared IB path). The only thing I'd genuinely push on before merge is #1 —
  confirm the 2D ICPP STL path actually works and add a 2D test, since the new test only covers 3D and that's the path most likely to silently
  misbehave. #2 is a free fix while you're in there.

kokomoor:
@sbryngelson Thank you! and sounds good to me. I'm investigating your agents findings now. If you concur with them, I'll add a 2D test as well.

danieljvickers:
You correctly removed the old subroutines/functions. There are other subroutines that mimic those functinoalities that end `_flat` to denote them as compatible with GPU compute and the associated flattened STL data. Since we do not have duplicate-purpose subroutines here, I would suggest modifying the STL model function and subroutine names to no longer have the ending `_flat`.  It is a minor code cleanliness option, and should be a straight-forward find-replace job.

sbryngelson:
i think a test makes sense (a small/short one) for 2D. and i guess the other issue is like a two character change so no big deal. thanks @kokomoor

kokomoor:
No problem. I'll add the 2D test, fix the error message, and remove the `_flat` endings from the STL model functions and subroutine names, should have a new commit pushed up shortly. 

I had some fun testing out the STL ICPP functionality as well.

I call him "Mach Lion"
https://github.com/user-attachments/assets/f86d3109-0efb-40d7-a26c-4199116ad5c0

sbryngelson:
> No problem. I'll add the 2D test, fix the error message, and remove the `_flat` endings from the STL model functions and subroutine names, should have a new commit pushed up shortly.
> 
> I had some fun testing out the STL ICPP functionality as well.
> 
> I call him "Mach Lion" [github.com/user-attachments/assets/f86d3109-0efb-40d7-a26c-4199116ad5c0](https://github.com/user-attachments/assets/f86d3109-0efb-40d7-a26c-4199116ad5c0)

Nice work @kokomoor. This is paraview, I take it? What kind of workstations/clusters do you have access to? You should be able to run some pretty big cases on a single GPU if the Re and Ma are not setup poorly (normal Ma, high Re / no viscous)

kokomoor:
Thank you again @sbryngelson . I used paraview for interactive viewing per Dan's recommendation, but I got Claude to walk me through a pyvista recording implementation that wound up working really well. I have it in a scratch dir on my local if you're interested.

All of this work was done on my linux box with a Blackwell 5070. I have a 5090 sitting on a shelf that I'll be swapping the 5070 for in the coming weeks, no cluster access at present.

I'll want to take the 5090 through its paces once I have it installed, so I'll definitely be looking for/making some bigger sims to test on it.

All discussed edits are pushed up on my branch, 2D test is showing green, it _should_ be all set.

sbryngelson:
very cool. a lot of work can get done on a 5070, but only at ~6M grid points or so due to limited DRAM (12gb i think). also if it only computes in single precision then it emulates double prec. (slows things down some).

sbryngelson:
also using pyvista is good. if you try `./mfc.sh viz --interactive` you'll get that environment, roughly. I think that's the command anyway. it has a few options that are useful.

sbryngelson (approved):

PR #1547, opened 2026-06-09
Unify geometry IDs for Immersed Boundaries and ICPP Fixes #1543

### LINKED ISSUE #1543: Unify 2D and 3D Patch Geometries
We use various integers to differentite between geometries in 2D and 3D. For example, 2D STL models and 3D STL models are differentiated. Airfoils, and more are also included. Use of the code may be simpler if these patch geometry definitions were unified into a single value.

I propose the following use the same patch geometry values:
1. 2D and 3D STL models
2. 2D and 3D airfoils
3. circles and cylinders (or spheres depending on discretion)
4. rectangles and cuboids

To maintain legacy functionality, we can have the 3D cases support either value.

Riship749:
This PR unifies the redundant 2D and 3D geometry integers into single, dimension-agnostic IDs within the pre-processor for both `patch_ib` and `patch_icpp`.

**Changes:**
* Merged Rectangles (3) and Cuboids (9) -> ID 3.
* Merged 2D Airfoils (4) and 3D Airfoils (11) -> ID 4.
* Merged Circles (2), Spheres (8), and Cylinders (10) -> ID 2.

The unified subroutines now dynamically check for 3D constraints (e.g., Z-centroids and Z-lengths) using the `p > 0` condition. Legacy compatibility is maintained by routing the old 3D IDs directly to the newly unified subroutines. Obsolete 3D-specific subroutines have been removed to reduce code duplication.

**Testing:**
Compiled successfully locally and passed the full `./mfc.sh test` regression suite.

Fixes #1543


src/pre_process/m_check_patches.fpp
```diff
@@ -54,20 +54,14 @@ contains
                 ! Constraints on the geometric initial condition patch parameters
                 if (patch_icpp(i)%geometry == 1) then
                     call s_check_line_segment_patch_geometry(i)
-                else if (patch_icpp(i)%geometry == 2) then
+                else if (patch_icpp(i)%geometry == 2 .or. patch_icpp(i)%geometry == 8 .or. patch_icpp(i)%geometry == 10) then
```
sbryngelson, line 58: **Dimensionality guards for 3D IDs are dropped (high).** Each original check rejected the wrong-dimension case: sphere/cuboid/cylinder all had `p == 0` prohibits, and 2D circle/rectangle had `p > 0` prohibits. The unified checks only *add* constraints under `if (p > 0)` and never reject `p == 0`.

Crucially, **generation was not unified** — `m_icpp_patches.fpp:64-73` still dispatches `geometry == 8/9/10/11` to the dedicated 3D generators, and `:101-104` still generates a 2D circle/rectangle for `geometry == 2/3`. So:
- `geometry = 8` (sphere) in a 2D run (`p == 0`) now passes validation as a "circle", then runs the 3D sphere generator on a 2D grid → garbage/OOB instead of the old clear error.
- `geometry = 2` (circle) in 3D passes the new sphere-style validation (requires `z_centroid`) but still generates a flat 2D circle — so the intended "ID 2 = circle-or-sphere" merge doesn't actually work end-to-end.

Either keep rejecting mismatched dimension/ID combos, or unify generation too so the dimension-agnostic ID picks the generator from `p`. Right now it's neither.

src/pre_process/m_check_patches.fpp
```diff
@@ -147,11 +141,30 @@ contains
...
         call s_int_to_str(patch_id, iStr)
 
-        @:PROHIBIT(n == 0, "Circle patch "//trim(iStr)//": n must be zero")
-        @:PROHIBIT(p > 0, "Circle patch "//trim(iStr)//": p must be greater than zero")
-        @:PROHIBIT(patch_icpp(patch_id)%radius <= 0._wp, "Circle patch "//trim(iStr)//": radius must be greater than zero")
-        @:PROHIBIT(f_is_default(patch_icpp(patch_id)%x_centroid), "Circle patch "//trim(iStr)//": x_centroid must be set")
-        @:PROHIBIT(f_is_default(patch_icpp(patch_id)%y_centroid), "Circle patch "//trim(iStr)//": y_centroid must be set")
+        ! Core checks for all (Circle, Sphere, Cylinder)
+        @:PROHIBIT(f_is_default(patch_icpp(patch_id)%x_centroid), &
+                   & "Circle/Sphere/Cylinder patch "//trim(iStr)//": x_centroid must be set")
```
sbryngelson, line 148: **The `n == 0` guard was dropped (high).** The old circle check had `@:PROHIBIT(n == 0, "Circle patch ...: n must be zero")`, but it's gone here. A circle (`geometry = 2`) in a 1D run (`n == 0`) is now accepted and then generated on a 1D grid.

This looks accidental rather than intended: the IB circle check (`m_check_ib_patches.fpp:75`) and the rectangle check below (`:174`) both still keep their `n`-guard. Please re-add `@:PROHIBIT(n == 0, ...)` here.

src/pre_process/m_check_patches.fpp
```diff
@@ -147,11 +141,30 @@ contains
...
+        @:PROHIBIT(patch_icpp(patch_id)%radius <= 0._wp, &
+                   & "Circle/Sphere/Cylinder patch "//trim(iStr)//": radius must be greater than zero")
+
+        ! 3D-specific checks (Spheres and Cylinders)
+        if (p > 0) then
+            @:PROHIBIT(f_is_default(patch_icpp(patch_id)%z_centroid), &
+                       & "3D Sphere/Cylinder patch "//trim(iStr)//": z_centroid must be set")
+
+            ! If any extrusion length is set, it's a Cylinder. Verify exactly ONE length axis is defined.
+            if (patch_icpp(patch_id)%length_x > 0._wp .or. patch_icpp(patch_id)%length_y > 0._wp &
```
sbryngelson: **Cylinder length validation is weakened (medium).** The old `s_check_cylinder_patch_geometry` required `count([length_x>0, length_y>0, length_z>0]) == 1` (exactly one length), plus a positivity check on any defined length. Gating the check behind `if (length_x > 0 .or. length_y > 0 .or. length_z > 0)` loses two cases:

- **No length set:** a cylinder (`geometry = 10`) with all lengths default passes here as a "sphere", but generation still runs the cylinder generator with the default length `-1e6`.
- **Explicit non-positive length:** `length_x = -5` (or `0`) makes the trigger false, so all length validation is skipped — the old code caught `.not. f_is_default(length_x) .and. length_x <= 0`.

Suggest restoring the original `count(...) == 1` form (it's clearer than the copied nested `f_is_default` expression) and re-adding the positivity check. Same applies to the IB copy of this block.

src/pre_process/m_check_ib_patches.fpp
```diff
@@ -40,20 +40,12 @@ contains
                 @:PROHIBIT(patch_ib(i)%geometry == dflt_int, "IB patch undefined. patch_ib("//trim(iStr)//")%geometry must be set.")
 
                 ! Constraints on the geometric initial condition patch parameters
-                if (patch_ib(i)%geometry == 2) then
+                if (patch_ib(i)%geometry == 2 .or. patch_ib(i)%geometry == 8 .or. patch_ib(i)%geometry == 10) then
```
sbryngelson, line 43: Same dimensionality-guard concern as the ICPP file applies here: the old sphere/cuboid/cylinder/3D-airfoil IB checks each rejected `p == 0`, and that rejection is gone in the unified subroutines. A 3D IB ID (8/9/10/11) used in a 2D run now passes validation but is generated by the corresponding 3D IB generator. Please confirm the intended dimensional contract and reject mismatches (or unify generation accordingly).

src/pre_process/m_check_ib_patches.fpp
```diff
@@ -80,8 +72,24 @@ contains
 
         call s_int_to_str(patch_id, iStr)
 
-        @:PROHIBIT(n == 0 .or. p > 0 .or. patch_ib(patch_id)%radius <= 0._wp .or. f_is_default(patch_ib(patch_id)%x_centroid) &
-                   & .or. f_is_default(patch_ib(patch_id)%y_centroid), 'in circle IB patch ' // trim(iStr))
+        @:PROHIBIT(n == 0 .or. patch_ib(patch_id)%radius <= 0._wp .or. f_is_default(patch_ib(patch_id)%x_centroid) &
+                   & .or. f_is_default(patch_ib(patch_id)%y_centroid), 'in circle/cylilder/sphere IB patch ' // trim(iStr))
```
sbryngelson: Nit: typos in the error messages — `cylilder` here, and `cyilinder` on line 80 (plus the related `cyilders` comment). Also line 80 is missing a space before `trim(iStr)`: `'in 3D sphere/cyilinder IB patch' //trim(iStr)` renders as `...IB patch3`. Add a trailing space inside the string.

sbryngelson (dismissed):
Nice dedup direction — collapsing the near-identical 2D/3D check subroutines is worthwhile. But the unification is applied to the **validation layer only**, while patch *generation* (`m_icpp_patches.fpp`, IB equivalent) still dispatches on the original IDs (2/3/8/9/10/11). That mismatch turns several of the simplified checks into silent validation gaps:

1. **3D IDs no longer rejected in lower-D runs** — the original `p == 0` prohibits on sphere/cuboid/cylinder/3D-airfoil are gone, so e.g. `geometry = 8` in a 2D case passes validation but hits the 3D generator.
2. **ICPP circle dropped its `n == 0` guard** — a circle in a 1D run is now accepted (IB + rectangle checks still keep this guard, so it looks accidental).
3. **Cylinder length validation weakened** — gating "exactly one length" behind "any length > 0" lets a no-length or negative-length cylinder through.

Details inline. The "compiles + `./mfc.sh test` passes" result is expected — these are all gaps in *rejecting bad input*, which a suite of valid cases won't exercise. Worth adding (or confirming) a case that feeds a 3D ID into a 2D run.

Smaller nits (typos / missing space in messages) also inline.

Riship749:
The generation dispatchers in m_icpp_patches and m_ib_patches are now fully unified and handle the legacy IDs natively based on the p > 0 and n > 0 dimensional blocks. I also added the count and n==0 validator checks you requested.
Codecov might still flag the new @:PROHIBIT lines since the standard test suite doesn't feed bad dimension/length inputs to trigger them.

sbryngelson:
Thanks for taking this on, @Riship749 — the direction here was right, and I've pushed a follow-up commit (e91aa2b7) onto your branch that builds on it. A summary of what changed and why:

The main issue was that the unification only existed in the pre-processor's checkers. The simulation side (`m_ib_patches.fpp`, `m_compute_levelset.fpp`, and the moment-of-inertia code in `m_ibm.fpp`) still dispatches on the old strict 2D/3D IDs, so e.g. a 3D case with `patch_ib(1)%geometry = 2` passed all checks but then silently produced **no immersed boundary at all** — the worst kind of failure for a CFD code, since the run completes and looks plausible.

Rather than teaching every downstream dispatch site about the new IDs, the follow-up collapses unified/legacy IDs to one canonical ID per dimensionality at a single point — `f_canonical_geometry`, called right after the case file is read in both `pre_process` and `simulation`. Everything downstream then works unchanged, which also let me:

- revert the `m_icpp_patches.fpp` dispatch edits (no longer needed), including the icpp 4/11 merge — those are line/plane sweeps in the `patch_icpp` namespace (not airfoils, that's `patch_ib`), they take mutually exclusive parameters, and the checkers made those branches unreachable anyway;
- restore the cylinder strictness that got lost: geometry 10 with no lengths (or a negative one) is a hard error again instead of silently becoming a sphere, keyed on the canonical ID;
- extend the unification to the airfoil (4/11) and STL model (5/12) IB pairs, completing the list in #1543;
- add two golden tests declaring a 3D IB sphere and cylinder via geometry 2 — the unified-sphere golden is bit-identical to the legacy geometry-8 case, which is a nice end-to-end proof the canonicalization works.

Net effect vs master is −41 source LOC with the docs tables updated to match. Full test suite passes locally apart from a few chemistry cases that failed for environmental reasons unrelated to this PR; CI will confirm.

sbryngelson:
@Riship749 heads up — your force-push earlier today removed the follow-up commit I had pushed onto this branch (see the comment above). I've re-applied it on top of your amended commit (now `68891386`). Please **pull before pushing again** (`git pull --rebase`) so we don't lose it a second time.

On your amendments: the tightened two-step cylinder length check is covered by the re-applied commit (equivalent pair of checks, keyed on the canonical ID so that a cylinder with *no* lengths is also an error). I deliberately did not keep the line-sweep checker relaxation (allowing `p > 0` for geometry 4): the icpp 4/11 pair are line/plane sweeps with mutually exclusive parameters (`normal(3)` forbidden vs required), so they can't be unified the way the circle/sphere/cylinder family can — the follow-up commit reverts that part and keeps icpp 4 and 11 as separate shapes. The airfoil 4/11 unification from #1543 lives in the `patch_ib` namespace and is handled by the canonicalization.

Issue #1532, opened 2026-06-03
Deprecate IB Cylinder Length Values


danieljvickers:
In legacy MFC, cylinders could only be oriented along primary axis, and the direciton was based upon which length value is default-valued. Since we now have a rotation method in place that can accomplish this same outcome, the length_y and length_z values are no longer needed.

The task is to make edits to the m_ib_patches.fpp file and refactor the cylinder code so that it only checks length_x. Appropriate changes to prohibit statements for the rest of the code should be modified that do not allow length_x to be default and require length_y and length_z to be default.

danieljvickers:
I closed this issue with #1454

PR #1549, opened 2026-06-09
1454-projection-optimization 

### LINKED ISSUE #1454: Projection Optimization

danieljvickers:
## Description

In many-particle cases, the limiting factor in the IBM compute is by far the time it takes to project the immersed boundaries onto the grid. This is fundamentally rooted in how we parallelize the work. The current code parallelizes of x, y, and z, but sequentially iterates through the IB patches. In cases where there are many IBs that are small, we are launching several (thousands) of GPU kernels each time step, but each kernel only works on hundreds of grid cells. Adding an option to parallelize over the thousands of particles should be significantly larger parallelism and thus optimize the projection.

In order to maintain the functionality of both parallelism methods, I need a separate set of geometry bounding checks. Since we perform a check in icpp patches and now 2 in IB patches, this is a significant amount of redundant code that must be maintained. To be somewhat forward-looking, I opted to merge all geometry checking into a single module that can be called from both forms of IB parallelism and the icpp pre_processing code. This should clean up the code nicely and significantly reduce code maintenance going forward. Since we can now change cylinder orientation with angles, I also deprecate the unneeded cylinder length checks.

The end result is the creation of a new module in common, the deletion of duplicate code, and a new parallelism path for IB patches when utilizing GPU compute.

Closes #1454, #1532, #1543

- New feature
- Refactor

## Testing

All tests pass on GNU compiler

## Checklist

<details>

</details>

## AI code reviews


danieljvickers:
All AI review comments are now irrelevant because of significant changes that have occurred between now and that review.

I doubt that this will pass the test suite on the first try, but on the off chance that it does we should not yet merge it. A data product of computational optimization performance should be presented before this PR is merged. Otherwise, it is unnecessary refactoring of the code.

danieljvickers:
runtime_comparison
Successful runtime comparison on Frontier. This plot shows run times for the s_apply_ib_patches subroutines.

In this test, I ran with 8k particles being projected onto the grid. Master is the current master branch (before this is merged) as a baseline of performance. The other two plots show the two types of parallelism now available from this branch.

the grid parallelism is the current Master branch functionality. This demonstrates that the performance is equivalent. The IB parallelism is the new parallelism introduced by this feature. In this particular case of 8k sphere with only about 300 grid cells per sphere, there was a 150x performance improvement over the previous parallelism. This this subroutine is called 3x per time step in the test, this is a 327 ms time reduction PER TIME STEP.

This plot should demonstrate that this feature is useful and should be merged. I will now move on to resolving any remaining issues with the test suite and hopefully getting this on master.

sbryngelson:
I've pushed a merge of `master` into this branch (ac1681a96, a true merge commit — no history rewritten) to resolve the conflicts from the recently-merged global-parameters/registry refactors (#1550–#1556). What changed and why:

**`src/post_process/m_global_parameters.fpp`** — master moved `shear_*`, `proc_coords`, `start_idx` into the new shared `m_global_parameters_common`; kept your `ib_airfoil`/`ib_airfoil_grids` stubs that the new common `m_patch_geometries` module needs to compile for post_process.

**`src/simulation/m_global_parameters.fpp`** — the `GPU_DECLARE` lines for `ib`, `num_ibs`, and `ib_coefficient_of_friction` now live in `m_global_parameters_common`, so the manual list reduces to `ib_airfoil_grids`. Your addition of `stl_models` to the GPU declare list was **relocated** to the new parameter pipeline: derived-type declarations (and their sim-side `GPU_DECLARE`) are auto-generated from the `TYPED_DECLS` table in `toolchain/mfc/params/definitions.py` — I flipped the `stl_models` gpu flag there to `True`, which emits the same declare in the generated decls.

Heads-up for future commits on this branch: Fortran parameter declarations, namelist bindings, and MPI broadcasts are now generated from `toolchain/mfc/params/definitions.py` at build time (see `docs/documentation/contributing.md`). Your `many_ib_patch_parallelism` registration came through cleanly and its broadcast is auto-generated — no action needed.

Verification on my end: `./mfc.sh format`, `./mfc.sh precheck`, the toolchain pytest suite (342 passed), and a full CPU build of all three targets all pass on the merged tree. Nothing remains on your side; CI should run on the updated head.

danieljvickers:
I am very glad you have merged this for me because I screwed up the previous one significantly, lol.

danieljvickers:
@sbryngelson I assume you are going to want to wait merging this until after we resolve the master branch AMD issue with the 2D viscous shock tube case. So I am going to just table it here. But otherwise it looks like we pass the full test suite and it is ready* to merge.

sbryngelson:
> @sbryngelson I assume you are going to want to wait merging this until after we resolve the master branch AMD issue with the 2D viscous shock tube case. So I am going to just table it here. But otherwise it looks like we pass the full test suite and it is ready* to merge.

@danieljvickers noted!

PR #1551, opened 2026-06-10
Complete named-constant adoption and parameter declaration generation

sbryngelson:
## Summary

Follow-up to #1550, completing two more pieces of the same strategy — **one source of truth in the parameter registry, generated everywhere, no behavior change**:

1. **Named-constant adoption finishes.** The 9 remaining enumerated vocabularies (`bubble_model`, `avg_state`, `wave_speeds`, `recon_type`, `muscl_order`, `muscl_lim`, `int_comp`, `format`, `precision`) are swept across all three executables — 99 decision sites move from integer literals to the generated named constants. The legacy `WENO_TYPE`/`MUSCL_TYPE` constants are retired in favor of the generated `recon_type_weno`/`recon_type_muscl` (same values, one naming scheme).
2. **The params pipeline now generates ALL namelist-bound declarations.** A `TYPED_DECLS` table in `definitions.py` drives generation of the 14 derived-type namelist declarations (`fluid_pp`, `patch_icpp`, `acoustic`, …) per target — including their `GPU_DECLARE` lines and Doxygen descriptions — and the simulation case-optimization `#:if MFC_CASE_OPTIMIZATION` block is emitted from a registry loop instead of being hand-maintained. The corresponding manual declarations in the three `m_global_parameters.fpp` are deleted (net −40 LOC of hand-synced Fortran). Adding a parameter no longer requires touching Fortran declarations at all; adding a case-optimization parameter no longer requires editing a parallel `#:else` block.

Two infrastructure hardenings that fell out of the work:

- **Generated includes regenerate at build time.** Previously they were written only at CMake configure time, so a cached build that skipped reconfigure compiled stale includes (a silent wrong-binary hazard we hit during development). A ninja-tracked `add_custom_command` now owns regeneration: `touch definitions.py && ninja` does the right thing without reconfigure.
- **Doc lint accepts registry-derived tokens.** `lint_docs.py` validates backticked tokens in `case.md` against the registry's value names and the analytic-expression intrinsics instead of a hand-grown skip list.

## Verification

- Every Fortran-touching commit was gated on the full golden suite (sharded across two parallel jobs) with **zero golden-file regenerations**, four full-suite runs in total including one against current master after #1546 landed.
- The declaration-generation steps were additionally proven by **declaration-set equivalence** on the Fypp-emitted `.f90`: every declaration and GPU-residency directive before the deletion is byte-identical (modulo relocation and comments) to the generated form after, per target and — for the case-optimization block — in both `MFC_CASE_OPTIMIZATION` configurations.
- Toolchain pytest suite extended to 323 tests; precheck green on every commit.

## What still requires manual edits (documented in `.claude/rules/common-pitfalls.md`)

Derived-type member definitions in `m_derived_types.fpp`, default-value assignments, the six case.py-computed case-opt extras, multi-variable declaration lines (`bc_x/y/z`, domains), and the MPI broadcast lists — the broadcast lists are the next generation target.


sbryngelson:
### Code review

No issues found in this PR's own commits. Checked for bugs and CLAUDE.md compliance across five independent passes (compliance, shallow bug scan, git-history conflicts with recent master merges, prior-PR feedback, and in-code guidance), with emitted-declaration equivalence verified per target. The codecov patch-coverage number reads low because the diff is dominated by generated Fortran declarations, which the golden regression suite (not unit tests) exercises; the toolchain Python in this PR is unit-covered.

PR #1552, opened 2026-06-11
Share the global-parameters core across executables; generate MPI broadcasts; build hygiene

### LINKED ISSUE #1558: post_process never broadcasts chem_wrt_Y: silent wrong species output on multi-rank chemistry post-processing
**Symptom:** multi-rank chemistry post-processing silently drops or garbles species (`Y_*`) output on ranks != 0.

**Mechanism:** `chem_wrt_Y` is namelist-read (rank 0 only) and consumed on **all** ranks by `s_save_data` (`src/post_process/m_start_up.fpp:292` on current master), but it is absent from `s_mpi_bcast_user_inputs` in `src/post_process/m_mpi_proxy.fpp` — its sibling `chem_wrt_T` is in the logical broadcast list, `chem_wrt_Y` is not. Non-root ranks always see the default.

**Introduced:** #544 (`17eb3def8`, 2024-09-16, @henryleberre) added the namelist binding and consumption but only updated the simulation proxy. #829 later fixed `chem_wrt_T` and missed `chem_wrt_Y`. Tagging @henryleberre for context as the original author.

**Fix:** #1552 replaces the hand-maintained broadcast lists with registry-generated ones (`generated_bcast.fpp`), which closes this by construction — a registered namelist parameter is broadcast or the generator is wrong.

### LINKED ISSUE #1563: patch_ib case validation accepts indices that overflow the namelist array (bound uses the wrong constant)
**Symptom:** a case setting `patch_ib(k)` for 54,000 < k <= 2,050,000 passes validation, then fails (or worse) at the Fortran namelist read — the array is dimensioned by the smaller constant.

**Mechanism:** `toolchain/mfc/params/definitions.py:34` reads `num_ib_patches_max` (= 2,050,000, `src/common/m_constants.fpp:30`) as the `patch_ib` max index, while the namelist array is dimensioned `num_ib_patches_max_namelist` (= 54,000, `m_constants.fpp:32`).

**Introduced:** #1524 (`16fed9bde`, 2026-06-03, @danieljvickers) flipped the constant name from `num_ib_patches_max_namelist` to `num_ib_patches_max`; before that (#1378 onward) the toolchain pointed at the namelist constant correctly. Tagging @danieljvickers for context.

**Fix:** #1552 repoints the bound at `num_ib_patches_max_namelist`.

### LINKED ISSUE #1565: post_process and simulation disagree on the Lagrangian void-fraction index position for bubbles_lagrange + mhd/elasticity/cont_damage
**Symptom:** post_process misreads simulation restart/output layouts for `bubbles_lagrange` combined with mhd, elasticity/stress, or cont_damage — the void-fraction field lands at a different index than post expects. No checker forbids the combinations.

**Mechanism:** on master, post folds beta into `sys_size` mid-build, BEFORE the mhd/elasticity/cont_damage blocks (`src/post_process/m_global_parameters.fpp:480-483` vs :485+), while simulation writes beta at `sys_size + 1` after the complete system (`src/common/m_mpi_common.fpp:115-126`).

**Introduced:** the original lagrange PR #749 was consistent; the divergence came with the MPI refactor #831 (`8e94d0a5f`, 2025-06-13, @wilfonba), which moved post's beta into sys_size ahead of the pre-existing feature blocks. Tagging @wilfonba for context.

**Fix:** #1552's unified equation-index builder (one `s_initialize_eqn_idx` shared by all three executables) makes the layouts agree by construction.

sbryngelson:
> **Stacked on #1551** — this branch includes those commits; the diff below them is this PR's own content (14 commits). Review after (or alongside) #1551; it will rebase cleanly once #1551 merges.

## Summary

Third installment of the parameter/declaration single-source-of-truth work (#1550, #1551). Two thrusts:

1. **Executable dedup.** The three per-executable `m_global_parameters.fpp` shared 50–90% of their content by copy-paste. A new `src/common/m_global_parameters_common.fpp` now owns the generated declaration includes, the shared state (`eqn_idx`, `sys_size`, `b_size`, `tensor_size`, shear bookkeeping), the equation-index setup (`s_initialize_eqn_idx`), parallel-IO init, the finalize core, and the common default assignments. The per-target modules keep only what genuinely differs (QBMM moment encodings, GPU update blocks, post's output state — each leave-behind verified divergent-by-need). The three `m_mpi_proxy.fpp` broadcast lists are now **generated from the parameter registry** (`generated_bcast.fpp` per target): a registered namelist parameter is broadcast by construction, with only computed variables (`m_glb`, `cfl_dt`, `bc_io`, …) remaining hand-listed. Net `src/` change: **−477 LOC**.

2. **Build hygiene.** `CMakeLists.txt` (1,135 lines) splits into a ~460-line root plus `cmake/{GPU,Fypp,ParamsCodegen,MFCTargets}.cmake` — pure motion, verified by byte-identical configure graphs. The generated includes (now 15: namelist, decls, constants, case-opt, bcast × 3 targets) regenerate via a single ninja-tracked custom command; the configure-time bootstrap and its mtime stamp are deleted (editing `params/*.py` + rebuild now always yields a correct binary, no reconfigure). Build variant directories get readable names: `build/install/gpu-acc-chem-f93aa400b9` instead of `de835ec46c` (hash inputs unchanged — variant identity preserved).

## Bug fix (deliberate behavior change #1)

`post_process` never broadcast `chem_wrt_Y` despite consuming it on all ranks in `s_save_data` — silent wrong/missing species output for multi-rank chemistry post-processing. The registry-driven generator fixes this by construction. (Broadcast tuple-set deltas vs. before, verified per target: simulation identical; pre_process adds only the dead-but-registered `n_start_old`; post_process adds only `chem_wrt_Y`.)

## Deliberate behavior change #2

`post_process`'s `beta_idx` (Lagrangian bubble void fraction) is now appended after the MHD/elasticity/surface-tension blocks, matching `simulation`'s writer layout — the previous post-only ordering misread restart data for `bubbles_lagrange` combined with those features. Covered by the full golden suite.

## Verification

- Full golden suite (sharded) green on the dedup stack and again on the final rebased tip (post-#1545), including the chemistry tests that exercise per-case rebuilds.
- Emitted-`.f90` declaration-set and GPU-directive-set equivalence per target for every hoist step; broadcast tuple-set equivalence per target; configure-graph byte-equivalence for the CMake split; cold-start + incremental-regeneration proofs for the generation rework.

## Notes for reviewers

- `src/common` continues to compile once per executable with `MFC_<TARGET>` defines — deliberate and now documented in `contributing.md` ("How the Build Fits Together"); it is what enables per-target generated includes and the shared module pattern.
- A follow-up PR (broadcast correctness) fixes three pre-existing multi-rank broadcast bugs found while auditing the generated lists; it is stacked on this branch.
- Old hash-named build directories orphan until `./mfc.sh clean`.

## Issues fixed

Fixes #1558 (chem_wrt_Y never broadcast in post_process)
Fixes #1563 (patch_ib validation bound uses the wrong constant)
Fixes #1565 (post/sim beta_idx layout disagreement for bubbles_lagrange + features)


sbryngelson:
Confirmed and fixed in e4bb56a65: the hardcoded fluid_pp emitter predated the #1545 merge and dropped the Herschel-Bulkley members (K, nn, tau0, hb_m, mu_min, mu_max, mu_bulk, non_newtonian) from the generated broadcast lists. The emitted sets are now verified member-identical to master's manual lists for all three targets, and the test suite pins the HB members so this class of escape fails loudly. The stacked follow-up PR (#1553) additionally converts this emitter to a registry walk, making new fluid_pp members broadcast by construction. Good catch.

sbryngelson:
Both review items now addressed on this branch: the fluid_pp broadcast regression (e4bb56a65, member-identical to master's manual lists, test-pinned) and the merged m_constants section heading (dac3d1ff0). Thanks for the thorough review.

sbryngelson:
### Code review

Found 1 issue (beyond the two from the earlier automated review, both already addressed):

1. `NIB` reads the wrong Fortran constant: `num_ib_patches_max` (2,050,000) instead of `num_ib_patches_max_namelist` (54,000), while the `patch_ib` namelist array is dimensioned by the latter — so case validation accepted `patch_ib` indices that overflow the Fortran array. The adjacent comment even claims it enforces the namelist limit. Fixed in f42e532a21fd376bc9a377ff8c2eafc6a08cf73b along with three doc passages made stale by the dedup.

https://github.com/MFlowCode/MFC/blob/dac3d1ff043ccdaa72629d7de1dda6f059e126ec/toolchain/mfc/params/definitions.py#L33-L35

PR #1553, opened 2026-06-11
Fix three pre-existing multi-rank broadcast defects; deregister dead parameters

sbryngelson:
> **Stacked on #1551 and the executable-dedup PR** — this branch includes both; its own content is the top 3 commits. It exists separately because, unlike its parents, it intentionally changes multi-rank behavior (it fixes bugs).

## Summary

While converting the MPI broadcast lists to registry-driven generation (parent PR), a member-by-member audit of every namelist parameter's broadcast/consumption pattern turned up three pre-existing defects on `master`. All three predate this work.

1. **`bc_x/y/z%{vb1..3, ve1..3}` were never broadcast in `pre_process`/`post_process`** but are consumed on all ranks by `s_slip_wall`/`s_no_slip_wall` in `m_boundary_common` — rank-dependent ghost cells for any multi-rank wall-velocity BC run in pre/post. Fixed by adding the member broadcasts (simulation already had them).

2. **`muscl_eps` was never broadcast**, and its post-read derivation runs only under `f_is_default(muscl_eps)` with defaults assigned rank-0-only — so non-root ranks carried divergent values on every multi-rank MUSCL run. Fixed by broadcasting it like every other registered REAL.

3. **14 dead registrations crash the namelist read**: `lag_params%{T0, Thost, c0, rho0, x0}` and `fluid_pp%{mul0, ss, pv, gamma_v, M_v, mu_v, k_v, cp_v, D_v}` were removed from the derived types (#1085/#1093) but never deregistered from the toolchain — setting any of them in a case file aborts with a misleading "datatype mismatch". Deregistered; the case validator now rejects them by name instead.

With the dead members gone, the `fluid_pp`/`lag_params` broadcast emitters convert from hardcoded member lists to registry walks — a member added to those types is now broadcast by construction, closing the gap that allowed bugs 1–2 to exist.

## Verification

- Full golden suite (sharded, multi-rank) green on the final tip.
- Broadcast tuple-set comparison per target: deltas are exactly the intended additions (vb/ve members in pre/post, `muscl_eps` everywhere it is registered) and the dead-member removals; everything else identical.
- Toolchain pytest green, including new tests pinning the registry-walk emission and the deregistrations.

## Issues fixed

Fixes #1559 (muscl_eps never broadcast)
Fixes #1560 (wall-velocity vb/ve members never broadcast in pre/post)
Fixes #1561 (pre_process broadcasts integer BC codes as mpi_p)
Fixes #1562 (14 phantom parameter registrations crash the namelist read)

PR #1556, opened 2026-06-11
Phase-2 completion: post-process context types, m_riemann_solvers split, cleanup

### LINKED ISSUE #1564: post_process: Q-criterion-only cases access unallocated finite-difference coefficient arrays
**Symptom:** a post_process case enabling only `qm_wrt` (no vorticity/schlieren/liutex) reads unallocated `fd_coeff_*` arrays — crash or garbage depending on compiler.

**Mechanism:** the allocation guards in `s_initialize_derived_variables_module` (`src/post_process/m_derived_variables.fpp:44/48/52` on master) test `omega_wrt`/`schlieren_wrt`/`liutex_wrt` but not `qm_wrt`, while the call guards in `m_start_up.fpp:216/220/224` (and the Q-criterion computation at :631) include `qm_wrt`.

**Introduced:** with the feature in #79 (`fd07dfe65`, 2023-01-30, @hyeoksu-lee) — the call guards gained `qm_wrt`, the allocation guards never did (liutex was later added to both; qm never was). Tagging @hyeoksu-lee for context.

**Fix:** #1556 makes the allocation guards mirror the call sites exactly.

sbryngelson:
> **Stacked on #1551 → #1552 → #1553 → #1555** — its own content is the top 10 commits. Completes the refactoring roadmap's Phase 2 using the patterns the pilots (#1555) validated.

## Summary

1. **Context types (post_process).** The remaining module-level state bundles follow the pilot pattern: `fd_context` (finite-difference workspace: `gm_rho_sf`, `fd_coeff_x/y/z`) and `output_context` (the 18-member output workspace: `q_sf` family, Silo extents/offsets, paths and handles). Strictly NFC, verified by emitted-Fortran equivalence after stripping the qualifier.

2. **`m_riemann_solvers.fpp` split (4,706 lines → 6 modules).** The four solvers extract to `m_riemann_solver_{hll,hllc,hlld,lf}.fpp`; the shared device state (14 GPU-declared arrays/bounds) and the per-sweep lifecycle + viscous helpers the solvers call land in `m_riemann_state.fpp` (the lowest layer — keeping them in the dispatching core would create a module cycle); the original module shrinks to a 116-line dispatcher that re-exports the public surface unchanged (zero external callers modified). Pure motion: statement-multiset equivalence under CPU/OpenACC/OpenMP emission; GPU-directive census identical (747 directives); every `declare target` verified to live in its declaring module (the Cray ftn-1448 rule).

3. **Two latent bug fixes** (found by review during the above):
   - post_process's finite-difference coefficient arrays were never allocated when only `qm_wrt` is set, while the coefficient computation runs for it — unallocated access for Q-criterion-only cases. The allocation guards now mirror the call sites.
   - (In the cleanup batch) the case validator accepted `patch_ib` counts above the namelist array bound — carried in #1552's follow-ups; noted here because the registry-driven message change rides this stack.

4. **Cleanup batch:** the `recon_type` validator message now derives from the parameter registry (the last hardcoded vocabulary site); fypp `#:for` set-literals become list-literals (emission order was Python-hash-dependent across processes — builds are now order-reproducible); one stale import removed; the slimmed Riemann core drops two unused imports.

## Verification

- Full golden suite (4-way sharded) green twice: once on the context/cleanup layer, once on the complete tip with the Riemann split.
- Per-commit batteries on the split: statement-multiset union equivalence (seed-pinned, three emission configs), GPU-directive multiset equality, declare-target scoping checks with a negative control, identifier sweeps, toolchain pytest.
- Independent adversarial review of every task; the split's review reproduced the call-graph that forced the state-module layout and the full equivalence evidence.

## Notes for reviewers

- This is MFC's largest GPU surface; the Frontier/NVHPC matrix on this PR is the decisive device-path check (all local proofs are emission-level by design — the working model runs no local GPU builds).
- Pre-existing, deliberately untouched (noted for future cleanup): `Gs_rs`/`Res_gs` allocate without a finalize-side deallocate (base-identical), and the single-precision `q_sf_s` triple is never deallocated.
- `m_riemann_state` holds the per-sweep helpers as well as state — the call graph forces it; the module doc enumerates the contents.

## Issues fixed

Fixes #1564 (Q-criterion-only post cases access unallocated FD coefficient arrays)

Tracked, not fixed here: #1568 and #1569 (deallocate gaps documented in the review notes above).


sbryngelson:
Both doc comments addressed — the passages predated the build-time generation command (added in #1552's build-hygiene work) and still described the configure-time era. Corrected text: a plain rebuild regenerates the includes automatically (the custom command is ninja-tracked against everything under `toolchain/mfc/params/`, generators included, via GLOB_RECURSE — so editing an existing generator file does retrigger); the one true reconfigure case is *adding a new file* there, since the dependency list is globbed at configure time. That caveat now appears where the misleading 'force regen by reconfiguring' advice was.

Issue #1571, opened 2026-06-11
HLL hyperelastic wave-speed branch reads shear moduli that only the hypoelastic path initializes


sbryngelson:
**Symptom:** an HLL + `hyperelasticity` run computes its direct elastic wave-speed estimates from uninitialized `G_L/G_R` — hyperelasticity is not restricted to any solver, so this is a live, reachable path.

**Mechanism:** HLL's `else if (hyperelasticity)` wave-speed branch (master `m_riemann_solvers.fpp`, post-split `m_riemann_solver_hll.fpp:364-367`) uses `G_L/G_R`, which HLL assigns only inside its `if (hypoelasticity)` energy block. HLLC, by contrast, loads G under the merged `elasticity` gate before using it.

**Introduced:** the hyperelastic wave-speed branch arrived with the elasticity series around #727 (@mrodrig6); the G-loading asymmetry between HLL and HLLC has existed since. Tagging @mrodrig6 for context. Related: #1557 (the hyper tau_e formulation divergence between HLL and HLLC — these two would naturally be ruled on together).

**Status:** untouched by the refactoring series (correctly — physics). Needs an elasticity-author ruling on what HLL's hyperelastic branch should read.

Issue #1570, opened 2026-06-11
HLL hypoelastic damage load reads the LEFT cell's damage for the RIGHT state (missing +1 offset)


sbryngelson:
**Symptom:** with `hypoelasticity` + `cont_damage` (HLL — the only legal hypoelastic solver), the right interface state's shear modulus is scaled by the LEFT cell's damage value.

**Mechanism:** every right-state load in the Riemann solvers uses the `+1`-offset cell — except this one: the damage scaling reads `qR_prim_rsx_vf(SF(''), eqn_idx%damage)` with no offset (post-split: `src/simulation/m_riemann_solver_hll.fpp` damage load; on master inside `s_hll_riemann_solver`'s elastic-energy block). `damage_L` and `damage_R` therefore read the same cell.

**Introduced:** with the continuum damage model, #816 (`068da2c05`, 2025-04-14, @ChrisZYJ). Tagging @ChrisZYJ for context — possibly intentional (damage as cell-centered non-Riemann quantity?), but it contradicts every neighboring load.

**Status:** deliberately preserved bit-for-bit by the Phase-3 refactoring (behavior changes need their own ruling). Needs an author/maintainer decision; the fix is a one-character offset if unintentional.

Issue #1557, opened 2026-06-11
Hyperelastic wave-speed estimates: HLLC includes the tau_e correction, HLL deliberately omits it


sbryngelson:
Found during the Phase-3 Riemann refactoring review (stacked PRs #1551-#1556 series).

**Divergence (live behavior):** HLLC applies the Rodriguez et al. (2019) elastic wave-speed correction (the `(4G/3 + tau_e(dir_idx_tau(1)))/rho` term in s_L/s_R, plus the stress-corrected s_S) under `if (elasticity)` — which covers **hyperelastic** runs, legal with HLLC. HLL has separate branches: its `hypoelasticity` branch uses the same tau_e algebra, but its `hyperelasticity` branch **deliberately omits the tau_e term**.

So a hyperelastic case run with HLLC vs HLL uses different wave-speed estimates as a matter of formulation, not scheme. One of these is presumably the intended physics for hyperelasticity.

Locations (post-#1556 module layout): `src/simulation/m_riemann_solver_hllc.fpp` elasticity blocks (~319-332, ~1354-1367); `src/simulation/m_riemann_solver_hll.fpp` hypo (~366-375) vs hyper (~376-380) branches.

Per maintainer decision (2026-06-11) both are left as-is pending a ruling from the elasticity model authors. Filing so it is tracked rather than folklore.

PR #1572, opened 2026-06-11
Phase-3: Riemann hot-path decomposition into shared GPU device helpers

### LINKED ISSUE #1566: HLLC hypoelastic energy block mis-doubles the theta-theta normal stress in 2D-axisymmetric (currently dead code)
**Symptom:** none today — the case validator restricts hypoelasticity to HLL, so HLLC's hypoelastic blocks are unreachable. If that restriction were lifted, 2D-axisymmetric hypoelastic HLLC runs would double a NORMAL stress as if it were shear.

**Mechanism:** HLLC's elastic-energy block uses the hard-coded shear test `(i == 2) .or. (i == 4) .or. (i == 5)` (`src/simulation/m_riemann_solvers.fpp:1914, :2990` on master), but in 2D-axisymmetric the 4th stress component is theta-theta (normal). HLL's `shear_indices`-based test is dimension-correct.

**Introduced:** #727 (`10b2209e7`, 2024-12-20, @mrodrig6); it became physically meaningful when #815 added axisymmetric hypoelasticity and fixed only HLL. Tagging @mrodrig6 for context.

**Fix:** pending in the Phase-3 refactoring series (stacked on #1556): the dead HLLC blocks are deleted outright and the surviving HLL logic is unified into one helper using the `shear_indices` form.

### LINKED ISSUE #1567: Uninitialized read into Re_avg on dead viscous lanes (UB; FP-trap hazard in debug builds)
**Symptom:** results-inert today (every consumer is gated on the matching `Re_size` lane), but the code computes `2/(1/uninitialized)` — undefined behavior that can trap in debug/FPE builds.

**Mechanism:** the unguarded `do i = 1, 2` Reynolds-averaging loops (`src/simulation/m_riemann_solvers.fpp:480-484`, `:1965-1968`, `:2627-2630`, `:3045-3048` on master) write both lanes of `Re_avg_rs*_vf` even when `Re_size(2)==0`, in which case the HLLC fill loop never initializes `Re_L(2)/Re_R(2)`.

**Introduced:** the unguarded form dates to the 2022 GPU development merge (`6def75421`, @anandrdbz, pre-PR era), when the lane was garbage-but-defined via `dflt_real`; it became genuinely uninitialized when #1035 dropped that initialization (2025-11-13). Tagging @anandrdbz for context.

**Fix:** pending in the Phase-3 refactoring series (stacked on #1556): the extracted Reynolds helper writes a defined `1/sgm_eps` into dead lanes.

sbryngelson:
> **Stacked on #1551 → #1552 → #1553 → #1555 → #1556** — its own content is the top 4 commits. Completes the refactoring roadmap's Phase 3 (hot-path decomposition), the final phase of the series.

## Summary

1. **Per-solver benchmark coverage.** `5eq_rk3_weno3_hll` and `5eq_rk3_weno3_lf` join the benchmark suite (previously only HLLC and hypo-HLL had solver-specific coverage), so the performance gates below actually measure every solver the changes touch.

2. **Tier-1 shared device helpers** (`m_riemann_state.fpp`, all `GPU_ROUTINE(seq, cray_inline)`): mixture-properties aggregation (14 call sites across HLL/HLLC/LF), interface-Reynolds averaging (10 sites), and the HLLC star-momentum flux expression (6 sites). Net −50 LOC from the hot loops. Each conversion verified bitwise-identical on live lanes by independent review; blocks whose arithmetic genuinely differs between solvers were deliberately left inline rather than homogenized. Incidental fix: dead viscous lanes previously stored `2/(1/uninitialized)` into `Re_avg` — now a defined value (fixes the UB/FP-trap hazard).

3. **Hypoelastic interface energy unified** under maintainer rulings (recorded in-repo): one shared helper with the `verysmall` gate (HLL's `G > 1000` stability floor and its `TODO take out` comment retired), unconditional damage scaling, and the dimension-correct `shear_indices` doubling test; the dead hypoelastic energy blocks in HLLC (×2) and LF deleted outright (the case validator restricts hypoelasticity to HLL). Net −56 LOC. Wave-speed estimates untouched everywhere.

## Performance gate (<5% ceiling, maintainer-set)

Simulation grind-time ratios vs. the pre-series baseline, all benchmark cases:

| stage | range | watched case |
|---|---|---|
| Tier-1 helpers | 0.99–1.04 | `5eq_rk3_weno3_hllc` **1.02** (faster) |
| Hypoelastic unification | 0.98–1.03 | `hypo_hll` **1.03** (faster) |

(Ratios are speedups: >1 is faster. Pre/post exec scatter is sub-second I/O noise, documented in the working notes.)

## Behavior notes

- Golden suite green (4-way sharded) after each stage — the changes are golden-test-silent, but two are real for untested regimes and deliberate: soft-G hypoelastic runs (`verysmall < G ≤ 1000`) now receive elastic energy the stability floor previously suppressed (maintainer ruling), and dead-lane `Re_avg` values changed from garbage to defined (consumers gated; results-inert).
- Fixes #1566 (the deleted HLLC blocks contained the hard-coded axisym shear-doubling bug).
- Fixes #1567 (the uninitialized `Re_avg` lanes).
- Deliberately NOT touched, tracked for physics rulings: #1557, #1570, #1571 (elastic wave-speed formulation divergence, the damage-R load offset, HLL's hyperelastic G initialization).

## Docs

`gpuParallelization.md` gains a "Writing GPU_ROUTINE Device Helpers" section (the macro was previously undocumented): the FLOP threshold, the `seq`+`cray_inline` idiom and why Cray requires it, the caller-loads/helper-computes pattern, AMD case-optimization dimensioning, and declare-scoping — with a real helper from this PR as the worked example.


sbryngelson:
Thanks Copilot — all five findings verified and addressed in b43e44589. Context: cmake/MFCTargets.cmake was created by pure motion from the monolithic CMakeLists.txt (the split was verified by byte-identical configure graphs), so all five lines pre-exist on master — but four were worth fixing on substance: the unified-memory flags (the NVHPC managedalloc compile/link pair and both FRONTIER_UNIFIED definitions) now apply to `${a_target}` so the NVHPC two-pass IPO lib target receives them consistently, and the leftover debug `message(STATUS)` is gone. The `set(ENV{OMP_TARGET_OFFLOAD} [MANDATORY])` line was deleted outright rather than de-bracketed: `set(ENV{...})` only affects the configure-time CMake process environment, so it never influenced the built binaries regardless of the bracket bug.

sbryngelson:
### Code review

Found 2 issues, both fixed in fc7f770c719f6668b09b97a6e6a8bfcaf47cc28b:

1. The hypoelastic energy-block deletions left LF and HLLC with hypoelasticity-gated flux/source blocks consuming `tau_e_L/R` that nothing loads (flagged independently by three review passes). Unreachable through the toolchain (`case_validator.py:577` restricts hypoelasticity to HLL), but a latent-UB trap if that restriction is ever lifted. Resolved by completing the dead-copy deletion under the same maintainer ruling: every hypoelasticity-gated block in LF/HLLC removed after a guard-by-guard inventory; all `elasticity`/`hyperelasticity`-gated code untouched (with HLLC, `elasticity` implies `hyperelasticity`, so the kept consumers are always fed by the kept loads). −60 LOC of dead code, plus orphaned locals.

https://github.com/MFlowCode/MFC/blob/a577f15e09134f4f12bd49646bee9ed8fcab84a0/src/simulation/m_riemann_solver_lf.fpp#L443-L450

2. The new GPU_ROUTINE documentation overstated `cray_inline=True` as required "always" while two same-module helpers in the same file correctly omit it, and misplaced the AMD dimensioning guard as "caller-side". Both passages corrected (cray_inline scoped to cross-module helpers; the guard sits on the helper's dummy declaration), and the hypoelastic helper's doc comment now records the deliberate verysmall-gate ruling.

https://github.com/MFlowCode/MFC/blob/a577f15e09134f4f12bd49646bee9ed8fcab84a0/docs/documentation/gpuParallelization.md#L636-L640

Verified sound across five independent review passes: helper algebra bitwise-faithful at all 30 call sites (sampled), no upstream hotfixes lost in any replaced block (history audit vs #1545/#1556 merges), Herschel-Bulkley viscous arguments preserved, cmake `a_target` fixes correct, all five prior Copilot points confirmed addressed, merge-base current. Known and deliberately preserved pending physics rulings: #1570 (damage-R offset), #1557/#1571 — declared in the PR description.

sbryngelson:
Device validation complete on the current head's content (860ecb3b1): full 594-test suite run on GPU nodes under both backends — OpenACC and OpenMP offload — ALL TESTS PASSED on each, alongside the 4-way CPU golden gate. Every change in this PR now has empirical GPU evidence in addition to the emission-level proofs.

sbryngelson:
will merge on ci completion

PR #1581, opened 2026-06-12
Fix three broadcast defects found by post-merge audit; harden the generator

### LINKED ISSUE #1579: pre_process broadcasts the integer hcid patch member with the real datatype (mpi_p)
**Found by the post-merge broadcast audit of the #1550-#1556 series.** `patch_icpp(i)%hcid` is `integer` (`m_derived_types.fpp:307`) but sits in pre_process's `mpi_p` patch-member broadcast loop (`src/pre_process/m_mpi_proxy.fpp:94` on master) — an 8-byte real transfer over a 4-byte integer, the same UB class as the `bc%beg/end` fix in #1553. Consumed on all ranks (`m_icpp_patches.fpp:210`+). Practical severity low (works by adjacency; size-exact under --single), but it is a type-mismatched MPI operation. Introduced 2023-07-31 (`3fdbddf2a`, @wilfonba). Fix: move `hcid` to the `MPI_INTEGER` group.

### LINKED ISSUE #1580: post_process never broadcasts recon_type and igr_order, which the all-ranks domain decomposition consumes
**Found by the post-merge broadcast audit (two independent passes converged).** `s_mpi_decompose_computational_domain` (`src/common/m_mpi_common.fpp:1035-1049`) reads `recon_type` and (under `igr`) `igr_order` on ALL ranks to shape the processor-topology factorization constraints — but post_process broadcasts neither, and assigns defaults rank-0-only. Worse, post never default-initializes `igr_order` at all, so non-root ranks read an uninitialized integer for IGR post runs; for MUSCL cases the rank-divergent `recon_order` can feed inconsistent dims to `MPI_CART_CREATE`. Pre-existing: the old hand-written post broadcast list never had them; the registry-generation work (#1552) preserved the call set faithfully but its `_POST_BCAST_EXCLUDE` comment wrongly claimed all six exclusions are rank-0-only — that classification error is ours (@sbryngelson). Fix: broadcast both in post + give `igr_order` its missing default.

sbryngelson:
Post-merge audit of the #1550-#1572 refactoring series: four independent verification passes over merged master (squash-tree integrity, end-to-end NFC equivalence, first-principles broadcast re-derivation, hostile fresh-eyes review of the riskiest artifacts). **Zero regressions from the series itself**; the audit instead surfaced three pre-existing defects and two hardening gaps, all fixed here.

Fixes #1579 (hcid integer broadcast as mpi_p in pre_process — moved to the MPI_INTEGER group).
Fixes #1580 (post_process never broadcasts recon_type/igr_order consumed by the all-ranks domain decomposition; igr_order additionally gains its missing default — found independently by two audit passes).

Hardening (second commit): `_mpi_type_for` now consults `storage_precision` (a future `real(stp)` namelist parameter would have broadcast with `mpi_p` — silently wrong under `--mixed`), and an unregistered namelist entry now raises instead of silently skipping (the missing-broadcast class the generator exists to prevent), with the one verified-dead exception (post's `G`) allowlisted explicitly. The four surviving `_POST_BCAST_EXCLUDE` entries were each verified benign with consumer-level evidence during the audit.

Full golden suite (4-way sharded) running on the fix commit; toolchain pytest green; precheck green.


sbryngelson:
Addressed - the message is now one explicit multi-line f-string (the implicit concatenation was a formatter artifact). Thanks.

sbryngelson:
Follow-up hardening on the generator/manual-residue boundary, from the other side: a new precheck/CI source-lint rule (`check_manual_registry_bcasts` in `toolchain/mfc/lint_source.py`, commit f2e661ecc on `fix-broadcast-audit`) now forbids hand-written `MPI_BCAST`s of registry-bound namelist scalars in the three `m_mpi_proxy.fpp` files. It extracts every broadcast's first-argument root — direct calls and Fypp `#:for` loop lists, including `&`-continuations and nested loops — and checks it against the per-target auto-broadcast set derived from the generator's own `_classify_scalar_vars`, so the lint and `generated_bcast.fpp` can never disagree. Struct members and computed variables (`m_glb`, `cfl_dt`, `bc_io`, ...) remain legitimate manual residue and are skipped. Combined with the generator's existing unregistered-namelist error, the boundary is now policed from both directions: the current tree scans clean, and any future duplicate broadcast or generator bypass fails precheck.

PR #1582, opened 2026-06-12
CI: shard the AMD case-optimization pre-build

sbryngelson:
## Problem

The "Case Opt | Frontier (AMD) (gpu-omp)" job times out. Evidence: job run 80982103009 ran 2h05m and died in the **Pre-Build (SLURM)** step — `prebuild-case-optimization.sh` builds every case-optimized benchmark variant serially in a single SLURM job under AMD flang (the slowest compiler of the set), and the serial build alone now exceeds the job's walltime.

## Fix

Shard the AMD pre-build across **two concurrent SLURM jobs**:

- `prebuild-case-optimization.sh` now honors an optional shard spec (`$job_shard`, format `i/N`, already plumbed through `submit-slurm-job.sh`'s 5th argument): shard `i` builds every Nth case of the sorted benchmark list (index mod N == i-1). The two shards are deterministic, disjoint, and together cover all cases.
- The frontier_amd Pre-Build step in `test.yml` submits shards `1/2` and `2/2` as background processes and waits on both, failing if either fails. Output/job-id files get unique `-1-of-2`/`-2-of-2` suffixes (existing `submit-slurm-job.sh` behavior); the Print Logs and Archive Logs steps pick them up.

Two concurrency details, since both shards share one workspace:

- The in-job staging clean is skipped when sharded (it would wipe the sibling shard's in-progress build); the workflow runs the clean once on the login node before submitting.
- The case-optimized **simulation** builds land in per-case hash-named staging dirs (no collision), but `syscheck`/`pre_process`/`post_process` hash identically across these benchmarks. Shard 1 builds those shared targets first and drops a marker file; other shards wait for it, after which their builds no-op in the shared dirs. This avoids two cmake/ninja invocations ever running concurrently in the same staging directory.

## Unchanged

- **No shard set = exactly the old behavior** (all cases, in-job clean), so the Phoenix pre-build path and any direct invocation are unaffected.
- The Frontier (non-AMD) path doesn't use the pre-build mechanism at all — it builds inside the GPU run job via `run_case_optimization.sh` — so it is left alone.

## Validation

- `bash -n` + shellcheck clean on both scripts; `yaml.safe_load` and `./mfc.sh precheck` pass.
- Shard partition logic reproduced in a standalone shell test against the real benchmark list: `1/2` -> {5eq_rk3_weno3_hllc, ibm, viscous_weno5_sgb_acoustic}, `2/2` -> {hypo_hll, igr}; union covers all 5 cases with zero overlap; malformed shard specs (`0/2`, `3/2`, `a/b`, `12`) are rejected.


sbryngelson:
Added: the Frontier dependency-build step timeout is doubled to 120 minutes — four non-AMD Frontier jobs (including cpu) died uniformly at ~51-55 minutes in run 27371326575, consistent with login-node contention rather than code. With the prebuild sharding, this makes #1582 the complete Frontier-timeout fix.

sbryngelson:
Third fix added: the Frontier AMD **cpu** test job ran the full suite in a single 1:59-walltime SLURM job (run 27398258302 died at the limit in Test) while the AMD gpu jobs were already sharded — it now gets the same 2-way sharding. Note the check name changes to include [1/2]/[2/2]; adjust branch protection if it pinned the old name.

sbryngelson:
All three review findings addressed: shard validation now checks each part independently and enforces the full i/N shape (`1/`, `/2`, and slashless forms rejected); shard 1 clears both coordination markers at its own start, so stale state from reruns or manual invocations cannot skip the wait; and a failure marker written via ERR trap makes waiting shards fail fast with a pointer to shard 1's log instead of waiting out the 90-minute timeout. 22 standalone test cases over the validation and marker protocol, bash -n and shellcheck clean.

sbryngelson:
Added a fourth hardening fix (cf065ceab): the benchmark job now fails when the PR benchmark run genuinely fails, not only when its output YAML is missing. `run_parallel_benchmarks.sh` previously downgraded a non-zero PR SLURM exit to a warning and only hard-failed on an absent YAML — so a PR whose cases all SIGTERM'd/crashed/hung still passed green as long as a partial YAML was written (seen on #1588, where phoenix-bench node contention killed every PR case yet the check stayed green). Now a real PR-job failure exits 1. Scoped to PR only (a master/baseline infra flake stays a warning); `pr_exit` is reliable because run_monitored_slurm_job.sh already re-checks sacct and returns 0 when the job actually completed, so this won't red-cross on monitor false-positives. Trade-off: genuine phoenix-bench infra flakes on the PR side will now require a re-run rather than passing silently — correct, since a real hang must fail.

PR #1591, opened 2026-06-13
Fix .silol.series generation

wilfonba:
This PR moves the generation of the .silo.series file before the failure message in MFC.sh. If post process failed, then the silo series file wouldn't be generated. I'm not sure how this got reordered, because I thought I'd done the correct ordering when I first merged this. Git blame shows that I just did it wrong though.


sbryngelson:
@wilfonba, what do you think of the Claude review above?

sbryngelson (commented):
Reorder looks right — moving the `.silo.series` generation ahead of the `code -eq 22` / failure-`exit` blocks ensures it actually runs. One thing I checked that's reassuring: `code=$?` is captured at line 86, before this python call, so even if `generate_silo_series.py` errors it won't clobber the real post_process exit status used by the checks below. Minor inline note.

toolchain/templates/include/helpers.mako
```diff
@@ -87,6 +87,10 @@ END
 
     t_${target.name}_stop=$(python3 -c 'import time; print(time.time())')
 
+% if target.name == 'post_process':
+    python3 "${MFC_ROOT_DIR}/toolchain/templates/include/generate_silo_series.py" '${os.path.dirname(input)}'
```
sbryngelson, line 91: Non-blocking: this now runs unconditionally (for `post_process`) even when post_process actually failed — i.e. it may execute `generate_silo_series.py` over missing or partial silo output. Since `code` is already captured above, it won't mask the real error, but worth confirming `generate_silo_series.py` degrades gracefully (no silo files / incomplete set) rather than dumping a traceback. If you'd rather only generate on success, gating this on `[ $code -eq 0 ]` would do it — though I suspect you want it to run on the spurious-failure case, which is the whole point of the PR.

wilfonba: `generate_silo_series.py` reads the list of files present in the root directory when it is called, so it should never fail if files exist in the root directory. I don't know what it would do if there were no files in the root directory. I'll have to check the actual script when I'm on the right computer.

wilfonba: @sbryngelson I tried stress testing this by adding a `rm -rf silo_hdf5` in the `MFC.sh` file before the call to `generate_silo_series.py` and it fails gracefully. There's an 
```
if not collection_files:
	return
```
that does an early return if no files are present. This should be fine to merge.

Issue #1523, opened 2026-06-01
post_process: extract s_write_field helper to collapse the 38 repeated fill->name->write->clear stanzas in s_save_data


sbryngelson:
`s_save_data` in `src/post_process/m_start_up.fpp` (lines 169–747, ~580 lines) writes every output variable with the same repeated scaffolding. The call `s_write_variable_to_formatted_database_file(varname, t_step)` appears **38 times**, and nearly every one is wrapped in the identical "fill → name → write → clear" stanza:

```fortran
q_sf(:,:,:) = q_cons_vf(i)%sf(x_beg:x_end, y_beg:y_end, z_beg:z_end)   ! fill the module scratch field
write (varname, '(A,I0)') 'alpha_rho', i                              ! build the name
call s_write_variable_to_formatted_database_file(varname, t_step)     ! write it
varname(:) = ' '                                                      ! clear the name buffer
```

That 4-line pattern repeats for densities, momenta, velocities, energies, species, stresses, B-fields, vorticity, etc. Two things make it worth tidying:
- the `q_sf(:,:,:) = …(x_beg:x_end, y_beg:y_end, z_beg:z_end)` slice and the `varname(:) = ' '` clear are **identical every time** and pure boilerplate;
- the `varname(:) = ' '` clear is easy to forget when adding a new output variable (a missed clear leaves stale characters in the name buffer).

## Proposed helper

`q_sf` is a module-global scratch field (it's how `s_write_variable_to_formatted_database_file` receives the data — it reads `q_sf` directly, not via an argument). So a small private helper in `m_start_up` can wrap the common tail, with an **optional** source that handles the dominant "slice a `scalar_field`" case:

```fortran
!> Fill q_sf from `src` (if given) over the interior, write it as `varname`, and clear varname.
!! @param varname  field name (set by caller); blanked on return
!! @param t_step   current time step
!! @param src      optional scalar_field to slice into q_sf; omit if q_sf is already filled
subroutine s_write_field(varname, t_step, src)
    character(LEN=name_len), intent(inout)       :: varname
    integer,                 intent(in)          :: t_step
    type(scalar_field),      intent(in), optional :: src

    if (present(src)) then
        q_sf(:,:,:) = src%sf(x_beg:x_end, y_beg:y_end, z_beg:z_end)
    end if
    call s_write_variable_to_formatted_database_file(varname, t_step)
    varname(:) = ' '
end subroutine s_write_field
```

## Before → after

[...]

PR #1600, opened 2026-06-15
refactor: extract s_write_field helper in post_process s_save_data

SVS87:
## Description

Collapse the 38 repeated fill->name->write->clear stanzas in s_save_data into calls to a new private helper s_write_field. The helper optionally slices a scalar_field into out%q_sf, writes varname to the database, and clears varname on return.

Also fixes the ib_markers block which was missing the varname clear, and moves the elasticity/hyperelasticity varname clear inside the prim_vars_wrt guard where it belongs.

Part of #1523

- Bug fix
- Refactor

## Testing

Ran ./mfc.sh test -j 4 locally on Ubuntu/WSL with GNU compiler and MPI. 578 tests passed, 0 failed. Output-neutral by construction — the helper performs the exact same slice, write, and clear in the same order, so database files are bit-identical.

## Checklist

## AI code reviews


src/post_process/m_start_up.fpp
```diff
@@ -748,6 +656,27 @@ contains
...
+    !! @param x_beg, x_end, y_beg, y_end, z_beg, z_end  output region bounds (required if src present)
+
+    impure subroutine s_write_field(varname, t_step, src, x_beg, x_end, y_beg, y_end, z_beg, z_end)
+
+        character(LEN=name_len), intent(inout)   :: varname
+        integer, intent(in)                      :: t_step
+        type(scalar_field), intent(in), optional :: src
+        integer, intent(in), optional            :: x_beg, x_end, y_beg, y_end, z_beg, z_end
+
+        if (present(src)) then
```
sbryngelson, line 671: Latent interface weakness (non-blocking): the slice is guarded on `present(src)` only, not on the six bound args. The interface therefore permits a `src`-without-bounds call — e.g. `call s_write_field(varname, t_step, q_cons_vf(k))` — which would dereference non-present optional integers `x_beg..z_end` (undefined behavior), with no compiler diagnostic. None of the current call sites do this (all pass `src` and the six bounds together, or omit both), so there's no live bug. Worth either tightening the guard (also assert `present(x_beg)`) or documenting that `src` and the bounds are all-or-nothing, to protect future callers.

sbryngelson (commented):
Reviewed as a behavior-equivalence audit of the refactor. Traced all 38 converted stanzas against their originals:

- Every `src`-passing call uses the same field, index expression, and `x_beg:x_end,y_beg:y_end,z_beg:z_end` slice as the original inline fill (checked the index-bearing ones: `mom`, `vel`, `Y`, `alpha_rho_e`, `tau`, `xi`, `alpha`, `color_function`, all qbmm `rs/vs/ps/ms(i)`).
- Every no-`src` call still has `out%q_sf` populated beforehand by a retained inline slice, an `s_derive_*` call, or a manual loop — none orphaned.
- `s_write_field` is a sibling module procedure; `out` resolves via use-association from `m_data_output` (public module var), and `name_len`/`scalar_field`/`wp` from their respective modules. ✓
- The `varname`-clear relocation is inert: every DB write is preceded by a full internal `write(varname,...)`, which blank-pads the whole fixed-length scalar, so the clears were always redundant.

Output-neutral, as claimed. One latent (non-blocking) note inline.

src/post_process/m_start_up.fpp
```diff
@@ -748,6 +656,27 @@ contains
 
     end subroutine s_save_data
 
+    !> Fill out%q_sf from src (if given), write varname to the database, and clear varname.
+    !! @param varname  field name (set by caller); blanked on return
+    !! @param t_step   current time step
+    !! @param src      optional scalar_field to slice into out%q_sf
+    !! @param x_beg, x_end, y_beg, y_end, z_beg, z_end  output region bounds (required if src present)
+
```
sbryngelson: remove empty line

PR #1610, opened 2026-06-18
fix-force-integration-race-condition

danieljvickers:
## Description

Mark found a race condition in the way that we compute IB forces

- Bug fix


danieljvickers:
@sbryngelson This branch, and two others that I have opened, are ready for review to finish the test suite.

PR #1614, opened 2026-06-22
1104 mibm higher order central differencing

### LINKED ISSUE #1104: MIBM Higher Order Central Differencing
### Problem it solves
Much of the moving immersed boundary code uses second-order central differences, which is obviously sub-ideal.

### Proposed solution
Replace the hard-coded central differences with a method that uses the fd_coefficients in the code

danieljvickers:
## Description

For a while I have been putting off adding higher-order central difference mostly because I did not see an immediate need. However, it appears that the current viscous force calculation in MFC has some significant errors for low Reynolds cases. Increasing the finite-difference order for the derivatives to 4th order appears to significantly reduce this error back within acceptable tolerances. This branch integrates the IB code with the existing finite-difference coefficients in MFC to be utilized in the force calculation loop. This has added the side-benefit of somewhat reducing the amount of code in the `s_compute_ib_forces` subroutine

Closes #1104 

- Bug fix
- New feature
- Refactor

## Testing

I did a significant amount of post-processing analysis of the data to determine the best integration method for low-Reynolds cases, which wound up being 

## Checklist

<details>


danieljvickers:
Fixed the AI review comment

danieljvickers:
I reran a convergence test and compared the force integration to my independent python implementation of all of the drag coefficient for on a 900x900 grid with Re=40 Mach 0.5 cylinder using `fd_order: 4`:
image

You can see that we now obtain the 4th-order integral. I take this as now working, and will proceed to updating tests before calling this complete.

sbryngelson:
it certainly seems better, but remind me what we think accounts for the difference between the reference and our result

danieljvickers:
@sbryngelson This is only on a 900x900 grid. I get much less error as I converge towards larger grid sizes in this case. This is to demonstrate that it matches the python solver that I have been using for analysis. I have separate results in my slides showing the accuracy of the python solver. Rerunning at a much higher resolution will just take a few hours, but I can do that if you want me to include it before merge.

danieljvickers:
@sbryngelson Did a higher resolution sim over night. This is 3000x3000 (100 grid cells across the cylinder diameter) of the Re=40 Ma=0.1 case
image

MFC computed a drag coefficient of 1.549 compared to the reference value of 1.540 from Canuto D, Taira K. Two-dimensional compressible viscous flow around a circular cylinder. Journal of Fluid Mechanics. 2015;785:349-371. doi:10.1017/jfm.2015.635

sbryngelson:
This is worth documenting in an example and the readme since it's central to our claim that we have IBM working with the correct forces.

sbryngelson:
otherwise it seems ready to merge

danieljvickers:
@sbryngelson Done

https://github.com/danieljvickers/MFC-Dan/tree/1104-mibm-higher-order-central-differencing/examples/2D_ibm_viscous_drag_over_cylinder

sbryngelson (approved):

sbryngelson:
the `fd_order` input variable is terribly overloaded. It's used for ~3 different purposes now. This seems to be working at the moment, but it can break if you use the right (wrong) parameter set and it's not a great strategy anyway. Can you open a GH issue to disambiguate the various `fd_order`'s we use in the code somehow? FYI this pr looks good, I'll probably just let CI run then merge.

danieljvickers:
I don't think it is as clear-cut as that. When we run with `fd_order`, MFC on start up instantiated the `fd_coeffs_x[yz]` arrays that hold 5 or 3 floating point values at every grid cell, which are the coefficients taking in the local dx value.

Separating them gives more flexibility, but setting all 4 FD orders would instantiate a whole lot of memory that is somewhat wasted. The benefit of doing it the way I did was first that I did not add more case file parameters, but mostly that I am piggy-backing off already instantiated memory values.

I think a better refactor would probably be drop the coefficient arrays altogether. They are extremely extremely cheap to compute, so storing them is wasteful. I feel like it is more efficient to compute the coefficient subroutine directly inside the loop to get the coefficient values. But doing this way would mean that we can have different finite difference orders for all 4 different cases in the same case file and that the net memory usage goes down. I do not know exactly how much it costs to compute them, but my intuition and basic reasoning tells me that it is almost zero.

sbryngelson:
these input parameters are also used by post process in an entirely different manner, e.g., for schlieren i think. that's mostly what i mean. so we have a checker that checks if X postprocess parameter is true and if fd_order > 2 (for example). and setting fd_order the way you do conflicts with that.

the main problem isn't your idea, it's that MFC already overloaded this parameter. 

re: compute at runtime vs. store in array. i think keeping in array might make sense because the array is small and used in many places. DRY.

danieljvickers:
Opened the issue and copied the relevant comments. Discussion about this can move there.

PR #1615, opened 2026-06-22
IGR integration

StevenXue98:
## Description

This PR integrates IGR with more of the shared RHS/Riemann infrastructure while preserving the existing IGR numerical behavior. This is an intermediate integration milestone, not the final IGR architecture. It intentionally preserves the current IGR reconstruction/Sigma/viscous numerics while moving RHS accumulation, RK update behavior, state layout, and LF Riemann dispatch toward shared MFC subroutines.

Main changes:

- Store the final volume fraction for IGR so the IGR state layout matches the normal model-equations layout more closely.
- Route IGR face states through the shared `s_riemann_solver` path using the LF Riemann solver.
- Use shared RHS accumulation helpers for IGR:
  - `s_compute_advection_source_term`
  - `s_compute_additional_physics_rhs`
- Split the old fused `s_igr_riemann_solver` behavior into smaller IGR-specific routines called from the shared RHS path:
  - `s_igr_compute_jac_rhs`
  - `s_igr_reconstruct_cell_boundary_values`
  - `s_reconstruct_cell_boundary_values_visc_deriv`
  - `s_igr_correct_lf_fluxes`
  - `s_igr_sigma`
- Reconstruct stored IGR velocity-gradient arrays into the shared LF-facing derivative arrays, so the LF solver constructs viscous source fluxes for IGR.
- Preserve legacy IGR behavior where needed:
  - skip the normal LF MPP limiter for IGR, since it was not part of the old IGR path
  - keep IGR’s fixed-coefficient viscous-derivative reconstruction so the shared LF viscous-source construction preserves legacy IGR numerics
- Remove the special padded `rhs_vf` allocation previously needed by IGR-specific direct RHS updates.

IGR-specific face reconstruction, Sigma correction, and viscous source-flux construction are still present in `m_igr.fpp`. Subroutine `s_igr_riemann_solver` is also still present in `m_igr.fpp`, which explains the increased line count.

Deferred follow-up work:

- Evaluate whether IGR can switch from reconstruct-conservative-then-convert to the normal convert-to-primitive-then-reconstruct order, i.e. using `s_reconstruct_cell_boundary_values`. This means switching from the current cons -> reconstruction -> prim order for IGR to the cons -> prim -> reconstruction order used in the standard path, which will likely cause numerical differences. Some preliminary experiments showed differences as large as ~`1e-3` on some cases.
- evaluate whether IGR viscous derivatives can eventually be routed through `s_get_viscous` without changing numerics
- if numerically acceptable, route IGR face reconstruction and viscous derivatives through the standard path
- revisit compatibility with MHD and other currently prohibited physics

### Type of change

- Refactor

## Testing
Local testing:
```bash
./mfc.sh test --add-new-variables --only IGR
./mfc.sh test --add-new-variables --from F926448C --to F926448C
./mfc.sh test --add-new-variables --from 18A71AB2 --to 18A71AB2
./mfc.sh test --add-new-variables --from 43B5FEBD --to 43B5FEBD
./mfc.sh test --add-new-variables --from E49EF7B6 --to E49EF7B6
./mfc.sh test --only IGR
```

Additional local comparison/debug checks used during development:
```bash
python3 local_check_igr_full_alpha.py --default-igr
python3 local_igr_diagnostics/scripts/compare_igr_packs.py --default-igr
```
The IGR golden files were updated with `--add-new-variables` to include the newly stored final volume fraction. Existing golden entries were preserved, and the added final-alpha fields satisfy the expected volume-fraction sum checks, which is verified by the local Python scripts. The updated IGR tests match the previous conserved-variable outputs to machine-level roundoff, aside from the newly added final volume-fraction fields.

## Checklist

<details>

- [x] Verified CPU only
</details>

## AI code reviews


StevenXue98:
I ran some local performance comparisons against the legacy IGR path. 

Representative CPU results:

18A71AB2, 50 steps, 100 runs:
- runtime: 3.48s refactor vs 2.74s legacy (~1.27x)
- peak RSS: 71.6 MB refactor vs 57.4 MB legacy (~1.25x)

18A71AB2, 500 steps, 10 runs:
- runtime: 29.64s refactor vs 22.61s legacy (~1.31x)
- peak RSS: 71.6 MB refactor vs 57.3 MB legacy (~1.25x)

18A71AB2, larger grid, ~100^3, 20 steps, 3 runs:
- runtime: 58.54s refactor vs 55.31s legacy (~1.06x)
- peak RSS: 848.2 MB refactor vs 276.5 MB legacy (~3.07x)

RSS stands for Maximum resident set size, which is measured by the command:
/usr/bin/time -v -o "${time_file}" mpirun -np 1 "${exe}"

StevenXue98:
Regarding the Claude code review comment, it was correct that the new subroutine `s_igr_reconstruct_cell_boundary_values` is storing in this fashion: 
```
qL_rsx_vf(j + 1, k, l, i) = alpha_rho_L(i)
qR_rsx_vf(j, k, l, i) = alpha_rho_R(i)
```
This is the same convention used by `s_weno` and `s_muscl`; for example MUSCL writes `vL_rs_vf_x(j)` as the left cell boundary and `vR_rs_vf_x(j)` as the right cell boundary. I believe the reason this works is that `s(_igr)_reconstruct_cell_boundary_values` and `s_lf_riemann_solver` interpret L/R differently. In `s(_igr)_reconstruct_cell_boundary_values`, L/R face values are reconstructed relative to cell centers, but in `s_lf_riemann_solver`, L/R face values are read relative to cell interfaces. So the left cell face value of cell `j + 1` is the right state at cell interface `j + 1/2`, and the right cell face value of cell `j` is the left state at cell interface `j + 1/2`. This is why we do
```
call s_riemann_solver(qR_rsx_vf, dqR_prim_dx_n(id)%vf, dqR_prim_dy_n(id)%vf, dqR_prim_dz_n(id)%vf, qR_prim(id)%vf, &
                                  & qL_rsx_vf, dqL_prim_dx_n(id)%vf, dqL_prim_dy_n(id)%vf, dqL_prim_dz_n(id)%vf, qL_prim(id)%vf, &
                                  & q_prim_qp%vf, flux_n(id)%vf, flux_src_n(id)%vf, flux_gsrc_n(id)%vf, id, irx, iry, irz)
```
in `s_compute_rhs`. As we store the face values at stencil `j` to slot `j + 1` for the left arrays and stencil `j` to slot `j` for the right arrays, they are used in `s_lf_riemann_solver` as (shifted +1) for the right state and (unshifted) for the left state, which is consistent with the codebase's established convention.

PR #1616, opened 2026-06-22
Force reshape

### LINKED ISSUE #1452: Force Compute Reshaping
In the new IB refactor, the force arrays are written in an inefficient manner that causes slow down because they must later be reshaped. There is no particular reason for their original shape, and should be refactored accordingly. The result of this refactor should be the deletion of several lines of code, simplifying the force communication, and a negligible-to-mild performance improvement.

joshgillis:
## Description

Refactored IB force and torque storage to avoid reshaping during communication.

Forces and torques are now stored in ib_ft as (6, num_ibs), where entries 1:3 are for the force components, and 4:6 are for the torque components. Additional related arrays such as recv_forces/torques and recv_forces/torques_snap are now also combined to match the formatting, labeled as recv_ft and recv_ft_snap.

Additional changes were made to m_collisions to reflect these updates, as well as an additional GPU_UPDATE in m_time_steppers for multi-rank GPU runs.

Closes #1452 .

- Refactor

## Testing

Tested with test suite, and compared results from a non-mpi and mpi on GPU to ensure results matched.

## Checklist

<details open>

</details>

## AI code reviews


src/simulation/m_collisions.fpp
```diff
@@ -243,7 +243,7 @@ contains
         logical                                           :: already_found
 
         ! Temporary array to hold all detected pairs (with potential duplicates)
-        integer, dimension(num_gps, 2) :: raw_pairs
+        integer, dimension(2, num_gps) :: raw_pairs
```
sbryngelson: This declaration was transposed to `(2, num_gps)`, but none of the `raw_pairs` accesses were updated — they still use the old `(pair, 1|2)` layout: `raw_pairs(local_num_raw, 1|2)` (L274–275), the decode (L292–293), the in-place swap (L303–305), and the `collision_lookup` writes (L321–322), with `pair_idx`/`local_num_raw` ranging up to `num_raw` (≤ `num_gps`).

Consequences:
- The first dimension now has extent **2**, so any `pair_idx > 2` indexes out of bounds — memory corruption on the host and inside the `copy='[raw_pairs]'` GPU loop (L253).
- Even for a single pair the data is effectively transposed, so the swap/decode operate on the wrong slots.

(This looks unrelated to the force/torque reshape — likely an accidental edit.) Tests probably passed because the 2-sphere head-on example doesn't drive `num_raw > 2`. Simplest fix is to revert the declaration so it matches the unchanged accesses:

```suggestion
        integer, dimension(num_gps, 2) :: raw_pairs
```

(Alternatively, keep `(2, num_gps)` and transpose every access to `raw_pairs(1|2, p)` — but the revert is the minimal, lowest-risk fix.)

sbryngelson (commented):
The core refactor is sound — I verified it end to end. Collapsing `forces(num_ibs,3)`/`torques(num_ibs,3)` into `ib_ft(6,num_ibs)` preserves the index mapping everywhere (`forces(i,l)→ib_ft(l,i)`, `torques(i,l)→ib_ft(l+3,i)`), the MPI path is equivalent (the old `send_ft` was already `(6,num_ibs)` column-major, so packing `ib_ft` directly is bit-identical), the snap-buffer merge into `recv_ft_snap(:,j)` matches the old per-component update, and the alloc/dealloc and host/device data movement all line up. The added `m_time_steppers` host update before the periodic-wrap/ownership handoff looks correct for multi-rank GPU.

One issue: there's an **unrelated, broken change** bundled in — the `raw_pairs` declaration in `m_collisions.fpp` was transposed but its accesses weren't. Details inline. I'd treat that as a blocker; the rest is good to go once it's resolved.

joshgillis:
Sounds good, definitely an accidental change. Went ahead and reverted it. Will submit for review once I verify my CPU results.

src/simulation/m_collisions.fpp
```diff
@@ -48,12 +48,12 @@ contains
 
     end subroutine s_initialize_collisions_module
 
-    subroutine s_apply_collision_forces(ghost_points, num_gps, ib_markers, forces, torques)
+    subroutine s_apply_collision_forces(ghost_points, num_gps, ib_markers, ib_ft)
```
danieljvickers, line 51: This is a point of some disagreement in coding philosophy, but I very-much like when variables are named extremely explicity. Physicists/mathemeticians/engineers are notorious for naming variiables things like `mu_0` instead of `permeability_of_free_space`. The former saves much space when it is used in several equations, but the latter is extremely explicity about what the value is, even to non-technical users.

it is clear that `ib_ft` is "immersed boundary force and torque", but it could just as easily be something like `forces_and_torques` without causing too much clutter. But a name like this makes the code read more like a book. No one needs to go up a layer of abstraction to see what is being passed into this subroutine and understand what the array has in it. It is obvious from the name. This is a good thought exercises to participate in when naming variables. How to maximize clarity without making your equations bloated. And if you have to choose one or the other, which is more important?

src/simulation/m_ibm.fpp
```diff
@@ -1036,29 +1033,29 @@ contains
...
 
         ! reduce the forces across local neighborhood ranks
-        call s_communicate_ib_forces(forces, torques)
+        call s_communicate_ib_forces(ib_ft)
 
         ! consider body forces after reducing to avoid double counting
         do i = 1, num_ibs
             if (bf_x) then
-                forces(i, 1) = forces(i, 1) + accel_bf(1)*patch_ib(i)%mass
+                ib_ft(1, i) = ib_ft(1, i) + accel_bf(1)*patch_ib(i)%mass
```
danieljvickers, line 1019: I don't know if this was intentional, but you did a good job of respecting fortran's column-major ordering for memory layout. Something I forgot to consider in my original implementation.

src/simulation/m_ibm.fpp
```diff
@@ -908,7 +906,7 @@ contains
         type(scalar_field), dimension(1:sys_size), intent(in)          :: q_prim_vf
         type(physical_parameters), dimension(1:num_fluids), intent(in) :: fluid_pp
         integer                                                        :: i, j, k, l, encoded_ib_idx, ib_idx, ib_idx_temp, fluid_idx
-        real(wp), dimension(num_ibs, 3)                                :: forces, torques
+        real(wp), dimension(6, num_ibs)                                :: ib_ft
```
danieljvickers, line 909: This would be a good place to put a comment about the contents of this array. Like I said, I already think it would be best renamed. But it is not clear immediately to a reader how memory is written here. A comment briefly describing that it is f_x, f_y, f_x, tau_x, tau_y, tau_z contiguously would make it easier for anyone reading the code to understand how they are supposed to fill in the array. Although the array bounds help clarify this a little for you.

danieljvickers (commented):
This looks good. I see you were not able to delete one of the array assignments on the GPU, but it is very clear why you had to leave it. When my main review comments are about naming conventions, it is a good day. I do think the name should be considered being changed. But that is about it. Otherwise, this is pretty much how it should be added.

danieljvickers:
@joshgillis I saw you had a single failure on Phoenix that looked spurious to me. I reran just that test. If it fails in the same manner, it is something in your branch. If it passes, ping me here and I will restart the rest of the tests

Edit: Rerun still has the failing. So this is a real failure somehow. But out of bounds errors are some of the easiest to track down. Phoenix uses NVHPC/24.5.

src/simulation/m_collisions.fpp
joshgillis: @danieljvickers I see, I definitely fall into the pattern of saving space, but I also agree that sometimes gets messy when it comes to legibility. What about ib_forces_torques for the name? I wanted to preserve the "ib" part of the name just so it doesn't get confused with other force/torque arrays in m_collisions if they are added later on, unless if you don't see that being necessary.

Edit: Do you also want this to extend to recv_ft and recv_ft_snap?

src/simulation/m_ibm.fpp
joshgillis: Funny, I was looking into array shaping and saw the column-major preference Fortran has and thought that is what you had been implying with "no particular reason for their original shape" in your issue. Glad I could make an improvement.

sbryngelson:
closed due to inactivity

PR #1618, opened 2026-06-24
Fix periodic ib issues

danieljvickers:
## Description

I was made aware of several bugs in the IB MPI communication in periodic cases. This branch resolves all of the issues that I have been made aware of. They include

1. periodic IBs not having the radial vector account for their periodicity
2. Force integration can double-count if the Ib neighbors itself or has the same neighbor periodically
3. Collisions can be missed as neither processor believes they own the collision
4. There is drift in IB information between ranks.

The good news is that the fix for number 2 is also a computational optimization in small cases by limiting the total number of communications to the minimum required to successfully sum. So we have a slight upside.

A new example will be added that should check for this behavior and ensure that it does not break again.

- Bug fix


danieljvickers:
Fixed all of the bugs that I have been made aware of, and now all minimum-reproducers I was given behave as anticipated.

Final list of bugfixes:
1. Updated the n2 collision lookup to take into account periodicity
2. Added an NVHPC/24.5- compiler workaround for the moment of inertia calculation 
3. Reduced number of communications to the minimum required to accomplish the correct summation, which prevents periodic cases from wrapping back on themselves and double-counting
4. Uses the correct periodic location when computing the radial vector in various subroutines.

Will add an example to make sure we don't regress on this.

danieljvickers:
Did a lot of debugging here and determined that this is a machine-precision issue with the levelset. If I compute the levelset on the CPU always and gopy to the GPU, the GPU test passes the test suite. This makes me quite sad. The test is pretty poorly posed. I can get around this for now, but it is another time that this issue has cropped up.

danieljvickers:
@sbryngelson can I get approval for case op tests?

sbryngelson:
you only need approval for benchmark not case opt. though i think case opt only runs once your other tests pass (automatically regardless).

sbryngelson (approved):

sbryngelson:
nice work. bench is clean. code is clean. merging.

PR #1620, opened 2026-06-30
feat(validate): add --all flag to validate all example case files

### LINKED ISSUE #1529: All-File Validation for Toolchain
### Problem it solves
When one runs the `precheck` command in the toolchain, one of the steps runs `validate` on all examples and reports the number of examples that fail, but not which examples. This requires the used to check cases one-at-a-time manually. One can accomplish this with BASH commands, but it would be much nicer if the validator had something like an `--all` flag and just tries to validate all examples.

### Proposed solution

1. Add an `--all` flag to validator that checks all examples.
2. The flag should accurately report which cases fail validation
3. the precheck for commits should swap from what it uses now and instead use this command
4. (Optional) Validation is also very slow, which leads me to always commit with `--no-verify`. It would be nice if it were much faster.

### Alternatives
You could also simply let the `validate` command accept wildcard values, like `*`, or directories and the user could simply pass in something like `./mfc.sh validate examples` or `./mfc.sh validate examples/*` to accomplish the same thing. But this is likely much harder to code this than the `--all` flag.

adityaraut09:
Closes #1529

Adds an `--all` flag to `./mfc.sh validate` that loops through all example case files and reports which ones pass or fail, rather than requiring the user to check them one at a time manually.

### Type of change
- New feature

### Changes
- `toolchain/mfc/cli/commands.py`: made `input` positional optional (`nargs="?"`), added `--all` / `-a` flag to `VALIDATE_COMMAND`
- `toolchain/mfc/validate.py`: added `_validate_one()` helper and `--all` branch that globs `examples/**/case.py`, validates each, and prints a pass/fail summary

### Testing
Verified syntax with `python3 -c "import ast; ast.parse(...)"`. Full integration test requires the MFC build environment.

### DIFF
--- a/toolchain/mfc/cli/commands.py
+++ b/toolchain/mfc/cli/commands.py
@@ -547,6 +547,8 @@ VALIDATE_COMMAND = Command(
         Positional(
             name="input",
             help="Path to case file to validate.",
+            nargs="?",
+            default=None,
             completion=Completion(type=CompletionType.FILES_PY),
         ),
     ],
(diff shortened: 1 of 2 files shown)


sbryngelson:
closing (spam)

PR #1621, opened 2026-06-30
Feature/only case insensitive

### LINKED ISSUE #1533: Remove test --only case sensitivity
When wee run test with the `--only` flag, it is case sensitive. This is slightly frustrating because flags include "STL", "IBM", "Sphere", "Cuboid", "slip", etc. Since there is no convention in the way test labels are captialized, it requires the used to remember the specific capitalization for the test that they want to run. It would make much more sense to not be case-sensitive.

This issue can be solved by modifications to the toolchain for the `./mfc.sh test --only <label>` feature. When complete, the test suite should run the same with various spellings: `STL`, `Stl`, `stl`, sTl`, etc.

RuTh-git:
## Description
The `--only` flag in `./mfc.sh test --only <label>` performed a case-sensitive 
string comparison against test labels. Since MFC's test labels follow no 
consistent capitalization convention (e.g. `STL`, `slip`, `Sphere`), users had 
to remember the exact casing of every label to filter tests correctly. This PR 
normalizes the label comparison to lowercase so any casing works.

Closes #1533.

- Bug fix

## Testing
Manually verified the fix by running the following before and after the change:
- `./mfc.sh test --only stl -j 1` → previously failed with "matched zero test 
  cases", now passes
- `./mfc.sh test --only STL -j 1` → passes (unchanged)
- `./mfc.sh test --only Stl -j 1` → previously failed, now passes
- `ruff check toolchain/mfc/test/test.py` → all style checks passed

## Checklist

### DIFF
--- a/toolchain/mfc/test/test.py
+++ b/toolchain/mfc/test/test.py
@@ -66,7 +66,6 @@ def _filter_only(cases, skipped_cases):
         check = set(case.trace.split(" -> "))
         check.add(case.get_uuid())
 
-        #label_ok = all(label in check for label in labels) if labels else True
         label_ok = all(label.lower() in {c.lower() for c in check} for label in labels) if labels else True
         uuid_ok = any(u in check for u in uuids) if uuids else True
 
--- a/toolchain/mfc/test/test.py
+++ b/toolchain/mfc/test/test.py
@@ -66,7 +66,8 @@ def _filter_only(cases, skipped_cases):
         check = set(case.trace.split(" -> "))
         check.add(case.get_uuid())
 
-        label_ok = all(label in check for label in labels) if labels else True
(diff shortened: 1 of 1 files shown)


sbryngelson:
@RuTh-git copilot flagged meaningful problems with this PR.

danieljvickers:
Ooh, I look forward to this being merged. The copilot review on this seems correct that it may be worth just pre-generating a `casefold` list for this once and comparing off of that. It makes the PR touch a few more lines of code, but makes the actual checking line slightly cleaner. I don't know how the convergence testing flag works or about the other comment, but I definately trust @sbryngelson's opinion on that one.

sbryngelson:
closed due to inactivity [will fix myself if necessary]

PR #1628, opened 2026-07-03
Block-structured AMR with batched fine advance, load-balance and active-box infrastructure

sbryngelson:
## Summary

Block-structured adaptive mesh refinement for MFC, with the load-balance and active-box infrastructure it was built on. All of it is default-off: a case without `amr = T` runs the same code as master, and every golden file that exists on master is unchanged.

- **Block-structured AMR** (`src/simulation/m_amr*.fpp`, 12.5k lines): a 2:1 refined block hierarchy of up to `amr_max_level` levels over the base grid, Berger-Rigoutsos clustering on a density-gradient tag field, conservative restriction and conservative-linear prolongation with physics-specific closures, flux registers and refluxing, dynamic regrid on a fixed interval, restart. Blocks live in one device-resident flat store. Fine blocks advance in batches of equal shape (one RHS call per batch), every level in lock-step at the case `dt`. Every inter-rank transfer is a posted wave on one engine (`m_amr_wave`): keyed tags, per-peer aggregation, device-resident wire pools under `rdma_mpi`, and an identity-header audit under `MFC_DEBUG`.
- **Load-balance infrastructure** (`m_box`, `m_load_weight`, `m_sfc_partition`, `m_load_balance`): partition arithmetic, a per-cell load-weight field, Morton space-filling-curve partitioning and tile migration for the base grid.
- **Active-box windowing** (`m_active_box`): restricts reconstruction, Riemann and RK windows to a light-cone-grown box around non-ambient flow.

User documentation: `docs/documentation/amr.md`. Internals: `docs/documentation/amr_implementation.md`. The performance campaign's records, design memos and what was learned: `misc/amr_ledger/README.md`.

## Physics support

The batched fine advance is the only fine advance. It runs on uniform Cartesian grids for the 5- and 6-equation models, MHD and relativistic MHD (1D), hypoelasticity and continuum damage, Lagrangian bubbles, chemistry (with and without species diffusion), IGR and static or prescribed-motion immersed bodies. Subcycling, stretched or cylindrical grids, QBMM, Euler bubbles, surface tension, phase change and moving particle clouds are rejected by the case validator with a named message, never silently. The per-block fine advance that served those (correct, golden-tested, about 1.2x slower per step) lives on `sbryngelson/MFC:amr-per-block` for a follow-on PR.

## Correctness evidence

- Conservation is reflux-exact (roundoff defects) across single- and multi-fluid, viscous, Lagrangian-bubble and chemistry cases, through regridding advances; free-stream preservation is exact with regrid armed, including 3D, IGR and Lagrangian configurations.
- np=1 and np=2 agree element-exactly with blocks spanning rank seams, chemistry species included.
- Every exchange family carries keyed wire headers under `MFC_DEBUG`, checked at unpack; a seeded-fault control proves the check can fail.
- 69 AMR/L0-tiling/active-box goldens (14 of them multi-level, up to np=8 dynamic regrid). On the final tree every CI lane is green: all 15 Frontier lanes pass (CCE cpu, gpu-acc and gpu-omp; AMD cpu and gpu-omp; the Case Opt and Bench lanes for both), alongside the GitHub, NVHPC and Phoenix lanes. Locally on Frontier, the 64 AMR/L0/active-box goldens pass 64/64 on CCE gpu-acc, gpu-omp and CPU.

## Code shape (for reviewers of the AMR modules)

The AMR code is a chain of modules, lowest first, each using only the ones before it: `m_amr_state` (all module-scope state, `GPU_DECLARE`s, leaf helpers), `m_amr_distribution` (rank decomposition, box arithmetic, SFC cut, ownership), `m_amr_wave` (the exchange engine: plan sides, pools, keyed tags, audit records, the wait), `m_amr_store` (flat-store slot machinery, bridge loaders), `m_amr_exchange` (the fill waves, seams, ghost fills), `m_amr_frame` (the grid-state swap), `m_amr_transfer` (prolong, restrict, reflux and their waves), `m_amr_advance` (the batched fine stage advance, IB/Lagrange fine services), `m_amr_l0` (level-0 tiling), `m_amr_stage` (the four hooks the RK stage loop calls); `m_amr` is the init/finalize driver and re-exports them all. `m_amr_regrid`, `m_amr_registers`, `m_amr_restart` and `m_amr_xchg_audit` sit beside the chain; `src/common/m_amr_restart_io` is the restart layout shared with post_process.

What the solver itself sees: `m_time_steppers` calls `s_amr_stage_begin` / `s_amr_stage_fine` / `s_amr_l0_stage_update` / `s_amr_step_fold`; `m_rhs` carries the flux-register capture and the batched slab's reconstruction window; `m_mpi_common` gains the box-based equal split (byte-identical to the inline arithmetic) and a few reductions. `m_riemann_solver_hllc` shows +1408/-1371: that is the kernel restructuring the batched slab needed (per-cell bodies hoisted so one launch serves a slab of blocks), with every golden unchanged.

The series of cleanups before review (each gated on the same goldens, CPU and GPU, byte-identical): one exchange engine for the seven waves; the fine advance reduced to the batched lock-step path; one gather mechanism (init and the regrid rebuild fill coarse patches with the same fill waves the step uses); a twelve-phase step budget under `rank_time_wrt` in place of the campaign's instrumentation; the stage hooks pulled out of `m_time_steppers`; the two largest routines split; dead machinery and stale comments removed; then a de-duplication pass (-1,300 lines) that gave the code one of each thing: one flux-register capture kernel for both register families, one L0 tile box router, one restriction body for its three targets, one stable key sort, one receive drain for the fill waves, one request routine per wire buffer kind, and an `amr_dim` dimension mask plus (x, y, z) Fypp lists in place of hand-copied 1D/2D/3D branches. Parameters are defined in `toolchain/mfc/params/definitions.py` and constrained in `case_validator.py`; the batched-advance admissibility rules live there.

## Performance

Measured on one node of eight AMD MI250X GCDs, OpenMP offload, one rank per GCD, double precision in both codes. The comparable quantity between MFC and AMReX is each code's excess over its own uniform-grid run at the same number of advanced cells:

```
excess = AMR s/step - uniform s/step x (cells advanced per step / base cells)
```

On a 400^3 base with two refinement levels, regrid every 20 steps, RK3, an advected density blob (AMReX: `Tests/GPU/CNS` on the matched deck), same node, same session, three reps per arm, control and treatment interleaved:

| | excess (s/step) | ratio |
|---|---|---|
| MFC (this PR) | 0.42 (sd 0.02) | 1.08x |
| AMReX CNS | 0.38 to 0.39 (sd 0.02) | |

The instrument's floor is about 0.03 s/step (control-to-control on the same node). Relative to each code's own solver, MFC's AMR machinery inflates its step by 1.78x and AMReX's by 1.86x. MFC's AMR step is slower than AMReX's in absolute terms (0.96 vs 0.82 s/step) because its base scheme (WENO5 with mapped weights and monotonicity preservation, HLLC, the 5-equation model) costs 1.7x per cell, which is a scheme choice and not an AMR term. The ratio depends on the node: the same binary reads up to 1.9x on nodes with throttling GPUs or dead InfiniBand ports, so every figure above was read with its control on the same node.

Weak scaling from 8 to 32 GPUs at fixed work per rank is 1.28x per doubling against AMReX's 1.23x on matched cross-node rungs; the growth is the base grid's cross-node halo, not the regrid family.

Where the remaining 0.42 s/step goes: about a fifth is fine-solver inflation (ghost planes per block, batch padding, dispatch floor), about half is MPI wait set by the two heaviest ranks at roughly fourteen rendezvous per step, and the rest is regrid, restriction, gather and reflux work. The levers that were tried, what each bought, and what was found to be false are in `misc/amr_ledger/README.md`.

## Known issues

- Subcycling, cylindrical and stretched grids, QBMM, Euler bubbles, surface tension, phase change and moving particle clouds are not supported under `amr` on this branch (validator-rejected); the per-block advance that supported all but surface tension is on `amr-per-block` for a follow-on PR.
- CCE 19.0.0 cannot compile two or more of the batched reflux loop nests in a single procedure, and fails differently per target: the offload lanes abort with `ftn-7991 "PDG node N requires copy"` (`pdgcs/v_pdg.c`), while the CPU lane compiles and silently emits bad code that segfaults as soon as refluxing runs. `s_amr_apply_reflux` therefore keeps one nest per procedure (`s_amr_apply_reflux_d1/d2/d3`, Fypp-generated from one body), which compiles correctly at full optimization on all three lanes -- no `-Oipa0` exemption. Folding them back together reintroduces both failures at once. Bounds checking finds nothing and a debug build passes, because there is no out-of-range access in the source.
- `eos_state_dependent` is a compile-time constant in every build (`Case.get_fpp`'s `_prepend`, like `chemistry`), so a Mie-Gruneisen, JWL or Vinet fluid moves the build slug of every target. The Frontier AMD gpu-omp lane pre-builds a fixed variant list rather than enumerating them (each amdflang device link is ~1 h), so it gains a third `eos` variant, and upstream's `lint_test_suite` gate knows about it. `./mfc.sh test --no-build` now verifies that each selected case's binaries exist instead of silently skipping the build, so a future compile-time axis fails in the lint gate rather than as an `execve()` on a missing path deep in a cluster run.


sbryngelson:
**Upstream latent gap found during the MHD+AMR investigation** (independent of this PR): `m_riemann_solver_hlld.fpp` contains no GLM terms, so `hyper_cleaning = T` with `riemann_solver = 4` (HLLD) is silently inert — psi never evolves and no cleaning occurs. The existing hyper_cleaning golden uses HLL, which does carry the GLM coupling, so CI never sees this. Worth either adding GLM terms to HLLD or a validator prohibit on the combination.

Also for the record: MHD+AMR was attempted and re-gated on measured evidence rather than assumption — the coarse/fine seam is a continuous O(1) div(B) source that cleaning spreads but cannot remove (details in the amr.md support matrix row and commit ac203b1b).

sbryngelson:
### `active_box` is single-rank-only — make it MPI-real, or defer it?

Raising this as a self-review point. `active_box` disables itself with a warning at `num_procs > 1` and falls back to full-domain compute. Since production runs here are essentially always multi-rank, that makes the optimization a no-op (plus a warning line and some maintenance surface) in exactly the configurations that matter. We should decide what to do with it.

**Why it's single-rank today.** The active box is a single global, coarse-cell-indexed region. On one rank you just clip the RHS loops (convert / reconstruct / Riemann) to it. Under domain decomposition the *geometry* is easy — each rank intersects the global box with its subdomain and freezes the exterior — but it creates a **load-imbalance** problem: if the disturbance occupies only a few ranks' territory, those ranks do all the work while the rest idle, and the slowest rank sets wall-clock, so you save nothing. The current code sidesteps this by disabling rather than solving it.

**The enabling pieces are already in this PR.** A multi-rank active box needs (a) per-rank clipping of the box (straightforward), and (b) a way to keep otherwise-idle ranks busy — i.e. redistribute the domain so the active region spreads across ranks. That second part is exactly what this PR's load-balance arc (`m_load_balance` / `m_sfc_partition` / `m_load_weight`) provides, and the active box already produces the natural work-weight signal (active vs. frozen cells) that a rebalancer would consume. The two features are built to compose — active box marks *where* the work is, load-balance *moves ranks to it* — they're just not wired together yet.

**Options:**
1. **Make it MPI-real.** Per-rank box clipping + feed the active/frozen cell count into the load-weight so `load_balance` redistributes toward the active region. This is the version that actually helps at scale, and it's the natural payoff of shipping the load-balance infrastructure in the same PR.
2. **Defer / gate it.** Drop `active_box` from this PR (or keep it behind an explicit experimental flag with the single-rank limitation documented up front) and keep the performance family focused on AMR, which is already multi-rank-capable via the P2P device-path gather/scatter. Re-introduce active box once the load-balance integration is designed.

Leaning toward (1) if there's appetite, since the enabling infra is right here; otherwise (2) keeps the PR's surface honest — a default-off single-node optimization advertised in an exascale code invites confusion. At minimum, the docs should state plainly that it is single-rank-only today.

PR #1632, opened 2026-07-05
Add AMD HPCFund (amdfund) amdflang OpenMP-offload build/run support

sbryngelson:
## Description

Adds first-class support for building and running MFC on **AMD HPCFund** (`hpcfund`,
MI250X / gfx90a) with the AMD AFAR **LLVM Flang** (`amdflang`) compiler using OpenMP
`target` offload against the system GNU-built OpenMPI.

- **New `amdfund` load slug** (`toolchain/modules`, `toolchain/bootstrap/modules.sh`):
  loads `gnu12 openmpi4/4.1.8 cmake rocm/7.2.0` and sets `FC=amdflang` plus the MPI and
  hipfort paths. `OLCF_AFAR_ROOT` uses the direct WORK-filesystem path (mounted at the
  same point on login and compute nodes) and is overridable — `export OLCF_AFAR_ROOT=...`
  before loading to select a different AFAR drop.
- **New `amdfund.mako` run template**: SLURM, interactive + batch. hpcfund's Slurm has
  `GresTypes=null`, so no `--gres`/`--gpu-bind` is emitted (the MI250X nodes hand out
  every GCD with the node); default partition `mi2508x`.
- **`cmake/MFCTargets.cmake`** (refactor):
  - Resolve `hipfort-amdgcn` (library + Fortran module include) under the newer AFAR
    *therock* `lib/llvm/` layout in addition to Frontier's `lib/` layout.
  - Feed the amdflang (`LLVMFlang`) path its MPI/hipfort locations through compiler-neutral
    `MFC_FLANG_MPI_INC/LIB` and `MFC_FLANG_HIPFORT_LIB` env vars, **falling back to the
    legacy `CRAY_*` names** when the neutral ones are unset. These hooks are consumed only
    by the `LLVMFlang` branch (never the Cray CCE compiler), so the naming was misleading;
    the fallback keeps the existing Frontier `famd` (amdflang + Cray MPICH) build working
    unchanged.

**Note on the default `OLCF_AFAR_ROOT`:** hpcfund has no shared, writable software
location, so the committed default points at a concrete AFAR install path; it is
overridable per the above. This mirrors how the existing Frontier `famd` slug hard-codes
its AFAR path.

No solver source (`src/`) changes — this is toolchain and build-configuration only.

### Type of change

- New feature
- Refactor

## Testing

On hpcfund (MI250X, gfx90a, AFAR drop 23.2.1, flang 23.0.0git):

- `source ./mfc.sh load -c amdfund -m g && ./mfc.sh build -t simulation --gpu mp -j 8`
  configures, builds, and installs the `simulation` target (exit 0, ~23 MB binary).
- Verified both env paths resolve correctly: the legacy `CRAY_*` fallback and the neutral
  `MFC_FLANG_*` vars. Confirmed the `simulation` link line carries `libhipfort-amdgcn.a`,
  `libhipfft.so`, `libmpi.so`, and `libmpi_mpifh.so`.
- `./mfc.sh precheck` passes all checks.
- Frontier `famd` is preserved by construction (its `CRAY_*` block is untouched and the
  CMake fallback reads it); not re-run on Frontier hardware in this change.

## Checklist

- [ ] I added or updated tests for new behavior — N/A (build-config/toolchain only)
- [x] I updated documentation if user-facing behavior changed — `./mfc.sh load` help text
  and system menu updated with the new `amdfund` slug

### DIFF
--- a/cmake/MFCTargets.cmake
+++ b/cmake/MFCTargets.cmake
@@ -106,6 +106,10 @@ exit 0
                 set(_flang_mpi_lib "$ENV{CRAY_MPICH_LIB}")
             endif()
             if(CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang" AND NOT "${_flang_mpi_inc}" STREQUAL "")
+                if("${_flang_mpi_lib}" STREQUAL "")
+                    message(FATAL_ERROR "amdflang MPI include path is set but its link libraries are not. "
+                        "Set both MFC_FLANG_MPI_INC and MFC_FLANG_MPI_LIB (or both CRAY_MPICH_INC and CRAY_MPICH_LIB).")
+                endif()
                 target_compile_options(${a_target} PRIVATE "${_flang_mpi_inc}")
                 target_link_libraries(${a_target} PRIVATE ${_flang_mpi_lib})
             else()
--- a/cmake/MFCTargets.cmake
+++ b/cmake/MFCTargets.cmake
@@ -96,10 +96,18 @@ exit 0
             find_package(MPI COMPONENTS Fortran REQUIRED)
 
             target_compile_definitions(${a_target} PRIVATE MFC_MPI)
-            if(CMAKE_Fortran_COMPILER_ID STREQUAL "LLVMFlang" AND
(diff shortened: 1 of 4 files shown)


sbryngelson:
@claude full review

PR #1641, opened 2026-07-11
toolchain: fix KeyError('num_fluids') in case-optimization fpp generation

sbryngelson:
## Summary

`Case.__get_sim_fpp` (`toolchain/mfc/case.py`) reads `self.params["num_fluids"]` by direct subscript when emitting the case-optimization Fypp header, while every other parameter in that same block uses `self.params.get(key, default)`.

The packer constructs a **dummy case with empty params** to enumerate configured targets for the golden metadata file:

```python
# toolchain/mfc/packer/pack.py
case = input.load(None, {}, {})
for target in get_configured_targets(case):
    ...
```

Under `--case-optimization`, that empty-param case flows through `get_fpp` → `__get_sim_fpp`, and the bare subscript raises `KeyError: 'num_fluids'`.

## Impact

`./mfc.sh test --case-optimization` fails for **every** golden case with `Error: 'num_fluids'` — even though the case-optimized simulation itself builds and runs correctly. The failure is in the post-run golden-packing/metadata step, not the compiled binary (the run completes with exit 0 and correct output before the packer is invoked).

CI's case-optimization job runs `./mfc.sh run --case-optimization` + a custom output check (`.github/scripts/run_case_optimization.sh`), **not** the golden harness, so it never hits this path — which is why it went unnoticed.

## Fix

One line, matching the `.get(..., default)` pattern of every sibling in the same block (`nb`, `weno_order`, `muscl_order`, …):

```diff
-#:set num_fluids            = {int(self.params["num_fluids"])}
+#:set num_fluids            = {int(self.params.get("num_fluids", 1))}
```

For the dummy empty case the default is only used to compute a throwaway staging slug (which won't match any real build and is skipped); for real cases `num_fluids` is present, so behavior is unchanged.

## Testing

- **Reproduced:** `./mfc.sh test --only <AMR golden> --case-optimization` → `KeyError: 'num_fluids'` before the change; passes after.
- **Verified on GPU (V100)** with both OpenACC and OpenMP offload: two AMR multi-level goldens pass under `--case-optimization` (2/2 on each backend) after the fix.


sbryngelson:
apropos of little

> The fix itself → closing #1641 does have a (small, real) consequence: it abandons a correct 1-line bug fix and leaves ./mfc.sh test --case-optimization broken (KeyError: 'num_fluids') on master.
> 
> Recall what #1641 actually fixes: case.py:394 reads self.params["num_fluids"] by bracket while every sibling uses .get(...). The packer builds an empty-params dummy case, so under --case-optimization that path KeyErrors — I hit it live earlier this session. It's a genuine latent bug, present on master right now.
> 
> The nuance:
> - It's niche. CI's case-opt (and the normal workflow) use ./mfc.sh run --case-optimization + a custom validator, which never touches the golden-packer path. So CI doesn't exercise the bug — which is exactly why it went unnoticed.
> - But it's a real dev-facing break. Anyone who runs the golden test harness with --case-optimization (a legitimate way to check case-opt against goldens) hits the KeyError immediately.

PR #1647, opened 2026-07-16
Operator-split α-QSS sub-stepped chemistry integrator with per-rank adaptive substepping

sbryngelson:
## Summary

Adds an operator-split integrator for the chemistry reaction source, run as a fixed number of sub-steps per flow step (`chem_params%reaction_substeps`). The sub-step uses an **α-QSS** (Mott/CHEMEQ2 predictor–corrector) update on a constant-(ρ,e) reactor — matrix-free, Jacobian-free, and GPU-friendly (deterministic per-cell work, no thread divergence). This keeps stiff mechanisms stable where an explicit source in the flow RHS diverges (e.g. GRI-Mech methane).

On top of that, an optional **per-rank adaptive sub-step count** (`chem_params%adap_substeps`): each rank sizes its `nsub` from a local max-reduction of a stiffness metric over its own cells (floor `reaction_substeps`, ceiling `reaction_substeps_max`). **Zero MPI collectives** — bottleneck-equivalent to a global reduction since the slowest rank dominates anyway, but safe at scale. Default off, so existing behavior is unchanged.

## Dependency

The integrator calls `get_creation_destruction_rates` (a single-pass routine filling creation + destruction in one concentration/rate evaluation, halving the rate-of-progress work per sub-step), added in **pyrometheus/pyrometheus#116** and released in **pyrometheus 1.1.1**. `toolchain/pyproject.toml` pins `pyrometheus == 1.1.1` — verified end-to-end that MFC installs it from PyPI and builds against it.

## Validation

- Chemistry regression suite: only the sub-stepped reactor golden (`5D59DDF0`) shifts (forward-Euler → α-QSS), max ~0.5% and confined to trace radicals; all `reactions=F`/`substeps=0` goldens unchanged. Golden regenerated for α-QSS.
- Adaptive `nsub` verified to track stiffness across [floor, ceiling] (floor in quiescent gas, ceiling only at the reaction front).
- 1D reactive shocktube + 2D cellular/expanding/colliding detonations run stably at CFL ~0.5 (≈8× the `substeps=0` stability limit); `substeps=0` at that dt NaNs.
- Builds + runs CPU and GPU (nvfortran/OpenACC); adaptive-off is bit-identical to the fixed-sub-step path by construction.

## Files
- `src/common/m_chemistry.fpp` — α-QSS integrator + adaptive probe
- params/checker plumbing: `m_derived_types.fpp`, `toolchain/mfc/params/definitions.py`, `m_global_parameters.fpp` (×3), `src/simulation/m_checker.fpp`
- `src/simulation/m_rhs.fpp`, `m_time_steppers.fpp` — operator-split call site
- examples: `2D_reactive_shock_bubble`, `1D_reactive_shocktube_adaptive`, `2D_detonation_cell` (the latter two demonstrate the automated `cfl_const_dt` + adaptive pipeline)
- sub-stepped reactor test + docs


sbryngelson:
Thanks for the review — addressed in `40420e1` and `d4d0af9`.

**Fixed**
- **`Ysum` division-by-zero in the corrector** — good catch. Guarded: when every species clips to zero, fall back to the sub-step's starting composition `y0` (already normalized and mass-consistent) instead of dividing by a vanishing sum.
- **Docs said "forward-Euler"** — corrected in `case.md` to describe the α-QSS (Mott/CHEMEQ2) predictor–corrector with creation/destruction splitting and the Padé α-weighting, and documented `adap_substeps` / `reaction_substeps_max`.
- **Stale PR description / dependency pin** — the description was written when this was a draft blocked on pyrometheus#116. That PR is now merged and released as **pyrometheus 1.1.1**; `toolchain/pyproject.toml` pins `== 1.1.1` (kept exact, matching the repo's convention, for reproducibility). Verified end-to-end that MFC installs 1.1.1 from PyPI and builds against it. Description updated.
- **Golden generated from a dirty tree** — regenerated `5D59DDF0` from a clean commit in a release (`--no-debug`) build; the values are byte-identical, only the provenance metadata is now deterministic.

**Respectfully keeping as-is (with rationale)**
- **`dimension(10)` in the `not MFC_CASE_OPTIMIZATION and USING_AMD` branch** — this follows a pervasive existing MFC idiom (the same fixed-size-10 fallback appears in ~7 places across `m_chemistry.fpp`, `m_variables_conversion.fpp`, `m_cbc.fpp`, `m_hyperelastic.fpp`); it's an AMD-flang automatic-array workaround. Case-optimized builds (the norm, and required for any >10-species mechanism like GRI-Mech anyway) take the `#:else` branch and size to `num_species` correctly. Changing the constant is a codebase-wide decision that risks the AMD build, so it's out of scope for this PR — happy to file a separate issue to unify these.
- **Defaulting `reaction_substeps_max` to the floor** — this would make `adap_substeps = T` a silent no-op (nsub could never rise above `reaction_substeps`, i.e. adaptivity that never adapts), which is a subtler footgun than the current behavior. Enabling adaptivity is a deliberate choice, so the checker fails loud with an actionable message (`reaction_substeps_max must be >= reaction_substeps when adap_substeps = T`). I'd rather keep that explicit.

PR #1653, opened 2026-07-17
fix(phase_change): bound Newton relaxation loops by max_iter

sbryngelson:
## Description

The three Newton solvers in `src/common/m_phase_change.fpp` — `s_infinite_pt_relaxation_k`, `s_infinite_ptg_relaxation_k`, and `s_TSat` — looped on a convergence-only `do while` condition with **no iteration limit**. The module parameter `max_iter` was declared but **never referenced anywhere** (grep-confirmed): it has been dead since the original phase-change PR (#179), which declared it — and even added it to the `!$acc declare` list — but never wired it into the loop conditions. A later cleanup (#999) removed the stale GPU declaration while migrating to the macro, leaving a fully orphaned declaration. So it is a latent birth defect, not a regression.

When a cell fails to converge, the loop never exits and the run hangs indefinitely (on GPU the kernel never returns and the host blocks in `cuStreamSynchronize`). Reproduced with the stock, unmodified `examples/2D_phasechange_bubble`, which hangs at `t_step = 8` independent of compiler, MPI, and grid resolution.

Root cause of the non-convergence itself (tracked as a separate follow-up): `s_TSat` and the pTg solver compare a *dimensional* Gibbs-free-energy residual (natural magnitude ~10²–10³) against the fixed tolerance `ptgalpha_eps = 1e-2` under heavy `Om = 1e-3` underrelaxation, so for low-pressure vapor cells the exit condition is effectively unreachable. This PR does **not** change that convergence behavior — it only removes the ability to hang.

This change:
- Bounds each of the three loops with `if (ns >= max_iter) exit`, accepting the last iterate on exhaustion. An unbounded hang becomes a bounded, finite step.
- Reduces the misleading real-literal initialization `max_iter = 1e8_wp` (an `integer` initialized from a real literal) to a clean integer `100000`, comfortably above the observed legitimate convergence ceiling.

Closes #1646.

### Type of change

- Bug fix

## Testing

- All 18 phase-change regression tests (`Phase Change model 5/6`, 1D/2D/3D, 2/3 fluids, model_eqns 2/3) pass **byte-identical** against the existing golden files. This confirms the cap sits above the legitimate convergence ceiling and truncates no converging cell — one case (2D, model 6) legitimately iterates deep (~145 s vs ~7 s for its siblings) yet still matches bit-for-bit, so the cap is neither too low (clipping valid slow cells) nor merely cosmetic.
[...]

PR #1654, opened 2026-07-17
fix(phase_change): globalize pTg/TSat Newton solvers so they converge (esp. GPU)

sbryngelson:
## Description

Follow-up to #1653 (fyi to @JRChreim) That PR bounded the phase-change Newton loops so a non-convergent cell can no longer hang the solver. This PR fixes the **underlying non-convergence** so those cells actually reach equilibrium — and does so with bounded, uniform iteration counts, which matters most on GPU.

**Root cause of the non-convergence.** Instrumenting the stock `examples/2D_phasechange_bubble` showed the non-convergers were `s_TSat` and `s_infinite_ptg_relaxation_k` (the pT solver was fine). Both applied a fixed `Om = 1e-3` underrelaxation that *stalled* the iterates far from the root — e.g. the pTg Gibbs residual froze at ~1e3 for the low-pressure vapor cells and never moved. The convergence criterion was never the problem; the solver was.

### `s_TSat` — reformulated + safeguarded
The saturation temperature is the root of Gibbs equality `g_l = g_v`, which for stiffened gases is the **dimensionless, strictly-monotone** relation
```
G(T) = A + B/T + C·ln(T) + D·ln(p_sat + p_inf,l) − ln(p_sat + p_inf,v) = 0
```
The `A/B/C/D` coefficients were already computed in `s_initialize_phasechange_module` — and previously **dead** (referenced nowhere). This PR solves `G(T)=0` with a safeguarded Newton (rtsafe: Newton step when it stays bracketed and reduces the interval, bisection otherwise). `G` is strictly increasing over `[Tsat_min, TCr]` for a liquid–vapor pair, so the root is unique; the solver cannot diverge and needs no accurate initial guess. Out-of-window pressures clamp gracefully.

### `s_infinite_ptg_relaxation_k` — globalized
Same analytic 2×2 Jacobian, but the fixed underrelaxation is replaced by a **damped Newton with backtracking line search**, every step **projected onto the physical bounds** `0 ≤ ml ≤ mT` and `pS > pmin`, plus a singular-Jacobian guard. The residual evaluation is extracted into a device helper `s_compute_ptg_residual` so trial states can be evaluated without mutating `q_cons_vf` (the total reacting mass `mT` is conserved by construction, so `ml` is the only mass unknown).

### Incidental (in the routines being rewritten)
- Removed dead locals `p_infpTg` and `TJac`, and the now-unused `p_infpT` argument.
- Rewrote the pT relative-convergence test in multiply form to avoid dividing by `pO` (`pO = 0` on the first pass) — addresses @copilot's note on #1653.

### Why this matters on GPU
These solvers run inside `$:GPU_PARALLEL_LOOP` calling `[seq]` device routines, so a warp cannot retire until its slowest cell's loop exits — a single stalled cell stalls 31 idle lanes. The reformulated solvers converge in a bounded, *uniform* handful of iterations, which is the real GPU win beyond "no hang." New device code uses the `GPU_*` macros only.

## Testing

- **Convergence:** instrumented the bubble case — every cell now converges in **≤25 (pTg) / ≤60 (TSat)** iterations instead of stalling; no cell hits the `max_iter` backstop.
- **CPU (gfortran):** all 18 phase-change golden tests pass. The 6 `model 6` goldens (which exercise pTg/TSat) are regenerated — they shift by ~**1e-6 relative** (same equilibrium root, now reached with consistent non-lagged auxiliaries); the 12 `model 5` / pT-only goldens are **unchanged**.
- **GPU (nvfortran/OpenACC, H200):** all 18 phase-change tests pass against the CPU-generated goldens — **CPU and GPU agree to tolerance**. The exact `2D_phasechange_bubble` case that hung at `t_step = 8` on GPU now runs cleanly (reached `t_step > 400`).
- The `max_iter` cap from #1653 is retained as a permanent backstop.

## Checklist

- [x] I added or updated tests for new behavior (regenerated the 6 affected `model 6` golden files)

<details>

- [x] GPU results match CPU results (18/18 phase-change tests, nvfortran/acc vs CPU goldens)
- [x] Tested on NVIDIA GPU (H200)

</details>

## Note on stacking

This branch builds on #1653, so until that merges the diff here shows both commits (the `max_iter` cap + this convergence rewrite). Happy to rebase once #1653 lands, or to fold #1653 into this PR if reviewers prefer the complete fix in one place.


tao13146500073 (approved):

tao13146500073 (commented):
Confirmed on my end — the stock 2D_phasechange_bubble that hung at t_step 8 now runs past 200 steps on NVHPC/OpenACC. Also verified convergence independently: dropping max_iter from 100000 to 100 leaves Time/step unchanged (1.422 s both), so no cell is hitting the backstop. My case is 3-fluid (liquid water / vapor / air) with cyl_coord = T, which I think exercises the inert-fluid path in s_compute_ptg_residual that Copilot flagged on pmin — no issue observed.

PR #1662, opened 2026-07-20
ci(frontier_amd): build base and chemistry simulation variants concurrently (2 h walltime overrun)

sbryngelson:
## Problem

Each `Frontier (AMD) gpu-omp` test job builds **two** simulation binaries — the base one and the chemistry one (the mechanism is compiled in, so chemistry cases need their own binary) — serially inside one 1:59:00 SLURM job. On amdflang each device LTO link takes 40–60 minutes, so the build phase alone nearly exhausts the walltime.

From this morning's master run (job `Oak Ridge | Frontier (AMD) (gpu-omp [2/2])`):

| phase | time |
|---|---|
| simulation, base | 2254.8 s (37.6 min) |
| simulation, chemistry h2o2 | 2588.0 s (43.1 min) |
| everything else + tests | ~23 min |
| **total in-job** | **~104 min of 119 available** |

master passes with ~15 minutes of headroom. Any growth in device code spends that headroom and the job is cancelled at the walltime before the tests finish — which is what happens on #1628, where the same two links take 53 min and ≥63 min.

## Change

The two variants write to disjoint staging directories, so they can be built at the same time. For `frontier_amd` GPU jobs the Build step now submits one SLURM job per variant and waits on both — the same fan-out `Pre-Build (SLURM)` already uses for case-optimization builds, for the same reason ("AMD flang case-opt builds are ~30 min/case, so one serial pre-build job exceeds its walltime").

- `submit-slurm-job.sh`: optional 6th argument selects a build variant; it is exported as `$job_variant` and included in the job slug so the two jobs get distinct log/ID files.
- `common/build.sh`: `base` builds the plain targets, `chem` builds the chemistry variant (`-o Chemistry`), unset keeps today's build-everything behavior — so every other cluster is untouched.
- `test.yml`: `frontier_amd` + `gpu` runs the two submissions concurrently; log printing/archiving globs the per-variant build logs.

Build-phase wall time becomes the longer of the two links (~45–60 min) instead of their sum, and the test job gets its full walltime back.

## Notes

- Verified from both shards' logs on master that this matrix builds exactly two variants (base and `Chemistry: h2o2.yaml`), so the split is complete — no test is left without its binary.
- LTO partition count is **not** a lever: `-flto-partitions=8` → 477 s vs `=64` → 479 s for the same link. The time is dominated by codegen of a few very large kernels and the serial bitcode merge, not by work that splits across partitions.
- No solver code changes; CI-only.


sbryngelson:
Good catch — fixed in b4ad4961. The variant now gets the same `sed 's/[^a-zA-Z0-9]/-/g'` treatment the rest of the slug already uses, so it can't produce out-of-tree or otherwise invalid `*.out`/`*.slurm_job_id` paths:

```
base            -> build-gpu-omp-base.out
chem            -> build-gpu-omp-chem.out
../../etc/passwd -> build-gpu-omp-------etc-passwd.out
'a b/c'         -> build-gpu-omp-a-b-c.out
```

Note the raw value is still what `build.sh` dispatches on, and it rejects anything outside `base|chem|""` — so an unexpected variant fails the job rather than silently building the wrong thing.

PR #1667, opened 2026-07-22
Add hemisphere shell particle cloud packing

BCKim55:
## Description

Adds a hemisphere-shell particle cloud packing option for immersed-boundary particle clouds.

This introduces `particle_cloud(i)%packing_method = 3`, which randomly places spherical/circular IBM particles inside a hemisphere-shell region while enforcing:
- inner and outer shell-radius clearance,
- hemisphere plane clearance,
- bounding-box clearance,
- particle-particle non-overlap using the existing spatial hash approach.

This also adds `shell_inner_radius` and `shell_outer_radius` particle cloud parameters. The local verification example was removed from this PR to avoid introducing a new golden test in the same change.

### Type of change

- New feature

## Testing

- `./mfc.sh format`
- `./mfc.sh validate examples/*/case.py`
- `./mfc.sh validate examples/3D_mibm_particle_cloud_hemi_shell/case.py`
- `./mfc.sh run examples/3D_mibm_particle_cloud_hemi_shell/case.py --clean --no-debug`

Additional local checks:
- Verified generated `ib_state_0.dat` particle positions.
- Confirmed no shell-boundary violations.
- Confirmed no particle overlap violations.
- Confirmed VF 0.2 succeeds with `min_spacing=0.02`.
- Confirmed VF 0.3 succeeds when `min_spacing=0.0`.

## Checklist

<details>

</details>


wilfonba:
I don't think this is implemented quite right. The example you added highlights this fact. It defines a hemispherical particle cloud using a cubic domain. I think instead you should add a `particle_cloud(i)%geometry` variable, and then treat the particle cloud more like the fluid patches are treated. Also, a hemispherical shell is not a packing method in the same way that rejection sampling and lattice placement are packing methods, so using the existing `acking_method` variable for this isn't the right thing to do. You can place particles in a hemispherical shell using either rejection sampling, lattice placement, or some other packing method that's not implemented yet.

Edit: Your `s_particle_cloud_random_hemi_shell` routine appears to do rejection sampling, which adds to the confusion of what `packing_method` means.

BCKim55:
Thanks Ben, that makes sense. I agree that the hemisphere shell should be treated as a particle-cloud geometry rather than a packing method.

I’ll refactor this so that `particle_cloud(i)%packing_method` remains responsible for the placement algorithm, e.g. rejection sampling or lattice, and add a separate `particle_cloud(i)%geometry` option for the cloud shape. Then the current hemisphere-shell implementation will become the rejection-sampling path for `geometry = hemisphere shell`, while unsupported combinations such as hemisphere-shell lattice packing can error out explicitly for now.

I’ll also remove or adjust the example so it does not imply that a cubic domain defines the hemispherical cloud.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +183,149 @@ contains
...
+
+        xmin = particle_cloud(cloud_idx)%x_centroid - 0.5_wp*particle_cloud(cloud_idx)%length_x
+        xmax = particle_cloud(cloud_idx)%x_centroid + 0.5_wp*particle_cloud(cloud_idx)%length_x
+        ymin = particle_cloud(cloud_idx)%y_centroid - 0.5_wp*particle_cloud(cloud_idx)%length_y
+        ymax = particle_cloud(cloud_idx)%y_centroid + 0.5_wp*particle_cloud(cloud_idx)%length_y
+        zmin = particle_cloud(cloud_idx)%z_centroid - 0.5_wp*particle_cloud(cloud_idx)%length_z
+        zmax = particle_cloud(cloud_idx)%z_centroid + 0.5_wp*particle_cloud(cloud_idx)%length_z
+
+        r_inner = particle_cloud(cloud_idx)%shell_inner_radius + particle_cloud(cloud_idx)%radius
+        r_outer = particle_cloud(cloud_idx)%shell_outer_radius - particle_cloud(cloud_idx)%radius
```
danieljvickers, line 171: Forgive me if this seems dense, but do you not already have the size on this? You have zmin, zmax, xmin, ..., , which gives you some hypothetical upper size of the shell, but then you also have radii. It seems like redundant information. You either use the radii and ignore the [xyz][min/max] or use the shape size and ignore the outer radii. Seems like wasted lines of code otherwise.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +183,149 @@ contains
...
+        hash_size = max(16, 4*particle_cloud(cloud_idx)%num_particles)
+        allocate (hash_head(hash_size))
+        allocate (chain_next(particle_cloud(cloud_idx)%num_particles))
+        hash_head = -1
+        chain_next = -1
+
+        do while (n_placed < particle_cloud(cloud_idx)%num_particles .and. n_attempts < max_attempts)
+            n_attempts = n_attempts + 1
+
+            if (p == 0) then
```
danieljvickers: `num_dims < 3` peferred in general. This likely does not affect things, but num_dims is a case-optimization parameter, and therefore the compiler can optimize this away if you make it a check on `num_dims`.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +183,149 @@ contains
...
+                ry = particle_cloud(cloud_idx)%y_centroid + r_shell*sin(theta)
+                rz = particle_cloud(cloud_idx)%z_centroid
+            else
+                phi = 2._wp*pi*f_xorshift(seed)
+                zdir = f_xorshift(seed)
+                rho = sqrt(max(0._wp, 1._wp - zdir**2._wp))
+                xdir = rho*cos(phi)
+                ydir = rho*sin(phi)
+                u = f_xorshift(seed)
+                r_shell = ((r_outer**3._wp - r_inner**3._wp)*u + r_inner**3._wp)**(1._wp/3._wp)
```
danieljvickers: This seems like a lot fo compute for your QOI. You do not use the [xyz]dir parameter at all after computing it. Again, either use it or do not computed.

danieljvickers: That said, you using a probability distribution function to weight your random seed is correct here. Just you are computing redundant quantities here that seem pointless.

src/simulation/m_particle_cloud.fpp
BCKim55: Good point. I kept the box bounds thinking about the spatial hash, but the hash only depends on the candidate position and `min_dist`, not on `length_x/y/z`. For the hemisphere-shell geometry, the inner/outer radii define the placement region, so I removed the redundant `[xyz]min/max` bounds logic from this path.

src/simulation/m_particle_cloud.fpp
BCKim55: Updated this to use `num_dims < 3` instead of checking `p == 0`.

src/simulation/m_particle_cloud.fpp
BCKim55: 1. Thanks for catching that! I completely missed that [xyz]dir wasn't being used later on. I've removed the redundant calculation to save compute.

2. Thanks for confirming the PDF logic. I've cleaned up all the redundant quantities you mentioned to optimize the compute.

BCKim55 (commented):
Thanks for the feedback. I updated the hemisphere-shell path to remove the redundant box-bounds logic and rely on the shell radii for the placement region. I also changed the dimensionality checks to use `num_dims < 3` and removed the extra direction variables in the 3D sampling path.

toolchain/mfc/case_validator.py
```diff
@@ -637,6 +642,16 @@ def check_ibm(self):
                 packing_method is not None and packing_method not in [1, 2],
                 f"particle_cloud({i})%packing_method must be 1 (rejection sampling) or 2 (lattice)",
             )
+            shell_outer_radius = self.get(f"particle_cloud({i})%shell_outer_radius", None)
+            shell_inner_radius = self.get(f"particle_cloud({i})%shell_inner_radius", None)
+            self.prohibit(
+                geometry == 2 and (shell_outer_radius is None or shell_inner_radius is None or shell_outer_radius <= shell_inner_radius),
```
sbryngelson: **Feasibility validation is incomplete.** This accepts (for example) `shell_inner_radius=0`, `shell_outer_radius=0.1`, and `radius=0.1`, even though no centre can satisfy the required radial clearances. The run then reaches the startup sampler and aborts at `r_outer <= r_inner`. Please validate `shell_inner_radius >= 0` and `shell_outer_radius > shell_inner_radius + 2*radius` here, and add the equivalent Fortran input check so invalid cases fail before initialization.

sbryngelson (changes requested):
**Request changes**

1. **Missing regression coverage.** This PR adds the hemisphere-shell execution path but removes the local example and adds no test fixture. The existing particle-cloud case is box-only, so CI never executes `geometry = 2`. Please add a small deterministic 2D or 3D case that checks generated particle positions for shell/plane clearance and non-overlap.

2. **Documentation was not updated.** `docs/documentation/case.md` does not document `geometry`, `shell_inner_radius`, or `shell_outer_radius`, and its `length_[x,y,z]` description is inaccurate for `geometry = 2`, where those extents are ignored. Please update the Particle Clouds section accordingly.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +183,135 @@ contains
...
+                if (rz < particle_cloud(cloud_idx)%z_centroid + particle_cloud(cloud_idx)%radius) cycle
+            end if
+
+            bx = int(floor(rx/min_dist))
+            by = int(floor(ry/min_dist))
+            bz = 0
+            if (num_dims == 3) bz = int(floor(rz/min_dist))
+
+            overlaps = .false.
+            outer: do dx_b = -1, 1
```
danieljvickers: This loop looks like you copied the code from the cuboid rejection-packing subroutine and dropped it here. If we are going to do that, seems like it may be best to create a separate routine to check for overlapping and to call it here and in the cuboid routine instead. It would probably delete 30 lines from this section of the code.

toolchain/mfc/case_validator.py
BCKim55: Thanks, I addressed these points.

For coverage, I added deterministic particle-cloud regression cases that exercise the `geometry = 2` hemisphere-shell path directly:
- 2D -> 1 Fluid(s) -> IBM -> Particle Cloud -> Hemisphere Shell
- 2D -> 2 Fluid(s) -> IBM -> Particle Cloud -> Hemisphere Shell

I generated the corresponding golden files and also added Python unit coverage for the shell constraints. The tests check shell/plane clearance, non-overlap, and validator rejection of infeasible shell radii.

For documentation, I updated `docs/documentation/case.md` to describe `geometry`, `shell_inner_radius`, `shell_outer_radius`, and to clarify that `length_x/y/z` apply to box clouds and are ignored for hemisphere-shell clouds.

For validation, both the Python validator and Fortran checker now require:
- `shell_inner_radius >= 0`
- `shell_outer_radius > shell_inner_radius + 2*radius`

src/simulation/m_particle_cloud.fpp
BCKim55: Thanks, I refactored the duplicated spatial-hash overlap check into a shared helper.

Both the box and hemisphere-shell rejection-packing routines now call the same overlap-check path through `f_cloud_particle_overlaps(...)`, with bin coordinate generation handled by `s_get_cloud_bin(...)`.

I also kept the earlier cleanup from your other comments: the hemisphere-shell path now uses the radii directly, uses `num_dims < 3`, and avoids the redundant direction variables.

BCKim55:
Hi Spencer, I addressed the requested validation, documentation, and regression coverage changes. CI is now passing, and the PR is ready for another look when you have time.

toolchain/mfc/test_particle_cloud.py
```diff
@@ -0,0 +1,170 @@
...
+        "particle_cloud(1)%num_particles": 8,
+        "particle_cloud(1)%radius": 0.04,
+        "particle_cloud(1)%mass": 1.0,
+        "particle_cloud(1)%min_spacing": 0.01,
+        "particle_cloud(1)%shell_inner_radius": 0.1,
+        "particle_cloud(1)%shell_outer_radius": 0.3,
+    }
+
+
+def test_hemi_shell_random_packing_respects_clearance_and_overlap():
```
sbryngelson: **This test does not exercise the code the PR adds.** `_pack_hemi_shell_3d` is a Python reimplementation of `s_particle_cloud_random_hemi_shell`, and the assertions below check that the *Python* implementation satisfies the shell and overlap constraints. Nothing here reads from `m_particle_cloud.fpp`, so no regression in the Fortran sampler can make this test fail.

What I asked for was verification of the generated particle positions. Please parse `ib_state_0.dat` from an actual run and assert the shell clearance, plane clearance, and pairwise non-overlap on those coordinates. A mirrored implementation cannot serve that purpose no matter how faithful it is.

toolchain/mfc/test_particle_cloud.py
```diff
@@ -0,0 +1,170 @@
...
+                neighbor = (cell[0] + dx, cell[1] + dy, cell[2] + dz)
+                for particle_idx in cells.get(neighbor, []):
+                    distance_squared = sum((candidate[axis] - placed[particle_idx][axis]) ** 2 for axis in range(3))
+                    if distance_squared < min_distance_squared:
+                        return True
+
+    return False
+
+
+def _pack_hemi_shell_3d(count, radius, min_spacing, inner_radius, outer_radius, seed_value):
```
sbryngelson: Separately from the point above: this model is not actually equivalent to the Fortran. `_is_overlapping` keys `cells` on exact bin tuples, whereas `f_cloud_particle_overlaps` folds bins through `f_bin_hash` into `hash_size` buckets. Collisions cause the Fortran to reject candidates this model accepts, so the two will diverge on particle placement for the same seed.

So even taken as documentation of the algorithm, it is misleading. Another reason to test the real output instead.

toolchain/mfc/test/cases.py
```diff
@@ -1008,6 +1008,34 @@ def alter_ib(dimInfo, six_eqn_model=False, viscous=False):
                     },
                 )
             )
+            cases.append(
+                define_case_d(
+                    stack,
+                    "IBM -> Particle Cloud -> Hemisphere Shell",
```
sbryngelson, line 1014: Two coverage gaps here.

**1. 3D is not covered.** This block sits inside `if len(dimInfo[0]) == 2 and not viscous:`, so CI only ever runs the 2D branch — which is a half-annulus, not a hemisphere. The 3D path (`phi`/`zdir`/`rho` with the cube-root radial CDF) is the one this feature is named for and it never executes. Please add a 3D golden.

**2. The box path this PR rewrote has no coverage at all.** `grep particle_cloud toolchain/mfc/test/cases.py` returns only this new case — it is the only particle-cloud golden in the suite. Meanwhile `s_particle_cloud_random_box` was rewritten here (`p == 0` -> `num_dims < 3`, the inlined overlap loop replaced by `f_cloud_particle_overlaps`, bin coordinates now computed after acceptance rather than before). That refactor is unverified. Please add a box golden as well, or demonstrate bit-identical output before and after the change on an existing example.

Note also that the goldens hold `cons.*` fields rather than `ib_state`. Placement perturbs the flow, so this is a real regression lock, but it does not directly assert the geometric constraints.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +147,100 @@ contains
...
+        r_inner = particle_cloud(cloud_idx)%shell_inner_radius + particle_cloud(cloud_idx)%radius
+        r_outer = particle_cloud(cloud_idx)%shell_outer_radius - particle_cloud(cloud_idx)%radius
+        min_dist = 2._wp*particle_cloud(cloud_idx)%radius + particle_cloud(cloud_idx)%min_spacing
+
+        if (r_inner < 0._wp .or. r_outer <= r_inner) then
+            call s_mpi_abort("Error :: Invalid hemisphere-shell radii for particle cloud packing")
+        end if
+
+        if (num_dims < 3) then
+            geom = 2
```
sbryngelson, line 179: `geometry` is overloaded in a way that will bite users. Here `particle_cloud(i)%geometry == 2` means hemisphere shell, while two lines down `geom = 2` means circle in the `patch_ib` vocabulary (and `geom = 8` means sphere). The same subroutine uses two different meanings of `geometry = 2` within six lines.

Please rename the cloud field to something like `region_shape` or `cloud_geometry`, or reuse the existing patch geometry codes. As written, a user who reads `particle_cloud(1)%geometry = 2` will reasonably expect a circle.

src/simulation/m_particle_cloud.fpp
```diff
@@ -173,6 +147,100 @@ contains
...
+                rho = sqrt(max(0._wp, 1._wp - zdir**2._wp))
+                u = f_xorshift(seed)
+                r_shell = ((r_outer**3._wp - r_inner**3._wp)*u + r_inner**3._wp)**(1._wp/3._wp)
+                rx = particle_cloud(cloud_idx)%x_centroid + r_shell*rho*cos(phi)
+                ry = particle_cloud(cloud_idx)%y_centroid + r_shell*rho*sin(phi)
+                rz = particle_cloud(cloud_idx)%z_centroid + r_shell*zdir
+            end if
+
+            if (num_dims < 3) then
+                if (ry < particle_cloud(cloud_idx)%y_centroid + particle_cloud(cloud_idx)%radius) cycle
```
sbryngelson, line 220: Two things about the region definition.

**Orientation is hard-coded and unconfigurable.** The flat face is at `y_centroid` in 2D and `z_centroid` in 3D, always opening toward +y / +z. That is a reasonable initial restriction, but it needs to be stated in `case.md` — right now nothing tells a user which way the hemisphere faces.

**Nothing bounds the cloud to the domain.** Dropping the `[xyz]min/max` logic was the right call per @danieljvickers, but nothing replaced it: `x_centroid +/- shell_outer_radius` can extend past `x_domain`, and `s_add_cloud_particle` performs no bounds check, so particles land off-grid silently. The validator already has `x[y,z]_centroid`, `shell_outer_radius`, and the domain extents, so this is a cheap check to add there.

src/simulation/m_particle_cloud.fpp
```diff
@@ -313,6 +381,69 @@ contains
...
+            do dy_b = -1, 1
+                do dz_b = dz_lo, dz_hi
+                    nbx = bx + dx_b
+                    nby = by + dy_b
+                    nbz = bz + dz_b
+                    slot = f_bin_hash(nbx, nby, nbz, hash_size)
+                    j = hash_head(slot)
+                    do while (j > 0)
+                        if (num_dims < 3) then
+                            dist_sq = (px - placed(1, j))**2._wp + (py - placed(2, j))**2._wp
```
sbryngelson: Minor: the extraction turned integer exponents into real ones. The original was `(rx - placed(1, j))**2`; this is now `**2._wp`, and `min_dist_sq = min_dist**2._wp` above is the same. gfortran folds `x**2.0` to `x*x` (I checked at both -O0 and -O2), but that is a compiler courtesy rather than a guarantee, and this is the innermost loop of the rejection sampler. Please use `**2`.

Also minor, in the same area: `s_get_cloud_bin` runs twice per accepted particle — once inside `f_cloud_particle_overlaps` and again at the insertion site on line 227 — and it tests `num_dims == 3` while everything else in the module tests `num_dims < 3`.

docs/documentation/case.md
```diff
@@ -425,18 +425,25 @@ A particle cloud is a compact specification of a bed of identical circular (2D)
...
+| `geometry`        | Integer | Shape of the cloud region. |
+| `shell_inner_radius` | Real | Inner radius for hemisphere-shell clouds (`geometry = 2`). |
+| `shell_outer_radius` | Real | Outer radius for hemisphere-shell clouds (`geometry = 2`). |
 | `moving_ibm`      | Integer | Motion flag applied to every particle (see `patch_ib(j)%%moving_ibm`). |
 | `seed`            | Integer | Random seed for reproducible placement (used by `packing_method = 1`). |
 | `packing_method`  | Integer | Algorithm used to place the particles. |
 
+- `geometry` selects the cloud region:
+  - `1` (box) uses `x[y,z]_centroid` and `length_x[y,z]` to define the region.
+  - `2` (hemisphere shell) uses `x[y,z]_centroid`, `shell_inner_radius`, and `shell_outer_radius` to define the region. Particle centres are sampled between `shell_inner_radius + radius` and `shell_outer_radius - radius`, and the flat hemisphere plane is kept clear by one particle radius.
```
sbryngelson: This is a clear improvement, but two facts a user needs are still missing:

- Which way the hemisphere opens. The flat face lies at `y_centroid` in 2D and `z_centroid` in 3D, and the filled half is +y / +z. This is not configurable and is not stated anywhere.
- That the 2D region is a half-annulus, not a hemisphere. Calling both cases "hemisphere shell" is going to confuse people setting up 2D cases.

sbryngelson (changes requested):
Thanks for the revisions. Documentation and validation are genuinely addressed:

- `case.md` now documents `geometry`, `shell_inner_radius`, `shell_outer_radius`, and qualifies `length_x[y,z]` as box-only.
- The Python validator and `m_checker.fpp` now enforce the same condition, and `shell_outer_radius > shell_inner_radius + 2*radius` is exactly the sampler's feasibility requirement (`r_outer > r_inner` after the +/- radius insets), so the `s_mpi_abort` path is no longer reachable from valid input.
- `geometry` and both new reals are broadcast in `m_mpi_proxy.fpp`.
- The sampling math is right: the 2D `sqrt` CDF and the 3D cube-root CDF with uniform `cos(polar)` both give uniform density in the shell.

The test coverage point is not yet addressed, though, and that is the blocking item. Details inline.

toolchain/mfc/test_particle_cloud.py
BCKim55: Addressed. Instead of validating a Python mirror implementation, I now run the actual MFC particle-cloud cases and validate the Fortran-generated ib_state_0.dat output. The test checks that generated particles satisfy the box/hemi-shell bounds and the minimum non-overlap distance.

toolchain/mfc/test_particle_cloud.py
BCKim55: Agreed. The Python mirror was not equivalent to the Fortran spatial hash implementation because it keyed cells directly rather than folding them through f_bin_hash, so it could accept candidates that the Fortran code rejects after hash collisions.

I removed the mirror implementation entirely. The tests now run the actual MFC/Fortran particle-cloud generation and validate the generated restart_data/ib_state_0.dat output directly for geometry bounds, hemi-shell clearance, and particle-particle non-overlap.

src/simulation/m_particle_cloud.fpp
BCKim55: Addressed. I renamed `particle_cloud%geometry` to `particle_cloud%cloud_geometry` throughout the Fortran code, parameter registry, validator, documentation, and tests to avoid confusion with `patch_ib%geometry`.

src/simulation/m_particle_cloud.fpp
BCKim55: Addressed. I expanded the documentation to define the hemi-shell orientation and domain assumptions explicitly. In 2D, this is treated as a half-annulus with the flat face at y_centroid and opening in +y. In 3D, this is a hemisphere shell with the flat face at z_centroid and opening in +z.

I also added validator checks to require the full shell extent to remain inside the computational domain.

docs/documentation/case.md
BCKim55: This overlaps with the geometry/orientation documentation issue discussed above, so I addressed it as part of that same update.

The documentation now explicitly distinguishes the 2D case as a half-annulus and the 3D case as a hemisphere shell. It also states the fixed orientation: the 2D half-annulus opens in +y from y_centroid, and the 3D hemisphere shell opens in +z from z_centroid.

src/simulation/m_particle_cloud.fpp
BCKim55: Follow-up: after the later rename, this is now the `cloud_geometry = 2` path rather than `particle_cloud%geometry = 2`.

src/simulation/m_particle_cloud.fpp
BCKim55: Addressed. I changed the distance calculations back to integer exponents, e.g. min_dist**2 and coordinate differences **2.

I also refactored the overlap check so it returns the candidate hash-bin indices along with the overlap flag. That way, accepted particles reuse the same bx/by/bz values at insertion instead of calling s_get_cloud_bin a second time. I also updated the dimensionality check to use num_dims < 3 for the 2D path.

toolchain/mfc/test/cases.py
BCKim55: Thanks, this was a good catch.

1. I added 3D particle-cloud regression coverage for the actual hemisphere-shell path. The test matrix now includes 3D hemisphere-shell particle-cloud cases, so the phi/zdir/rho sampling and cube-root radial CDF branch is exercised in CI.

2. I also added box particle-cloud regression coverage for the refactored random-box path. The test matrix now includes box particle-cloud cases, so the shared overlap-check refactor and num_dims-based 2D/3D logic are covered as well.

I also enabled ib_state_wrt for these particle-cloud cases and added a verifier that reads restart_data/ib_state_0.dat directly. The regular goldens still lock the resulting flow-field regression, while the new verifier directly checks the particle geometry constraints: box bounds, shell radial clearance, shell flat-plane clearance, and minimum particle-particle spacing.

BCKim55:
Hi Spencer,

I’ve pushed the latest fixes and the PR is ready for re-review.

The blocking regression coverage item is now addressed for particle clouds. The new checks validate generated `ib_state_0.dat` particle count, shell/plane clearance, box bounds, and non-overlap. I also fixed the single-precision test reader issue that caused the Ubuntu GNU no-debug CI failure.

All GitHub checks are now passing: 60 success, 5 skipped, 0 failures.

Thanks again!

toolchain/mfc/test/cases.py
```diff
@@ -1044,6 +1044,98 @@ def alter_ib(dimInfo, six_eqn_model=False, viscous=False):
...
+                        "m": 29,
+                        "n": 29,
+                        "p": 29,
+                        "particle_cloud(1)%cloud_geometry": 2,
+                        "particle_cloud(1)%packing_method": 1,
+                        "particle_cloud(1)%x_centroid": 0.5,
+                        "particle_cloud(1)%y_centroid": 0.5,
+                        "particle_cloud(1)%z_centroid": 0.1,
+                        "particle_cloud(1)%num_particles": 4,
+                        "particle_cloud(1)%radius": 0.02,
```
sbryngelson: **The 3D particles are sub-cell, so this golden may not be observing the feature.**

This case sets `m=n=p=29` on `[0,1]^3`, so `dx = 1/30 = 0.0333` and `radius = 0.02 = 0.6*dx`. A sphere smaller than one cell contains at most one cell centre, sometimes zero depending on where the seed lands it relative to a cell centre. Two problems follow:

- If a particle marks no cells, the golden is blind to the IB it was added to cover. The 3D path executes but nothing observes it.
- Where a particle does mark a cell, the marked set is a discontinuous function of position, so it can flip across the compiler and precision matrix this golden runs on.

For contrast, the two 2D goldens (`tests/E085CC5A` vs `tests/5A22B45F`) differ by up to 0.37 at `t_step 50`, so the 2D cases do carry real signal at `radius = 1*dx`. The 3D case is the one that needs attention.

Please raise `radius` (or refine the grid) so particles span 3-4 cells, adjusting `shell_inner_radius`/`shell_outer_radius` to keep the packing feasible. A quick way to confirm the fix: check that the 3D golden actually differs from the same case with `num_particle_clouds = 0`.

toolchain/mfc/test/test.py
```diff
@@ -48,6 +50,93 @@ class TestTimeoutError(MFCException):
...
+
+    for cloud_idx in range(1, num_particle_clouds + 1):
+        geometry = case.params.get(f"particle_cloud({cloud_idx})%cloud_geometry", 1)
+        count = case.params.get(f"particle_cloud({cloud_idx})%num_particles", 0) or 0
+        radius = case.params[f"particle_cloud({cloud_idx})%radius"]
+        records_end = start + count
+        if records_end > len(records):
+            raise MFCException(f"particle_cloud({cloud_idx}) expected {count} IB state records, found {len(records) - start}")
+
+        if geometry == 1:
```
sbryngelson: **This box assertion is only valid for `packing_method = 1`, and the lattice packer looks like it has a real bug.**

This branch asserts every particle centre lies within `centroid +/- length/2`. That invariant holds for rejection sampling but not for lattice packing. `s_particle_cloud_lattice` has no upper bound on its outer loops:

- 2D: `do while (n_placed < n_target)` increments `row` with `py = ymin + real(row, wp)*row_dy` and there is no `py <= ymax` guard.
- 3D: `kz` increments with no `zmax` guard.

`spacing` is derived so that the area/volume per particle yields exactly `n_target` sites in the region, but edge effects mean the sites that actually fit are generally fewer than `n_target`. So the loop runs extra rows or layers past `ymax`/`zmax`. That is the common case, not a corner case.

Two separate things to do:

1. In this PR: gate the box-bounds assertion on `packing_method == 1`. As written, adding `ib_state_wrt: "T"` to any lattice box-cloud case fails this assertion even though the harness is fine.
2. Separately: `s_particle_cloud_lattice` silently places particles outside the requested region, and potentially outside the domain. That is pre-existing and not this PR's to fix, but it is worth its own issue. This assertion is what exposed it.

src/simulation/m_checker.fpp
```diff
@@ -265,19 +265,30 @@ contains
...
             @:PROHIBIT(particle_cloud(i)%packing_method == dflt_int, &
                        & "particle_cloud("//trim(idxStr) &
                        & //")%packing_method must be specified (1 = rejection sampling, 2 = lattice)")
             @:PROHIBIT(particle_cloud(i)%packing_method /= 1 .and. particle_cloud(i)%packing_method /= 2, &
                        & "particle_cloud("//trim(idxStr) //")%packing_method must be 1 (rejection sampling) or 2 (lattice)")
+            @:PROHIBIT(particle_cloud(i)%cloud_geometry == 2 .and. particle_cloud(i)%shell_inner_radius < 0._wp, &
+                       & "particle_cloud("//trim(idxStr) //") hemisphere shell requires shell_inner_radius >= 0")
+            @:PROHIBIT(particle_cloud(i)%cloud_geometry == 2 &
+                       & .and. particle_cloud(i)%shell_outer_radius <= particle_cloud(i)%shell_inner_radius &
+                       & + 2._wp*particle_cloud(i)%radius, &
```
sbryngelson: **This guard is vacuous when `radius` is unset.**

Nothing in `s_check_inputs_particle_clouds` requires `radius` to be specified, so an unset `radius` is `dflt_real = -1.e6_wp`. The right-hand side becomes roughly `shell_inner_radius - 2e6`, and any positive `shell_outer_radius` passes. The run then dies inside `s_particle_cloud_random_hemi_shell` on the generic `"Invalid hemisphere-shell radii for particle cloud packing"` instead of naming the parameter the user actually forgot.

The Python validator catches this (`radius is None`), so it only bites the direct-binary path, which is the same asymmetry as the domain-containment comment below.

The broader version of this: neither `m_checker` nor `case_validator` requires `radius`, `mass`, or `num_particles` to be set for any cloud, box or shell. An unset `radius` also makes `min_dist` negative in the box path, which corrupts both `int(floor(px/min_dist))` and the `dist_sq < min_dist_sq` test. Please add unconditional positivity `PROHIBIT`s for all three, independent of `cloud_geometry`.

src/simulation/m_checker.fpp
```diff
@@ -265,19 +265,30 @@ contains
...
+                       & "particle_cloud("//trim(idxStr) &
+                       & //") hemisphere shell requires shell_outer_radius > shell_inner_radius + 2*radius")
+            @:PROHIBIT(particle_cloud(i)%cloud_geometry == 2 .and. particle_cloud(i)%packing_method == 2, &
+                       & "particle_cloud("//trim(idxStr) //") hemisphere-shell lattice packing is not implemented")
```
sbryngelson: **Domain containment is enforced only in Python.**

`case_validator.py` gained the x/y/z extent checks, but nothing equivalent landed here. MFC supports running the binaries against a hand-written input file, and that path skips the toolchain entirely, so a shell that extends past the domain is accepted with no diagnostic.

This is the same gap I raised for the radii feasibility checks, which you correctly mirrored into `m_checker`. Please mirror the containment checks too.

src/simulation/m_checker.fpp
```diff
@@ -265,19 +265,30 @@ contains
...
+    !> Checks that each active particle cloud has a valid geometry and packing_method specified
     impure subroutine s_check_inputs_particle_clouds
 
         integer          :: i
         character(len=5) :: idxStr
 
         do i = 1, num_particle_clouds
             call s_int_to_str(i, idxStr)
+            @:PROHIBIT(particle_cloud(i)%cloud_geometry /= 1 .and. particle_cloud(i)%cloud_geometry /= 2, &
+                       & "particle_cloud("//trim(idxStr) //")%cloud_geometry must be 1 (box) or 2 (hemisphere shell)")
```
sbryngelson: **`cloud_geometry = 2` is unguarded in 1D.**

With `n == 0`, `cloud_geometry = 2` falls into the `num_dims < 3` branch of the sampler and generates a half-annulus in a 1D run. Nothing rejects it here, and the Python validator's y-extent check is explicitly skipped when `n == 0`, so the one check that would have caught it is skipped precisely where it is needed.

Please add `@:PROHIBIT(particle_cloud(i)%cloud_geometry == 2 .and. n == 0, ...)`.

src/simulation/m_particle_cloud.fpp
```diff
@@ -207,6 +188,120 @@ contains
 
     end subroutine s_particle_cloud_random_box
 
+    !> Generates a random distribution of particles in a hemisphere shell with a minimum spacing. Like s_particle_cloud_random_box,
+    !! rejection sampling needs every placed particle tracked regardless of which rank's neighborhood it falls in, so cloud_ibs is
+    !! allocated here to the cloud's full requested particle count and pared down afterwards via s_reduce_particle_cloud_ibs.
+    subroutine s_particle_cloud_random_hemi_shell(cloud_idx, glbl_idx, cloud_ibs, num_cloud_ibs)
```
sbryngelson: **This is still a near-verbatim copy of `s_particle_cloud_random_box`.**

@danieljvickers' point was addressed for the overlap check, and extracting `s_check_cloud_particle_overlap` was the right move. But roughly 60 of this routine's 110 lines are still exact duplicates of the box routine: the `allocate` plus its error string, the `geom` selection, `max_attempts`, the seed fallback, the `placed`/`hash_head`/`chain_next` sizing and the comment above it, the accept-and-insert block, the failure abort, the `deallocate`, and the `s_reduce_particle_cloud_ibs` call.

The genuinely new code is about 18 lines of candidate generation plus the plane-clearance `cycle`.

Collapsing both routines into one `s_particle_cloud_rejection_pack`, with a `select case (particle_cloud(cloud_idx)%cloud_geometry)` inside the loop that yields `(rx, ry, rz)` and a reject flag, removes that duplication. It also lets `s_generate_particle_clouds` go back to a flat dispatch instead of the nested `select case` with two identical `case default` aborts.

I would push on this one: the `cloud_geometry` enum invites a third geometry, and a third geometry triples the copy. It is much cheaper to do now than after.

src/simulation/m_particle_cloud.fpp
```diff
@@ -207,6 +188,120 @@ contains
...
+            call s_mpi_abort("Error :: Ran out of CPU memory trying to allocate particle cloud IB array. " &
+                             & // "Current system resources cannot perform rejection packing with the specified number of particles.")
+        end if
+        ib_idx = 0
+
+        r_inner = particle_cloud(cloud_idx)%shell_inner_radius + particle_cloud(cloud_idx)%radius
+        r_outer = particle_cloud(cloud_idx)%shell_outer_radius - particle_cloud(cloud_idx)%radius
+        min_dist = 2._wp*particle_cloud(cloud_idx)%radius + particle_cloud(cloud_idx)%min_spacing
+
+        if (r_inner < 0._wp .or. r_outer <= r_inner) then
```
sbryngelson: **Dead code, and weaker than the check it duplicates.**

This is unreachable. `s_check_input_file` runs at `m_start_up.fpp:992`, and `s_generate_particle_clouds` sits inside `s_initialize_modules` at `m_start_up.fpp:907`, which runs after. Any input that reaches here has already passed the `m_checker` `PROHIBIT`s.

It is also strictly weaker than the check it duplicates: this admits `r_outer = r_inner + eps`, which `m_checker` rejects. Two guards on the same condition that disagree is worse than one.

Please delete it, or convert it to an assertion that documents the invariant rather than a live branch.

src/simulation/m_particle_cloud.fpp
```diff
@@ -53,13 +53,27 @@ contains
...
+                    call s_particle_cloud_lattice(cloud_idx, glbl_idx, cloud_ibs, num_cloud_ibs)
+                case default
+                    call s_mpi_abort("Particle cloud packing method is not a known packing method of MFC. Exiting.")
+                end select
+            case (2)  ! hemisphere-shell geometry
+                select case (particle_cloud(cloud_idx)%packing_method)
+                case (1)  ! random (rejection) packing method
+                    call s_particle_cloud_random_hemi_shell(cloud_idx, glbl_idx, cloud_ibs, num_cloud_ibs)
+                case (2)
+                    call s_mpi_abort("Error :: Lattice packing is not implemented for hemisphere-shell particle clouds")
```
sbryngelson: The "hemi-shell plus lattice is not implemented" rule is now encoded in three places: `case_validator.py`, `m_checker.fpp`, and here. The first two already make this branch unreachable.

If you take the single-rejection-packer suggestion above, this whole nested `select case` collapses back to a flat dispatch and this arm goes away along with the two duplicated `case default` aborts. If you keep the nested form, drop this arm and let the `m_checker` `PROHIBIT` be the single source of truth.

src/simulation/m_particle_cloud.fpp
```diff
@@ -207,6 +188,120 @@ contains
...
+                r_shell = sqrt((r_outer**2 - r_inner**2)*u + r_inner**2)
+                rx = particle_cloud(cloud_idx)%x_centroid + r_shell*cos(theta)
+                ry = particle_cloud(cloud_idx)%y_centroid + r_shell*sin(theta)
+                rz = particle_cloud(cloud_idx)%z_centroid
+            else
+                phi = 2._wp*pi*f_xorshift(seed)
+                zdir = f_xorshift(seed)
+                rho = sqrt(max(0._wp, 1._wp - zdir**2))
+                u = f_xorshift(seed)
+                r_shell = ((r_outer**3 - r_inner**3)*u + r_inner**3)**(1._wp/3._wp)
```
sbryngelson, line 231: Minor, robustness rather than a bug you will hit.

`f_xorshift` returns `abs(real(seed, wp))/real(huge(seed), wp)`. When `seed` reaches `-2147483648`, `abs()` overflows and gfortran returns the value unchanged, so `rval` is `-1.0`. In the box path a negative draw just places a particle outside the box. Here a negative `u` makes `(r_outer**3 - r_inner**3)*u + r_inner**3` negative whenever the shell is thick enough, so `r_shell` is NaN, and a NaN centre passes `s_check_cloud_particle_overlap` because every floating-point comparison against NaN is false. So it gets placed.

Probability is about 2^-32 per draw, and the `abs()` fold is pre-existing, but the shell path converts a benign misplacement into a silently-accepted NaN. One-line fix in `f_xorshift`:

```fortran
rval = real(iand(seed, huge(seed)), wp)/real(huge(seed), wp)
```

The 2D `sqrt` on line 254 has the same exposure.

Separately, and for the record: the sampling math itself is correct. 2D uses `theta` uniform on `[0, pi]` with the sqrt radial CDF, 3D uses uniform `phi`, uniform `cos(polar)` on `[0, 1]`, and the cube-root radial CDF. Both give uniform density in the shell, and the plane-clearance `cycle` is a hard geometric cut applied after sampling, so it preserves uniformity on the remaining region.

toolchain/mfc/test/test.py
```diff
@@ -48,6 +50,93 @@ class TestTimeoutError(MFCException):
...
+            if dist < min_dist - 1.0e-12:
+                raise MFCException(f"particle_cloud({cloud_idx}) particles overlap in ib_state_0.dat")
+
+
+def _assert_particle_cloud_ib_state(case: TestCase):
+    num_particle_clouds = case.params.get("num_particle_clouds", 0) or 0
+    if num_particle_clouds <= 0 or case.params.get("ib_state_wrt", "F") != "T":
+        return
+
+    ib_state_precision = 1 if ARG("single") or ARG("mixed") else case.params.get("precision", 2)
```
sbryngelson: **`case.params.get("precision", 2)` is the wrong knob.**

In MFC, `precision` selects the database output float format (see `case.md`). It has nothing to do with the working precision. `ib_state_*.dat` is written from `real(wp) :: ib_buf(20)` in `s_write_serial_ib_state` / `s_write_parallel_ib_state`, so its field width depends only on how the binary was built, that is on `--single` / `--mixed`.

Harmless today because no particle-cloud case sets `precision` and the `ARG` branch dominates. But a particle-cloud case with `precision: 1` in a double build would make this read 8-byte doubles as 4-byte floats. Worse, 160-byte records parsed as 80 bytes still satisfy `len(data) % record_size == 0`, so it passes the sanity check and yields garbage coordinates, surfacing as a bogus "particles overlap" or "outside its cloud bounds" failure.

Key off the build flags alone:

```python
field_size = 4 if (ARG("single") or ARG("mixed")) else 8
```

toolchain/mfc/test/test.py
```diff
@@ -48,6 +50,93 @@ class TestTimeoutError(MFCException):
...
+    radius = case.params[f"particle_cloud({cloud_idx})%radius"]
+    min_spacing = case.params.get(f"particle_cloud({cloud_idx})%min_spacing", 0.0)
+    min_dist = 2.0 * radius + min_spacing
+
+    for i in range(start, start + count):
+        xi, yi, zi = records[i][16:19]
+        for j in range(i + 1, start + count):
+            xj, yj, zj = records[j][16:19]
+            dist = math.sqrt((xi - xj) ** 2 + (yi - yj) ** 2 + (zi - zj) ** 2)
+            if dist < min_dist - 1.0e-12:
```
sbryngelson: **A fixed `1.0e-12` absolute tolerance does not survive single precision.**

`_read_ib_state_records` branches on `ARG("single") or ARG("mixed")` to pick 4-byte fields, but every comparison in this file then uses a fixed `1.0e-12` absolute slack. At these coordinates single precision carries roughly `1e-8` absolute error, four orders of magnitude past the slack.

It passes today only because four particles in a large shell never land near a constraint boundary. This check is the one that will bite first: rejection sampling drives accepted pairs toward exactly `min_dist` as density rises, so the first denser particle-cloud case anyone adds will fail the single and mixed CI jobs spuriously. The same applies to the radial and plane-clearance comparisons below.

Please scale the tolerance with the record precision rather than hardcoding it.

toolchain/mfc/test/test.py
```diff
@@ -48,6 +50,93 @@ class TestTimeoutError(MFCException):
     pass
 
 
+def _read_ib_state_records(filepath: str, precision: int):
+    if not os.path.isfile(filepath):
+        raise MFCException(f"Expected IB state file does not exist: {filepath}")
+
+    field_size = 8 if precision == 2 else 4
+    fmt = "<" + ("d" if precision == 2 else "f") * 20
+    record_size = 20 * field_size
```
sbryngelson: **The ib_state record layout is hardcoded here with no comment tying it to the writer.**

The `20` and the `record[16:19]` position slice below come from `m_data_output.fpp`: `NFIELDS_PER_IB = 20`, and `ib_buf(17:19)` holds `x/y/z_centroid`. Nothing in this file says so. If `ib_patch_parameters` gains a field or `ib_buf` is reordered, this reads the wrong columns and either passes vacuously or fails with a message that points nowhere useful.

Please name the constants (`NFIELDS_PER_IB`, a `CENTROID_SLICE`) and add a comment citing `s_write_serial_ib_state` as the source of truth.

Related, and worth a guard: this also assumes the non-`file_per_process` layout. Under `file_per_process = T` the writer emits `restart_data/lustre_0/ib_state_0_0000000.dat`, with a leading `num_local_ibs` integer and a `gbl_patch_id` integer before each 20-real record. A particle-cloud case with that flag set would abort on "Expected IB state file does not exist" rather than skipping the check. Add `file_per_process` to the early-return guard in `_assert_particle_cloud_ib_state`.

toolchain/mfc/test/cases.py
```diff
@@ -1044,6 +1044,98 @@ def alter_ib(dimInfo, six_eqn_model=False, viscous=False):
...
+                        "ib": "T",
+                        "num_ibs": 0,
+                        "num_particle_clouds": 1,
+                        "fd_order": 2,
+                        "ib_state_wrt": "T",
+                        "n": 49,
+                        "particle_cloud(1)%cloud_geometry": 2,
+                        "particle_cloud(1)%packing_method": 1,
+                        "particle_cloud(1)%x_centroid": 0.5,
+                        "particle_cloud(1)%y_centroid": 0.0,
```
sbryngelson: **The flat face sits exactly on the domain wall.**

`y_centroid = 0.0` with `y_domain%beg = 0.0` (set at cases.py:222). The plane-clearance rule then permits a particle tangent to the boundary: centre at `y = radius = 0.02`, surface at `y = 0`. Tangency is a degenerate configuration for IB ghost-point reconstruction and makes this golden more sensitive than it needs to be.

The 3D case sensibly uses `z_centroid = 0.1`. Please give the 2D case the same standoff.

Note also that the validator's `y_centroid < y_beg` check passes at exact equality, which is how this got through. Requiring at least one particle radius of clearance there would be more useful than requiring none.

toolchain/mfc/test/cases.py
```diff
@@ -1044,6 +1044,98 @@ def alter_ib(dimInfo, six_eqn_model=False, viscous=False):
...
+                        "particle_cloud(1)%num_particles": 4,
```
sbryngelson, line 1062: At 4 particles, `hash_size = max(16, 4*4) = 16` and every bucket chain has length at most 1. The shared `s_check_cloud_particle_overlap` these cases were added to cover is therefore only exercised in its trivial regime: the `do while (j > 0)` chain walk never iterates, and no hash collision ever occurs.

One case at roughly 50 particles would exercise the chain walk and the collision handling, which is where the refactor could actually regress. It would also give the rejection loop something to reject, which nothing currently does.

toolchain/mfc/case_validator.py
```diff
@@ -682,6 +687,55 @@ def check_ibm(self):
                 packing_method is not None and packing_method not in [1, 2],
                 f"particle_cloud({i})%packing_method must be 1 (rejection sampling) or 2 (lattice)",
             )
+            shell_outer_radius = self.get(f"particle_cloud({i})%shell_outer_radius", None)
+            shell_inner_radius = self.get(f"particle_cloud({i})%shell_inner_radius", None)
+            radius = self.get(f"particle_cloud({i})%radius", None)
+            self.prohibit(
+                geometry == 2 and (shell_inner_radius is None or shell_inner_radius < 0),
```
sbryngelson: Minor: this comparison and the `shell_outer_radius <= shell_inner_radius + 2 * radius` one below are evaluated without the `self._is_numeric(...)` guard the rest of this file applies consistently to real-valued params. The domain-extent block ten lines down does guard.

`None` is short-circuited correctly, but any non-numeric value raises an uncaught `TypeError` out of `validate()` instead of producing a clean `CaseConstraintError`.

toolchain/mfc/params/definitions.py
```diff
@@ -1096,11 +1099,8 @@ def _load():
             _r(f"simplex_params%perturb_vel_offset({d},{j})", REAL)
 
     # lag_params (Lagrangian bubbles)
-    # Members present in bubbles_lagrange_parameters: solver_approach, cluster_type,
-    # pressure_corrector, smooth_type, heatTransfer_model, massTransfer_model,
-    # write_bubbles, write_bubbles_stats, write_void_evol, pressure_force,
-    # gravity_force, nBubs_glb, epsilonb, charwidth, valmaxvoid. T0/Thost/c0/rho0/x0
-    # were removed from the Fortran type by upstream #1085/#1093 — they must NOT be
+    # Members present in bubbles_lagrange_parameters. T0/Thost/c0/rho0/x0 were
```
sbryngelson: Unrelated to this PR. This reflows a `lag_params` comment block that has nothing to do with particle clouds. Please drop it so the diff stays scoped.

sbryngelson (changes requested):
Thanks for the revisions. The design is now right, and everything from my previous two rounds landed:

- The `geometry` to `cloud_geometry` rename resolves the collision with `geom = 2` (circle) and `geom = 8` (sphere) in the same routine.
- Integer exponents are back in the innermost loop, and `s_get_cloud_bin` now runs once per candidate.
- The Python mirror implementation is gone in favour of parsing real `ib_state_0.dat` output.
- 2D, 3D, and box goldens all exist. All six golden dirs are new, so the squared-distance refactor did not perturb the box path.
- Sampling math verified independently: the 2D sqrt radial CDF and the 3D cube-root CDF with uniform `cos(polar)` both give uniform density, and the plane-clearance cut preserves uniformity on the remaining region.

`./mfc.sh precheck` passes all 7 checks, the 4 new pytest cases pass, and `./mfc.sh test -l` lists exactly the 6 new cases matching the 6 new golden dirs.

What is left is one test-fixture problem that undermines the 3D coverage I asked for, a validation-surface gap, and two latent bugs. Details inline.

**Blocking**

1. The 3D particles are sub-cell (`radius = 0.6*dx`), so the 3D golden may not be observing the feature it was added to cover. The 2D goldens do carry real signal; only the 3D case needs fixing.
2. The box branch of the new `ib_state` assertion is only valid for `packing_method = 1`. Separately, this exposed what looks like a genuine pre-existing bug in `s_particle_cloud_lattice`: its outer loops have no `ymax`/`zmax` bound, so lattice clouds can place particles outside the declared region and potentially outside the domain. Not yours to fix here, but it should get an issue.
3. The `m_checker` shell-thickness guard is vacuous when `radius` is unset, since `dflt_real = -1.e6_wp` swamps the right-hand side. More broadly, nothing requires `radius`, `mass`, or `num_particles` to be set for any cloud.
4. Domain containment is enforced only in `case_validator.py`. Please mirror it into `m_checker`, same as you did for the radii checks.
5. `cloud_geometry = 2` is unguarded in 1D.

**Design, and the one I would push hardest on**

`s_particle_cloud_random_hemi_shell` is still a near-verbatim copy of `s_particle_cloud_random_box`. Extracting the overlap check was the right first step, but roughly 60 of its 110 lines are still duplicated. Collapsing both into a single rejection packer with a geometry `select case` inside the loop removes that and flattens the dispatch in `s_generate_particle_clouds`. The `cloud_geometry` enum invites a third geometry, and a third geometry triples the copy.

**Housekeeping**

The PR description is stale. It still describes `particle_cloud(i)%packing_method = 3`, which is the design @wilfonba asked you to change, and it lists `examples/3D_mibm_particle_cloud_hemi_shell/case.py` under Testing though no such example is in the PR. Release notes get assembled from PR bodies, so please fix it before merge.

toolchain/mfc/test/test.py
sbryngelson: Correcting myself on the mechanism above, after actually simulating both loops rather than reasoning about them.

**2D does not escape.** I swept the triangular-lattice branch over several aspect ratios at `n` from 4 to 200 and `py` stayed inside `ymax` every time. The unbounded `row` loop looks wrong, but the in-region site count exceeds `n_target` in practice (the `floor(...)+1` edge effects work in favour), so it terminates before escaping. My "that is the common case" claim was wrong for 2D.

**3D does escape, but not via the unbounded `kz` loop.** The cause is `ncx = max(1, ceiling((xmax - xmin)/cell))` rounding the cell grid up past `xmax`, combined with the FCC basis `+0.5*cell` offsets landing sites in that overhang. On a unit region, `num_particles = 10` puts particles at 1.105 against a bound of 1.0, and `num_particles = 50` at 1.077. Anisotropic regions escape in y as well.

Filed as #1730 with a standalone repro.

None of this changes what I am asking for in this PR: please gate the box-bounds assertion on `packing_method == 1`. That is still correct, since lattice packing can violate the invariant this branch asserts. The rest is not yours to fix here.

sbryngelson (dismissed):

toolchain/mfc/test/cases.py
sbryngelson: Completed in 1fc7730 (goldens in 63444b9). The 3D hemisphere-shell path and the refactored box path are both exercised now with meaningful goldens (see the sub-cell and 50-particle threads), and the `ib_state` verifier added here checks the geometric constraints directly - radial/plane clearance, box bounds, min spacing - on the real Fortran output rather than a Python reimplementation.

toolchain/mfc/test/cases.py
sbryngelson: Fixed in 1fc7730 (goldens 63444b9): the 3D golden now uses `radius = 0.06` (diameter ~3.6 cells on the 29^3 grid) with `shell_outer_radius = 0.35`, so each sphere marks a multi-cell IB instead of a sub-cell one and the golden observes the feature.

toolchain/mfc/test/test.py
sbryngelson: Fixed in 1fc7730: the box-bounds assertion is now gated on `packing_method == 1`. The lattice out-of-region placement you traced is filed as #1730 and left untouched here.

toolchain/mfc/test/test.py
sbryngelson: Fixed in 1fc7730: named `_NFIELDS_PER_IB` and `_CENTROID_SLICE` with a comment citing `s_write_serial_ib_state` / `s_write_parallel_ib_state` as the source of truth, and added a `file_per_process` early return to `_assert_particle_cloud_ib_state`.

toolchain/mfc/test/cases.py
sbryngelson: Fixed in 1fc7730 (goldens 63444b9): the 2D case now uses `y_centroid = 0.1`, and `case_validator` now *requires* one particle radius of standoff (`y_centroid - radius >= y_beg`), so exact-wall tangency is rejected rather than accepted.

toolchain/mfc/test/cases.py
sbryngelson: Fixed in 1fc7730 (goldens 63444b9): the box case now packs 50 particles in a 0.6x0.6 region, so the hash chain walk, bucket collisions, and the rejection loop are all exercised.

src/simulation/m_checker.fpp
sbryngelson: Per the master merge (#1717, 1b4c55f), input-only checks were moved out of `m_checker` into `case_validator.py`, which already carries the x/y/z domain-containment checks - so mirroring back into `m_checker` would fight that split. Containment stays in the validator, now also enforcing one-radius flat-face clearance (1fc7730).

src/simulation/m_checker.fpp
sbryngelson: Added in 1fc7730, in `case_validator` (m_checker no longer hosts particle-cloud checks after #1717): `cloud_geometry == 2 and n == 0` is now rejected explicitly, ahead of the y-extent check that is skipped in 1D.

src/simulation/m_particle_cloud.fpp
sbryngelson: Orientation is documented in `case.md` and domain containment is enforced in `case_validator` (x/y/z extent plus one-radius flat-face standoff) as of 1fc7730.

docs/documentation/case.md
sbryngelson: Updated in 1fc7730: `case.md` now states the full-shell domain-containment requirement and the one-radius flat-face standoff alongside the orientation note.

src/simulation/m_particle_cloud.fpp
sbryngelson: Addressed by the single-packer refactor in 1fc7730: the box and hemisphere-shell rejection routines are now one `s_particle_cloud_rejection_pack`, with per-candidate geometry in `s_sample_cloud_candidate`.

src/simulation/m_checker.fpp
sbryngelson: Addressed in `case_validator` (1fc7730) rather than `m_checker`, per #1717: `radius`, `mass`, and `num_particles` now have unconditional positivity checks, so an unset `radius` is named directly instead of surfacing later as a negative `min_dist` or the generic shell abort.

src/simulation/m_particle_cloud.fpp
sbryngelson: Done in 1fc7730: `s_particle_cloud_random_box` and `s_particle_cloud_random_hemi_shell` are collapsed into one `s_particle_cloud_rejection_pack`, with a geometry `select case` in `s_sample_cloud_candidate` yielding `(rx, ry, rz)` plus a reject flag, and `s_generate_particle_clouds` back to a flat dispatch. Net -254 lines in `simulation/`; per-candidate draw order is preserved so placement is unchanged aside from the intended RNG fix.

src/simulation/m_particle_cloud.fpp
sbryngelson: Removed in 1fc7730: the redundant `r_inner`/`r_outer` check is gone; radii feasibility lives solely in `case_validator`.

src/simulation/m_particle_cloud.fpp
sbryngelson: Resolved by the refactor in 1fc7730: the nested `select case` collapsed to a flat dispatch on `packing_method`; the hemi-shell + lattice arm is gone, with `case_validator` the single source of truth.

src/simulation/m_particle_cloud.fpp
sbryngelson: Fixed in 1fc7730: `f_xorshift` now uses `real(iand(seed, huge(seed)), wp)/real(huge(seed), wp)` instead of `abs(...)`, so the sign-bit fold can no longer return -1 and produce a NaN shell centre. Goldens regenerated for the changed stream (63444b9).

toolchain/mfc/test/test.py
sbryngelson: Fixed in 1fc7730, with one correction: `wp` is single only under `--single` (`--mixed` keeps `wp` double, narrowing only `stp`), and `ib_buf` is `real(wp)` written with `mpi_p`, so the reader keys field width off `ARG('single')` alone rather than single-or-mixed.

toolchain/mfc/test/test.py
sbryngelson: Fixed in 1fc7730: the geometric tolerance now scales with precision (`1e-6` under `--single`, else `1e-12`), matching `case.py`'s single-precision convention.

toolchain/mfc/case_validator.py
sbryngelson: Fixed in 1fc7730: both radii comparisons are now `_is_numeric`-guarded, so a non-numeric value yields a clean `CaseConstraintError` rather than a `TypeError`.

toolchain/mfc/params/definitions.py
sbryngelson: Reverted in 1fc7730: the `lag_params` comment block is back to master's wording, so the diff stays scoped to particle clouds.

sbryngelson:
Architectural note for after this lands (tracked in #1737): the particle-cloud packer is really geometry *preprocessing* — rejection/lattice/hemisphere-shell/periodic sampling, spatial hashing — that currently sits in the solver startup only because IBM marker computation is simulation-side. It'd be cleaner to expand the compact `particle_cloud` spec in pre_process, write the IB-patch list to a file, and have the solver just read it (net ~-525 lines in `simulation/`). Not blocking this PR — flagging so the subsystem doesn't keep accreting knobs solver-side. Design sketch in #1737.

sbryngelson (approved):

BCKim55:
Thanks Spencer, agreed. Moving the particle-cloud packing into pre_process sounds like the right follow-up.

Also, thanks for pushing the final cleanup commits and helping get this PR over the line. I’ll keep this PR scoped to the current implementation, and #1737 can be handled separately after this lands.

sbryngelson (approved):

PR #1668, opened 2026-07-22
build: add -fopenmp-assume-no-nested-parallelism on the AMD offload path

sbryngelson:
## Description

Adds `-fopenmp-assume-no-nested-parallelism` to the `LLVMFlang` OpenMP-offload compile options. MFC has no nested parallel regions, so the assumption holds.

The reason is correctness, not speed. The two oversubscription flags already on that line are the trigger for [llvm/llvm-project#198621](https://github.com/llvm/llvm-project/issues/198621): with both set, a deterministic suffix of loop iterations is silently skipped in kernels whose bodies use array expressions. On AFAR 23.2.1 the upstream reproducer fails 31 of 64 iterations, three runs out of three. `-O3` masks it, which is why we are not hit today, but that is a property of the LTO pipeline rather than a guarantee. The flag folds `MayUseNestedParallelism` to 0 — the step #198621 identifies as blocking the `omp_get_num_threads()` fold — and the reproducer then passes.

It costs nothing on the compilers we ship against. Kernel resource usage is byte-identical with and without it:

| toolchain / target | NEQ=8 | NEQ=16 | NEQ=24 |
|---|---|---|---|
| AFAR 23.2.1, gfx90a | 72 / 0 / 7 | 124 / 0 / 4 | 70 / 392 / 7 |
| amdflang 22 (ROCm 7.2.0), gfx942 | 62 / 136 / 8 | 62 / 264 / 8 | 62 / 392 / 8 |

VGPR / scratch / occupancy, identical in every cell with the flag on or off, because both AMD forks already inline the outlined target region downstream. Throughput moves 0–2%, inside run-to-run spread.

It becomes a real gain if that downstream inlining goes away — [ROCm/llvm-project#3485](https://github.com/ROCm/llvm-project/pull/3485) removes it — or on an upstream-based toolchain, where the same flag is worth up to 2.5x on a WENO5 + HLLC kernel (upstream flang, gfx942, NEQ=16: 1598.7 to 3987.4 Mcell/s).

Not doing the obvious alternative: dropping the oversubscription flags instead would also remove the exposure, but they are worth 4–12% on both AMD compilers, so they earn their place.

Background on why the outlined region is left un-inlined at all, and the upstream fix for it, is in [llvm/llvm-project#211132](https://github.com/llvm/llvm-project/issues/211132) and [llvm#211287](https://github.com/llvm/llvm-project/pull/211287).

### Type of change

- Bug fix

## Testing

Full golden-file suite on gfx90a (MI250X) with AFAR 23.2.1, built and run both with and without the flag in the same job:

| | passed | failed |
|---|---|---|
| with flag | 568 | 21 |
| without flag | 568 | 21 |

Identical failing sets, not merely identical counts — all 21 are pre-existing `weno_order=7` tolerance mismatches from the AFAR 23.2.x `nuw` miscompile ([ROCm/llvm-project#3471](https://github.com/ROCm/llvm-project/issues/3471)), unrelated to this change.

Also verified on a standalone WENO5 + HLLC reproducer that output checksums are bit-identical with and without the flag, on gfx90a (AFAR 23.2.1) and gfx942 (ROCm 7.2.0 and upstream flang).

## Checklist

Neither applies: build-flag change, no user-facing behavior and no new code path to test. Coverage is the existing golden-file suite, run both ways above.

<details>

Golden-file suite is the CPU-reference comparison; run on MI250X and additionally checked on MI300X.

</details>


sbryngelson:
Both applied in c8a07e50 — one flag per line, and the reference spelled `llvm/llvm-project#198621`. Precheck clean.

PR #1675, opened 2026-07-25
Remove non-operational hyperelasticity capability

sbryngelson:
## Description

Remove the currently exposed hyperelastic reference-map capability while its implementation is repaired in a stacked restoration branch.

The audited implementation is not reachable through a normal case because `hyper_model` is absent from the parameter registry and generated namelist. It also contains a 3D-only kernel exposed to lower-dimensional cases, computes the wrong Cauchy–Green contraction for rotated finite deformations, and leaves derived state fluxes unassigned. Keeping the option visible therefore overstates the supported physics and permits undefined behavior.

This PR removes the user-facing parameters, reference-map state, constitutive module, Riemann-solver hooks, initialization/output paths, validator rules, and documentation claims. Shared hypoelastic infrastructure remains in place.

Related to #1674. A stacked restoration PR will add the corrected, constrained, and tested implementation back on top of this branch; after this PR merges, that restoration PR should be retargeted to `master`.

### Type of change (delete unused ones)

- Bug fix
- Refactor
- Documentation

## Testing

- `./mfc.sh precheck`
- `./mfc.sh build -t pre_process simulation post_process --no-mpi --debug -j 4`
- `./mfc.sh test --from 32A3A936 --to 32A3A936 --no-mpi --debug -j 1`
- Repository-wide search confirms no remaining hyperelastic parameter, state index, module, solver hook, or documentation claim outside build/test artifacts.

## Checklist

- [x] I added or updated tests for new behavior (existing hypoelastic regression remains passing; no hyperelastic regression existed)
- [x] I updated documentation if user-facing behavior changed

See the [developer guide](https://mflowcode.github.io/documentation/contributing.html) for full coding standards.

<details>
<summary><strong>GPU changes</strong> (expand if you modified <code>src/simulation/</code>)</summary>

- [ ] GPU results match CPU results
- [ ] Tested on NVIDIA GPU or AMD GPU

The PR removes an unvalidated GPU path; no GPU hardware was available for this audit.

</details>

## AI code reviews

Reviews are not retriggered automatically. To request a review, comment on the PR:
- `@claude full review` — Claude full review (also triggers on PR open/reopen/ready)
- Or add label `claude-full-review` — Claude full review via label

## AI-assistance disclosure

This removal and its supporting audit were prepared with AI assistance. The resulting diff was checked with MFC's precheck, a fresh CPU debug build, and a focused hypoelastic regression before publication.

PR #1677, opened 2026-07-25
Restore constrained neo-Hookean reference-map hyperelasticity

### LINKED ISSUE #1674: Hyperelastic reference-map implementation is currently non-operational and internally inconsistent
@mrodrig6 @ChrisZYJ — this audit found several independent problems in the current hyperelastic path. The proposed workflow is to remove the exposed, non-operational capability in one PR and add it back through a stacked restoration PR after the implementation and validation below are complete.

## Summary

At audited commit `8821173c20b645f1869b0914f4a64673d6021455`, the hyperelastic implementation should not be treated as a working user-facing capability:

1. `hyper_model` is initialized to `dflt_int` (`-100`) but is absent from the parameter registry and generated namelist. Therefore neither constitutive routine is callable from a normal MFC case. The routine nevertheless consumes the derived elastic-energy field immediately afterward.
2. `m_hyperelastic.fpp` is structurally 3D-only but validation permits 2D. The kernel unconditionally uses all nine gradient components and the y/z finite-difference arrays; in 2D, `tensor_size == 5` while entries through 9 are accessed.
3. The tensor contraction computes the wrong Cauchy–Green tensor. With the current storage order, the inverted tensor is `F^T`; the existing row-dot-row contraction produces `F^T F` (right Cauchy–Green), while the spatial Cauchy stress requires `b = F F^T` (left Cauchy–Green).
4. Stress and the derived elastic-energy coefficient are included in `sys_size` and therefore read by the generic RHS flux divergence, but HLLC does not assign fluxes for those derived entries. HLL does not assign the reference-map flux either.
5. `s_Mooney_Rivlin_cauchy_solver` is algebraically identical to the neo-Hookean routine and has no second-invariant contribution or independent material coefficient.
6. The HLLC elastic-energy flux uses `vel_[LR](dir_idx(i))` in one part of the expression but `vel_[LR](i)` in its star-state correction, so transverse-direction fluxes use inconsistent velocity components.
7. The HLLC energy initialization includes both left and right hyperelastic energies only when both `G_L` and `G_R` are nonzero. Each side must be gated independently.
8. If `det(grad xi) <= verysmall`, the constitutive projection silently retains stale stress/energy instead of reporting loss of map admissibility.
9. There are no hyperelastic regression cases or golden-test metadata in the repository.

## Intended provenance and constitutive decision

The reference-map approach itself is appropriate. The project-specific formulation is documented most directly by Carcana Barbosa, Rodriguez Jr., and Yang (2024):

- `F = (grad xi)^(-1)`
- `b = F F^T`
- `sigma_dev = (G/J) dev(b)`
- conservative reference-map transport `d_t(rho xi) + div(rho u tensor xi) = 0`

The same closure is reported in the MFC 5.0 paper. Therefore the immediate tensor fix should implement `b = F F^T`, not change the overall modeling family.

There is a separate constitutive question that should be decided explicitly with the model authors. Kamrin, Rycroft, and Nave's compressible neo-Hookean example uses the isochoric invariant `J^(-2/3) tr(b)` and obtains a deviatoric factor proportional to `G J^(-5/3) dev(b)`, whereas the MFC/Barbosa formulation uses `G J^(-1) dev(b)`. The restoration PR should not silently change this exponent; it should preserve the published MFC closure unless the authors confirm a different intended strain-energy split.

## Concrete correction for the tensor contraction

The current derivative layout is

```fortran
! [xi_x,x xi_y,x xi_z,x;
!  xi_x,y xi_y,y xi_z,y;
!  xi_x,z xi_y,z xi_z,z] = (grad xi)^T
```

After inversion, `tensora` stores `F^T`. Thus `b = F F^T = transpose(tensora)*tensora`:

```fortran
tensorb(1) = tensora(1)**2 + tensora(4)**2 + tensora(7)**2
tensorb(5) = tensora(2)**2 + tensora(5)**2 + tensora(8)**2
tensorb(9) = tensora(3)**2 + tensora(6)**2 + tensora(9)**2
tensorb(2) = tensora(1)*tensora(2) + tensora(4)*tensora(5) + tensora(7)*tensora(8)
tensorb(3) = tensora(1)*tensora(3) + tensora(4)*tensora(6) + tensora(7)*tensora(9)
tensorb(6) = tensora(2)*tensora(3) + tensora(5)*tensora(6) + tensora(8)*tensora(9)
```

A rotated uniaxial stretch is the essential regression test: it distinguishes `F F^T` from `F^T F`, unlike diagonal stretches and simple identity checks.

## Recommended restoration scope

1. Register `hyper_model`; default to neo-Hookean and reject unsupported values. Do not expose Mooney–Rivlin until it has a real constitutive law and parameters.
2. Initially constrain the restored implementation to 3D Cartesian HLLC, matching the actual kernel. Add 2D/axisymmetric support separately.
3. Correct `b = F F^T`.
4. Keep only `rho xi` as evolved hyperelastic state. Store derived stress and elastic energy as auxiliary/projected fields rather than conservation equations. If a minimal transitional implementation retains those slots, initialize every flux and re-project after every Runge–Kutta stage.
5. Compute the HLLC reference-map flux as the upwind reference map multiplied by the actual mixture mass flux; add an explicit HLL passive-scalar flux only if HLL is supported.
6. Fix the rotated-direction velocity indices and gate left/right energies independently.
7. Fail clearly or apply a documented admissibility policy when `det(grad xi) <= 0`.
8. Add CPU bounds-checking tests and analytic tests for identity, rigid rotation, simple shear, rotated uniaxial stretch, uniform reference-map advection, and determinant failure. Add GPU parity before declaring the restored capability ready.

## References

- Kamrin, Rycroft, and Nave, *Reference map technique for finite-strain elasticity and fluid–solid interaction*, JMPS 60 (2012), DOI: https://doi.org/10.1016/j.jmps.2012.06.003, open PDF: https://www.math.mcgill.ca/jcnave/papers/elasticity.pdf
- Carcana Barbosa, Rodriguez Jr., and Yang, *Numerical simulations of inertial microbubble collapse at a gel–water interface with finite elasticity and phase change*, CTR Summer Program 2024: https://web.stanford.edu/group/ctr/ctrsp24/iv07_CARCANA.pdf
- Wilfong et al., *MFC 5.0: An exascale many-physics flow solver*, CPC 322 (2026), DOI: https://doi.org/10.1016/j.cpc.2026.110055, open PDF: https://benwilfong.com/papers/wilfong-mfc-26.pdf

## AI-assistance disclosure

This issue was prepared with AI assistance. The assistant inspected the cited source files and generated namelist at the audited commit, compared the tensor layout and constitutive formulas against the references above, and ran repository searches for parameter registration and test coverage. The conclusions and proposed fixes should receive maintainer/model-author review before merge.

sbryngelson:
## Description

This is the add-back half of the stacked workflow proposed in #1674. It is intentionally based on the removal branch in #1675, so this PR shows the hyperelastic capability being added back from a clean baseline. After #1675 merges, this PR should be retargeted to `master`.

The restored implementation is deliberately narrower than the prior interface: neo-Hookean only, 3D Cartesian only, finite-difference gradients enabled, and HLLC only. It preserves the project-specific Barbosa/Rodriguez-family closure already used by MFC, `tau = (G/J) dev(b)`, rather than silently replacing it with the different `J^(-5/3)` closure in Kamrin et al.

The repair:

- registers `hyper_model` in the generated parameter/namelist system and accepts only `hyper_model = 1`;
- rejects unsupported dimensionality, coordinates, gradient configuration, model equations, and Riemann solvers during case validation;
- removes the undocumented spherical `pre_stress` initializer;
- corrects the tensor contraction to compute `b = F F^T` from the stored `F^T` layout;
- removes the duplicate pseudo-Mooney-Rivlin routine;
- initializes and consistently converts the derived stress and elastic-energy slots;
- gives every hyperelastic state slot a defined HLLC flux and uses the HLLC mass flux for the conservative reference-map transport;
- fixes direction indexing in the elastic energy flux and treats the left/right elastic energies independently;
- adds an always-run, deforming 3D shear canary with nonzero off-diagonal elastic stresses.

The audit, concrete equations, provenance discussion, references, and remaining architectural caveats are in #1674.

Closes #1674.

- Bug fix
- Refactor
- Documentation

## Testing

- `./mfc.sh precheck`
- `./mfc.sh build -t pre_process simulation post_process --no-mpi --debug -j 4`
- `./mfc.sh test --from C06DAA3C --to C06DAA3C --no-mpi --debug -j 1`
- `./mfc.sh test --from 32A3A936 --to 32A3A936 --no-mpi --debug -j 1`

The new `C06DAA3C` golden case is a 3D shear problem. It evolves the reference map away from the identity and produces nonzero off-diagonal elastic stresses, so it exercises the corrected `b = F F^T` path and the HLLC elastic/reference-map fluxes.

## Checklist

<details>

GPU validation remains to be done.

</details>

## AI assistance disclosure

This audit and patch were AI-assisted. The technical claims were checked against the repository source, CPU debug builds/runs, the new regression output, and the references collected in #1674. Human review is especially requested for the intended constitutive closure and the transitional treatment of derived stress/energy slots.

Review requested from @mrodrig6 and @ChrisZYJ.


sbryngelson:
## Follow-up correctness audit: this should remain a draft

This is an AI-assisted follow-up audit of the add-back at `4aa4488`. The constitutive core is substantially improved, but I do not think the current PR is yet a correct end-to-end restoration.

### Parts that appear correct

- With the storage layout used in `m_hyperelastic.fpp`, the corrected contraction computes `b = F F^T`, not `F^T F`.
- Retaining `tau = (G/J) dev(b)` is consistent with the Barbosa/Rodriguez-family formulation already used by MFC; it should not be silently replaced by the different isochoric closure in Kamrin et al.
- Conservative reference-map transport as `rho xi`, using the mass flux, is conceptually sound.
- Removing the unsupported pseudo-Mooney-Rivlin path and registering/constraining `hyper_model` are good changes.

### 1. Reference-map boundary conditions are currently incorrect

The new `C06DAA3C` test uses `BC_GHOST_EXTRAP = -3` on all six faces. The generic routine copies the nearest interior value into every ghost cell. For the undeformed map `xi_x = x`, second-order centered differentiation at the first cell therefore gives

```text
(d xi_x/dx)_0 = (xi_1 - xi_-1)/(2 dx) = 1/2
```

rather than one. On an x face this creates

```text
F = diag(2,1,1),  J = 2
```

and a nonzero elastic stress in an initially undeformed body. Thus the golden test starts with numerical pre-strain near its boundaries.

This should not be addressed by rejecting most MFC boundary conditions. The completed capability needs a dedicated reference-map boundary layer covering every applicable MFC BC:

- **Periodic (`-1`)**: the reference map is affine-periodic. Copy the tangential components normally, but offset the normal component by the reference-domain length. For example, at an x wrap,

  ```fortran
  xi_x(-j,k,l)  = xi_x(m-(j-1),k,l) - Lx_ref
  xi_x(m+j,k,l) = xi_x(j-1,k,l)     + Lx_ref
  ```

  The same offset must be applied at global periodic MPI wraps, not at ordinary internal rank boundaries.

- **Reflective/symmetry (`-2`) and characteristic/slip walls (`-5`, `-15`)**: reflect the normal reference coordinate about the reference-coordinate wall location,

  ```fortran
  xi_n(ghost) = 2*Xi_wall - xi_n(mirror)
  ```

  and mirror the tangential reference-map components evenly. The current unconditional `-xi_n(mirror)` is only valid when `Xi_wall = 0`; it is wrong at an upper wall in a general domain.

- **Ghost/Riemann extrapolation and outflow families (`-3`, `-4`, `-8`, `-9`, `-10`, `-12`)**: use affine-preserving extrapolation for every `xi` component. For second order,

  ```fortran
  xi_i(-j)  = (j+1)*xi_i(0) - j*xi_i(1)
  xi_i(m+j) = (j+1)*xi_i(m) - j*xi_i(m-1)
  ```

  This preserves the identity map and any other affine deformation exactly. If fourth-order reference-map gradients are supported, the ghost construction must be raised consistently or one-sided fourth-order derivative stencils must be used.

- **No-slip walls (`-16`)**: distinguish a fixed/clamped solid wall from a slip/symmetry wall. The normal reference component reflects about `Xi_wall`. Tangential components require the prescribed wall reference map for a clamped or moving wall; simply copying them imposes a zero-normal-gradient condition and is not a general no-slip solid boundary.

- **Subsonic/supersonic inflow and Dirichlet (`-6`, `-7`, `-11`, `-17`)**: incoming material needs prescribed reference-map data. Guessing `xi=x` at every time is not sufficient for uniform advection (`xi=x-u t`) and cannot represent a prescribed incoming deformation. Extend the BC input/buffer interface with reference-map data, preferably a boundary reference origin plus `F_in` (or equivalently `grad_xi_in`) and its time dependence. Dirichlet `bc_buffers` must populate these slots explicitly rather than relying on whatever values happen to occupy the expanded `sys_size`.

- **Null (`-13`)**: retain its existing no-op meaning, but do not advertise it as a usable physical hyperelastic boundary if the finite-difference stencil would read unfilled ghosts.

- **Axis (`-14`)**: this is not applicable to the PR's stated 3D Cartesian-only scope. Literal support for the axis BC requires cylindrical/axisymmetric reference-map gradients, tensor components, geometric sources, and constitutive terms; it is a separate model extension, not merely another ghost-fill case. It should remain explicitly tracked rather than silently accepted.

A useful implementation pattern is to run the existing fluid BC routine and then call one centralized `s_apply_reference_map_bc(...)` that overwrites only `eqn_idx%xi` according to the BC code. This avoids scattering slightly different `xi` rules through every generic boundary primitive.

### 2. The transverse HLLC elastic-energy correction is structurally wrong

The current correction contains, for transverse components,

```fortran
(s_S - u_t) * tau_nt / (s_K - u_t)
```

Here `s_S` and `s_K` are wave-normal speeds while `u_t` is a transverse velocity. An otherwise admissible state can have `u_t = s_K`, producing an artificial singularity. Also, `s_S` is not the transverse star velocity. Replacing only the denominator with `s_K-u_n` removes one singularity but does not provide the missing transverse star state.

The safest first restoration is to use the two-wave HLL flux for the complete conservative hyperelastic state:

```text
F_HLL = (s_R F_L - s_L F_R + s_L s_R (U_R-U_L))/(s_R-s_L)
```

with physical fluxes

```text
F_mom = rho u_n u + p e_n - tau e_n
F_E   = (E+p)u_n - u . (tau e_n)
F_rhoxi = rho u_n xi
```

MFC's HLL implementation already has partial hyperelastic wave-speed logic, but its stress-aware momentum and energy branches and all `rho xi` state fluxes need to be completed. This is more diffusive than HLLC but avoids inventing shear-star relations.

If HLLC is retained, the elastic branch should instead be rewritten from the Rankine-Hugoniot relations. Let `P_K = p_K - tau_nn,K`. A consistent reduced three-wave assumption is `u_t* = u_t` and `tau_nt* = tau_nt`, giving

```text
rho_K* = rho_K (S_K-u_n,K)/(S_K-S*)
P_K*   = P_K + rho_K (S_K-u_n,K)(S*-u_n,K)
E_K*   = ((S_K-u_n,K)E_K - P_K u_n,K + P_K* S*)/(S_K-S*)
F_E,K* = (E_K*+P_K*)S* - sum_t u_t,K tau_nt,K
```

This has no `S_K-u_t` denominator and satisfies the assumed three-wave jump relations. It still does not resolve the physical shear waves; resolving those requires a larger elastic HLLC/HLLD-type wave model.

Until that work is complete, validation should at least require direct wave speeds (`wave_speeds = 1`). The pressure-based branch currently omits the elastic changes used by the direct branch.

### 3. Invalid maps must not silently lose elasticity

`m_hyperelastic.fpp` zeros stress and elastic energy before checking `det(grad xi)`, and leaves them zero if the determinant is invalid. That silently turns a failed solid cell into a locally fluid-like cell.

Use a device-safe invalid-map counter/flag, record the minimum determinant, reduce it across MPI ranks, and terminate with a clear diagnostic after the kernel. If regularization is desired instead, it needs to be an explicit documented model with diagnostics, not implicit zeroing.

### 4. The current golden test is an execution canary, not a correctness test

Because its ghost extrapolation already strains the identity map, `C06DAA3C` cannot validate the mechanics. Add analytic assertions for:

- identity, including every supported boundary type: `F=I`, `J=1`, `tau=0`, elastic energy zero;
- rigid rotation: `b=I`, zero stress;
- simple shear with every component checked analytically;
- rotated stretch, which distinguishes `F F^T` from `F^T F`;
- affine-periodic maps across serial and MPI periodic wraps;
- uniform inflow/advection with the time-dependent reference map;
- x/y/z rotations of the same Riemann state;
- the formerly singular case `u_t = S_L`;
- invalid determinant producing the intended failure;
- model-equation and CPU/GPU parity for every configuration claimed as supported.

Until these points are addressed, this PR should remain draft and should not close #1674.

### References

- Carcana Barbosa, Rodriguez, Yang et al., CTR Summer Program 2024: https://web.stanford.edu/group/ctr/ctrsp24/iv07_CARCANA.pdf
- MFC equation documentation: https://mflowcode.github.io/documentation/equations.html
- Rodriguez et al., JCP 2019: https://doi.org/10.1016/j.jcp.2018.10.035

This comment and audit were AI-assisted; the findings above are based on direct inspection of the current PR source and the cited formulations.

sbryngelson:
can be restored later if needed

PR #1695, opened 2026-07-28
Move Frontier to CCE 21.0.2 / ROCm 7.2.0

sbryngelson:
Moves Frontier off `cce/19.0.0`. Context: #1684.

**Stacked on #1688** (QBMM hoist), which is the tenth fix and is deliberately not duplicated here. QBMM tests abort on CCE 21 until that merges; everything else in this PR is independent of it.

## Result

Frontier, CCE 21.0.2 / ROCm 7.2.0, MPI, full suite:

| backend | passed | failed | aborts |
|---|---|---|---|
| `--gpu acc` | **627** | 0 | 0 |
| `--gpu mp` | **627** | 0 | 0 |

And it is **5.8% faster than CCE 19.0.0** — same source, full fix set on both trees, six benchmark cases, faster in every one:

| case | CCE 19 | CCE 21 | |
|---|---|---|---|
| `5eq_rk3_weno3_hll` | 53.17 | 50.93 | −4.23% |
| `5eq_rk3_weno3_hllc` | 53.96 | 50.19 | −6.99% |
| `5eq_rk3_weno3_lf` | 53.06 | 50.27 | −5.24% |
| `hypo_hll` | 93.68 | 87.22 | −6.90% |
| `igr` | 18.33 | 17.93 | −2.16% |
| `viscous_weno5_sgb_acoustic` | 56.06 | 52.56 | −6.25% |
| **total** | **328.27** | **309.10** | **−5.84%** |

Single runs, no repeats; the sign is uniform. `-mai-insts` is in effect on the CCE 21 side, so the AGPR workaround's cost is already priced in.

Worth noting chemistry on GPU does not work at all under CCE 19 — every chemistry case memory-faults there — so this is the first configuration in which those cases run.

## The application-side fixes

**`fix: honor cray_inline/cray_noinline on Cray GPU builds and use INLINENEVER`** — `!DIR$ NOINLINE <name>` is not a valid named directive; ftn rejects it with ftn-790. The named form is `INLINENEVER`. Separately, the directive was being emitted *instead of* the offload directive rather than alongside it, so on Cray GPU builds `cray_noinline` was silently a no-op. Collapses three near-identical branches into one, −30 LOC.

**`fix: pass IBM pb/mv unconditionally instead of as optional args`** — an absent `optional` array reaches the OpenACC data environment as a null dope vector, which CCE 21 dereferences. `pb_ts`/`mv_ts` are always allocated (zero-sized when unused), so pass them unconditionally. No `present()` guards existed in `m_ibm`, so this is logic-neutral.

**`build: base dependency and configure skips on real state, not CMakeCache.txt`** — [#1689](https://github.com/MFlowCode/MFC/issues/1689). `is_configured()` checked only for `CMakeCache.txt`, which CMake writes before generating the build system and long before installing. An interrupted configure then makes MFC skip configure and fail with `No rule to make target 'Makefile'`; a crashed dependency build makes MFC skip that dependency *permanently*, pointing `CMAKE_PREFIX_PATH` at an empty install tree. Now requires the generator's build file for "configured" and `install_manifest.txt` for "installed".

This one is not CCE-specific and is worth taking on its own. Over this port it destroyed six batch jobs, produced a false negative in a compiler-flag experiment, broke a control run, and silently voided two benchmark campaigns — every time presenting as a compiler or missing-package problem.

## The compiler workarounds

Each of these works around a CCE 21 defect. Reproducers, with verified reproduce steps, are at <https://github.com/sbryngelson/compiler-bugs/tree/main/cce>.

**`fix: avoid CCE 21 dropping dynamically-indexed private-array stores`** — `AMDGPUPromoteAllocaToVector` silently discards a store to `arr(v)` when `arr` is 1-based and `v` is a runtime value. CCE canonicalizes the element index as `v + 0x3FFFFFFF`, harmless in wrapping address arithmetic but not once the array is registerized and the index becomes an equality comparison. `idx(norm_dir) = idx(norm_dir) + 1` never executed, so the viscous flux averaged the gradient at cell `j` instead of `j+1`. Three sites rewritten to use constant indices.

**`fix: keep Y_rs off private frame offset 0 in s_compute_pressure`** — ftn builds a flat pointer to a private object as `zext(offset)` with no aperture. The flat→private cast back is `select(p == null, ~0, trunc p)`, so an object at frame offset 0 yields `0xFFFFFFFF` and the store lands 4 GiB out of bounds. A lower bound of 0 as an unused pad keeps `Y_rs(1)` off offset 0.

**`fix: emit no defaultmap clauses for CCE OpenMP offload`** — the largest single effect. Any `defaultmap` clause makes a device-resident array read as **all zeros** inside the target region. MFC emitted them on every OpenMP `target` loop as its translation of `default(present)`, so `ib_markers` read as empty, the ghost-point count came back 0, the allocation was zero-length, and the first store faulted through a host pointer. Removing them took the OpenMP suite from 566/627 to 622/627 in one change, with nothing that previously passed regressed. Dropping them is also semantically free — OpenMP 5.0 already defaults scalars to firstprivate, and the AMD branch never emitted them.

**`build: move Frontier to CCE 21.0.2 and ROCm 7.2.0`** and **`build: add -disable-promote-alloca-to-vector`** — two `-plugin-opt` values in `CRAY_CCE_LLD_ARGS`. The first works around a back-end assertion in `AMDGPU Rewrite AGPR-Copy-MFMA` that blocks the device link outright; the second covers the promote-alloca sites the source fix above does not reach (`m_ibm`, `m_riemann_solver_hll`, `m_qbmm`, `m_collisions` all still carry markers). Both are needed: with only the source fix, five tests still fail, one of them by 1.9e-3.

**The quoting is load-bearing.** `modules.sh:126` does `eval "export $line"`, so an unquoted multi-word value exports only the first `-plugin-opt` and reports the rest to stderr, which `mfc.sh load` output is routinely discarded. Unquoted, this ships the AGPR workaround and silently loses the correctness fix.

## Reviewer notes

`toolchain/modules` pins compiler-internal flags, which will rot. The alternative is patching the remaining promote-alloca sites in source so the flag can be dropped — more work, but not dependent on a flag surviving future CCE releases. Happy to do that instead if preferred.

`./mfc.sh precheck` passes 7/7. `m_qbmm.fpp` is untouched here by design.

### DIFF
--- a/src/common/include/omp_macros.fpp
+++ b/src/common/include/omp_macros.fpp
@@ -26,7 +26,20 @@
             #:if MFC_COMPILER == NVIDIA_COMPILER_ID or MFC_COMPILER == PGI_COMPILER_ID
                 #:set default_val = 'defaultmap(tofrom:aggregate) defaultmap(tofrom:allocatable) defaultmap(tofrom:pointer) '
             #:elif MFC_COMPILER == CCE_COMPILER_ID
-                #:set default_val = 'defaultmap(tofrom:aggregate) defaultmap(present:allocatable) defaultmap(present:pointer) '
+                #! Emit no defaultmap at all on CCE. MEASURED on 21.0.2: the presence of
+                #! ANY defaultmap clause makes a resident (declare target + enter data)
+                #! array read as ZERO inside the target region. A 60-line reproducer with
+                #! a pointer component, an allocatable component and a bare module array
+                #! reads all three correctly with a bare directive and gets 0 from all
+                #! three with defaultmap added -- and each of tofrom:aggregate,
+                #! present:allocatable and present:pointer reproduces it on its own.
+                #! This is what made every IBM case fail under -homp: ib_markers read as
+                #! empty, so the ghost-point count came back 0.
+                #! Related: present:aggregate was tried and reverted earlier. It aborts
+                #! with "find_in_present_table failed for 'length(:)'" at
+                #! m_ib_patches.fpp:145 -- a genuine second residency bug that the tofrom
+                #! default had been masking, worth fixing separately.
+                #:set default_val = ''
             #:elif MFC_COMPILER == AMD_COMPILER_ID
                 #:set default_val = ''
             #:else
@@ -179,7 +192,12 @@
     #:if MFC_COMPILER == NVIDIA_COMPILER_ID or MFC_COMPILER == PGI_COMPILER_ID
         #:set omp_start_directive = '!$omp target teams loop defaultmap(firstprivate:scalar) bind(teams,parallel) '
     #:elif MFC_COMPILER == CCE_COMPILER_ID
-        #:set omp_start_directive = '!$omp target teams distribute parallel do defaultmap(firstprivate:scalar) '
+        #! No defaultmap(firstprivate:scalar) here: it only restates the OpenMP 5.0
+        #! default for scalars in a target region, and CCE 21 lets it override an
+        #! explicit map(to:) on a scalar. A scalar that is atomically incremented to
+        #! hand out array slots then becomes per-thread, so every thread gets the
+        #! same index -- this broke every immersed-boundary case under -homp.
+        #:set omp_start_directive = '!$omp target teams distribute parallel do '
     #:elif MFC_COMPILER == AMD_COMPILER_ID
         #:set omp_start_directive = '!$omp target teams distribute parallel do '
     #:else
--- a/src/common/include/omp_macros.fpp
+++ b/src/common/include/omp_macros.fpp
@@ -35,10 +35,14 @@
                 #! present:allocatable and present:pointer reproduces it on its own.
                 #! This is what made every IBM case fail under -homp: ib_markers read as
                 #! empty, so the ghost-point count came back 0.
-                #! Related: present:aggregate was tried and reverted earlier. It aborts
-                #! with "find_in_present_table failed for 'length(:)'" at
-                #! m_ib_patches.fpp:145 -- a genuine second residency bug that the tofrom
-                #! default had been masking, worth fixing separately.
+                #! Related: present:aggregate was tried and reverted. It aborts with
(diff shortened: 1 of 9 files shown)


sbryngelson:
Closing as a duplicate of #1694 — same port, opened in parallel from a second worktree. #1694 is the more complete of the two: it carries the QBMM hoist (so it does not depend on #1688) and a hipfort-tag fix that derives the tag from `ROCM_PATH`, neither of which is here.

**Withdrawing the performance claim in this description.** It stated CCE 21 is 5.8% faster than CCE 19. That was measured with

```
CRAY_CCE_LLD_ARGS=[-plugin-opt=-mattr=-mai-insts]
```

but this branch ships `-mai-insts` **and** `-disable-promote-alloca-to-vector`. The measurement therefore describes a configuration neither PR ships. #1694's own data suggests the shipped configuration is roughly 6% *slower* than the one I benchmarked, which is the expected direction — `-disable-promote-alloca-to-vector` disables an optimisation program-wide.

So the honest position is #1694's: **the correctness case stands on its own; the performance of the shipped configuration is unmeasured.** A clean CCE 19 vs shipped-CCE-21 comparison, one variable at a time and with repeats, has not been run.

That error is worth recording rather than just deleting. I compared the fastest CCE 21 configuration I had against CCE 19 and reported it as the port's speedup, without checking that it matched what the branch actually ships. The two flags went into `toolchain/modules` in separate commits and I benchmarked between them.

One item here is **not** in #1694 and should not be lost: `toolchain/mfc/build.py`, the fix for [#1689](https://github.com/MFlowCode/MFC/issues/1689) — `is_configured()` treating a bare `CMakeCache.txt` as built-and-installed. It is not CCE-specific and is worth taking on its own; I will open it separately rather than attaching it to the toolchain move. Over this port that one predicate destroyed six batch jobs, produced a false negative in a compiler-flag experiment, broke a control run, and silently voided two benchmark campaigns — every time presenting as a compiler or missing-package problem.

Branch `cce21-port` is left in place for reference.

PR #1698, opened 2026-07-29
Refactor: establish the thermodynamics interface

fahnab666:
Both sound-speed floor sites open-code the same `mixture_err` guard: `s_compute_speed_of_sound` in `m_variables_conversion.fpp` and the post-process derivation in `m_derived_variables.fpp`. This PR factors the test into one shared predicate, `f_validate_state`, a `[seq]` device routine with `cray_inline=True`, and rewires both sites to it. Each site keeps its own floor (`100*sgm_eps` simulation, `1.e-16_wp` post-process). 

The predicate is the old condition inverted, so no arithmetic changes. Diff: two source files, +21/-6. A second commit documents the interface contract in `equations.md` section 3.3.

This is the first step toward a multi-EOS thermodynamics interface (#1638): name the operations once, so a later EOS family adds a branch instead of editing call sites. No new physics, no new parameters, no runtime dispatch.

**Correctness.** Bit-identical by construction and by measurement, against a pristine clone of this repo at the branch point (782a1fdf). On GPU (LLNL Tuolumne, MI300A gfx942, CCE 19.0.0, OpenMP offload): restart files byte-identical in 16/16 comparisons across 10k/40k/160k cells (`2D_shockdroplet`, `mixture_err` on, HLLC, WENO5) plus a case-optimized build, and 32/32 scaling restarts at 1/2/4/8 APUs. 170/170 1D goldens pass. Zero NaN in every scanned restart.

**Performance.** `./mfc.sh bench` on Tuolumne, both trees run concurrently (one node each, 4 ranks x 1 APU), two passes. Grind delta, this branch vs upstream:

| case | Test 1 | Test 2|
|---|---|---|
| 5eq_rk3_weno3_hll | +0.0% | -2.0% |
| 5eq_rk3_weno3_hllc | +0.3% | +0.2% |
| 5eq_rk3_weno3_lf | -1.0% | +1.0% |
| hypo_hll | +0.9% | +0.7% |
| ibm | +0.9% | +1.4% |
| igr | -1.6% | -2.0% |
| viscous_weno5_sgb_acoustic | -0.9% | +0.7% |

Signs flip between passes, so there is no systematic trend; hllc, where the sound-speed path is hottest, sits at +0.3/+0.2%. The binary carries no separate device symbol for the predicate, so CCE inlines it on the offload path.

**Scaling.** Strong: 2.56M cells fixed on 1/2/4/8 APUs. Weak: 1.28M cells per APU. Grind (ns/gp/eq/RHS), lower is better: test in ns per grid point per equation per RHS, global-cell normalized:

| APUs | strong upstream | strong PR | weak upstream | weak PR |
|---|---|---|---|---|
| 1 | 0.9358 | 0.9334 | 1.2778 | 1.2856 |
| 2 | 0.6302 | 0.6371 | 0.6367 | 0.6257 |
| 4 | 0.4866 | 0.4863 | 0.3201 | 0.3190 |
| 8 | 0.4134 | 0.4188 | 0.1648 | 0.1657 |

Agreement within 1.7% at every point; weak efficiency at 8 APUs is 96.9% upstream, 97.0% PR. The strong-scaling solution is byte-identical across 1/2/4/8 ranks under a single hash, so the predicate introduces no rank dependence.

**Chemistry.** The chemistry adapter ran end to end on the machine: the build generated `m_thermochem.f90` with Pyrometheus 1.1.1 from a checksum-pinned Cantera 3.2.0 `h2o2.yaml`, byte-identical between the two trees. The reactive shocktube (400 cells, 18,681 steps) completed NaN-free on 1 and 8 APUs with output trees matching byte for byte between trees. Golden 5DCF300C passes on GPU in both.

**macOS CPU reference.** M4 Pro, gfortran, Open MPI 5.0.9, double precision. Restart files byte-identical to the baseline at all three grids. 625/627 goldens; the 2 failures are known load flakes that pass in isolation with and without the change. Debug build runs the shockdroplet case with bounds checks and asserts enabled, no trip. Single-rank simulation wall time, best of two:

| grid | branch | baseline | delta |
|---|---|---|---|
| 200x50 | 2.082 s | 2.120 s | -1.8% |
| 400x100 | 4.520 s | 4.625 s | -2.3% |
| 800x200 | 14.273 s | 14.210 s | +0.4% |

Strong scaling on 14 cores: 91/71/53% efficiency at 2/4/8 ranks; weak: 98/99/76%. The rolloff is single-socket bandwidth saturation, identical with and without the change.

Builds checked in both trees: double, single, chemistry, case-optimized, all clean with warning profiles identical line for line. `--mixed` fails identically in both trees inside untouched `m_derived_types.fpp` (CCE 19.0.0 rejects `real(kind=2)`); pre-existing, unrelated to this change.

### DIFF
--- a/src/common/m_variables_conversion.fpp
+++ b/src/common/m_variables_conversion.fpp
@@ -25,8 +25,8 @@ module m_variables_conversion
         & s_convert_mixture_to_mixture_variables, s_convert_species_to_mixture_variables, &
         & s_convert_species_to_mixture_variables_kernel, s_convert_conservative_to_primitive_variables, &
         & s_convert_primitive_to_conservative_variables, s_convert_primitive_to_flux_variables, s_compute_pressure, &
-        & s_compute_species_fraction, s_compute_speed_of_sound, s_compute_fast_magnetosonic_speed, &
-        & s_finalize_variables_conversion_module, gammas, gs_min, pi_infs, ps_inf, cvs, qvs, qvps
+        & s_compute_species_fraction, s_compute_speed_of_sound, s_compute_fast_magnetosonic_speed, f_internal_energy, &
+        & f_bulk_modulus, s_finalize_variables_conversion_module, gammas, gs_min, pi_infs, ps_inf, cvs, qvs, qvps
 
     real(wp), allocatable, dimension(:)   :: Gs_vc
     integer, allocatable, dimension(:)    :: bubrs_vc
@@ -88,6 +88,7 @@ contains
         real(wp)                                       :: T_guess
         integer                                        :: s  !< Generic loop iterator
         #:if not chemistry
+            ! Non-chemistry EOS branch (compile-time selected)
             ! Depending on model_eqns and bubbles_euler, the appropriate procedure for computing pressure is targeted by the
             ! procedure pointer
 
@@ -121,7 +122,7 @@ contains
                 pres = (energy - 0.5_wp*(mom**2._wp)/rho - pi_inf - qv - E_e)/gamma
             end if
         #:else
-            ! Reacting mixture pressure from temperature and species
+            ! Reacting mixture pressure from temperature and species (chemistry branch, via m_thermochem)
             Y_rs(:) = rhoYks(:)/rho
             e_Per_Kg = energy/rho
             Pdyn_Per_Kg = dyn_p/rho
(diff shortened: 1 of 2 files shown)


src/common/m_variables_conversion.fpp
```diff
@@ -1286,14 +1287,28 @@ contains
                 c = (H - 5.e-1*vel_sum - qv/rho)/gamma
             end if
 
-            if (mixture_err .and. c < 0._wp) then
-                c = 100._wp*sgm_eps
-            else
+            if (f_validate_state(c)) then
```
sbryngelson: can't have a function call here in the middle of a million hot loops.

PR #1659, opened 2026-07-19
JWL equation of state: core EOS + linear mixture closure + non-reactive IBM/EL coupling

fahnab666:
## Summary

Works toward #1638. This is a smaller, review-friendly slice of #1586 (which lands the
full detonation stack in one  about 10k-line PR). The capability here is one JWL products fluid
plus one ambient gas, the simplest possible mixture rule, and non-reactive immersed-boundary
and Euler-Lagrange bubble coupling. The diff against `master` is 3.8k lines.

## Why a second PR

#1586 delivers everything at once: the JWL EOS, a composition (heat-capacity) weighted
closure, a reactant/product energy offset, three reaction sources (program burn, afterburn,
JWL++ reactive), and IBM + Euler-Lagrange coupling. That is a lot to review in a single pass.
This PR is the **first tier**: the pieces can merge and trust on their own, split so each
commit is independently reviewable. Everything reactive and the composition-weighted closure
become follow-up PRs layered on this base.

## Core of this PR

- **JWL EOS** on the five-equation model (`model_eqns = 2`), selected per fluid with
  `fluid_pp(i)%eos = 2`, coexisting with an ideal-gas or stiffened-gas ambient.
- **Linear mass-fraction closure** (simpler than #1586's heat-capacity weight): every mixture
  coefficient is blended linearly in the products mass fraction `Y`, so pressure stays linear
  in internal energy. The `(rho, p, Y) -> e` inverse is closed form, the sound speed is the
  exact Grueneisen derivative, and the closure is endpoint-exact (pure JWL at `Y = 1`, ambient
  at `Y = 0`) and bit-identical to the ideal closure when the ambient `pi_inf` is zero. An
  init-time envelope scan aborts on any non-physical state.
- **Non-reactive IBM and Euler-Lagrange coupling**: the immersed-boundary ghost/fresh-cell
  energy rebuild and the bubble seed-pressure / Keller-Miksis sound-speed paths run through the
  JWL closure, all gated on `jwl_idx` so base behavior is bit-identical.

The base-without-JWL path is bit-identical throughout; incompatible features
(`wave_speeds = 2`, CBC boundaries, `alt_soundspeed`, elasticity, `igr`, `bubbles_euler`,
`mhd`, `chemistry`) are rejected at startup and in `./mfc.sh validate`.

## Commits

1. Core JWL EOS + linear mass-fraction closure + params, checkers, docs, and basic examples with goldens.
[...]

PR #1700, opened 2026-07-31
Add a per-fluid equation-of-state selector

fahnab666:
Adds `fluid_pp(i)%eos`, a stable enumeration for a fluid's equation of state, in place of the implicit backend selection. This is step 2 of the incremental EOS path tracked in #1638 ("add EOS selection and integrate the existing Pyrometheus path"). Note that step 1 — centralizing thermodynamic operations, attempted in #1663 — has not landed, so this selector arrives before the interface it will eventually dispatch through. That ordering is the reason nothing in `src/` reads `%eos` yet; see Scope below.

Only the two backends we already support are accepted:

- `stiffened_gas` (default)
- `ideal_gas_mixture` (chemistry builds, via Pyrometheus)

The enum holds exactly these two. Values for backends that do not exist yet are added alongside their backend rather than reserved up front, so the auto-generated "Valid values" hint never advertises a selector that `check_eos` will reject.

`ideal_gas_mixture` is the Pyrometheus mixture backend and is only valid in a chemistry build. On a non-chemistry build it is rejected, not silently downgraded. A plain ideal gas without chemistry is already covered by `stiffened_gas` with `pi_inf = 0`, which reduces the stiffened-gas law to the ideal-gas law, so it does not need its own enum value.

The default resolves from the compile-time `chemistry` flag, so existing cases do not change and the golden suite stays bit-identical (616/616, no goldens regenerated). The selector is validated at input time only, so nothing new enters the per-cell hot path. Case files use the readable names while the integer stays internal. The Fortran and Python checkers both reject unsupported values, `ideal_gas_mixture` on a non-chemistry build, and mixing EOS families within one run.

Both checkers cover every `fluid_pp` slot up to `num_fluids_max`, not just `num_fluids`: each slot is default-assigned and MPI-broadcast through the `_emit_fluid_pp` member loop, so a selector left on an unused slot still reaches the solver.

## Scope

Nothing in `src/` reads `%eos` yet, and the selector is fully determined by the `chemistry` flag, so there is exactly one legal value per build. That is deliberate — this PR establishes the interface and its validation, and the first backend to consume it lands next. Reviewers should weigh that tradeoff explicitly rather than treat it as an oversight.

## Tests

`toolchain/mfc/params_tests/test_eos_selector.py` covers the enum, the readable-name to integer resolution, and each `check_eos` branch. `test_fortran_and_python_enums_agree` compares the hand-written `eos_*` constants in `m_constants.fpp` against `_EOS_NAMES` — `generate_constants_fpp` skips compound registry keys, so nothing else forces the two to match.

---

# Handover notes

## Branch

```
repo    MFlowCode/MFC, PR #1700
head    fahnab666/mfc-nabid : feature/eos-selector-v2
```

External contribution (Fahad Nabid). Maintainer commits have been pushed directly to the contributor's branch — `maintainerCanModify` is true. If you push, use `--force-with-lease=<branch>:<sha>` with an explicit SHA; a bare `--force-with-lease` fails with "stale info" unless you have fetched the ref in the current clone.

Commits added on top of the contributor's work:

```
c9252fd  Trim the eos enum to the implemented backends and check every fluid_pp slot
b3efac1  Add tests for the eos selector
73e2150  Merge master into feature/eos-selector-v2   (conflict resolution, see below)
```

## Environment

```bash
./mfc.sh build -j 16
./mfc.sh test  -j 12
cd toolchain && python -m pytest mfc/params_tests -q
```

## Merge conflict with #1713 — resolved

After #1713 (remove the four-equation model) merged, this PR went `CONFLICTING` in exactly one file, `src/common/m_global_parameters_common.fpp`. Both branches edit the same `use m_constants` line: this PR adds `eos_stiffened_gas, eos_ideal_gas_mixture`, master removed `model_eqns_4eq`. Resolution keeps both. Verified after the merge: build clean, full suite 627 passed, 0 failed.

If master moves again, that same `use` line is the likely collision point.

## Plumbing that was verified, so it need not be re-derived

- `_r(f"{px}eos", INT, ...)` in `definitions.py` registers the parameter, so `_emit_fluid_pp` in `toolchain/mfc/params/generators/fortran_gen.py` auto-generates the MPI broadcast as `MPI_INTEGER` for all three targets. It reads the datatype from the registry, so no hand-editing of `m_mpi_proxy` is needed.
- `chemistry` is `logical, parameter :: chemistry = .${chemistry}$.` (`m_global_parameters_common.fpp`), i.e. a compile-time fypp parameter. That is what makes the pre-namelist-read default `merge(eos_ideal_gas_mixture, eos_stiffened_gas, chemistry)` correct and constant-folded.
- `Case.__init__` (`toolchain/mfc/case.py`) normalises readable names to integers **before** `CaseValidator` runs, so the validator sees ints. The name-resolution path works end to end.
- `fluid_pp` has no GPU declare, so widening `physical_parameters` is free.

## Design decisions taken during review

**The enum holds only implemented backends.** `mie_gruneisen`, `jwl` and `table` were removed. With them present, `choices` advertised all five values, so the auto-generated hint contradicted the error:

```
fluid_pp(1)%eos selects an equation of state that is not yet implemented; only ...
Valid values: 1 (stiffened-gas), 2 (ideal-gas mixture), 3 (Mie-Grueneisen), 4 (JWL), 5 (tabulated)
```

Note `test_names_are_wellformed_and_cover_choices` asserts `set(names.values()) == set(choices)`, so `names` and `choices` must shrink together.

**Membership is re-checked in `check_eos`.** The `choices` constraint is enforced by `validate_constraints`, a *different layer* from `CaseValidator` — relying on it lets an out-of-enum integer through. This was found by a test, not by reading.

**Both checkers loop to `num_fluids_max`.** Every `fluid_pp` slot is default-assigned and MPI-broadcast up to `num_fluids_max`, so a selector on an unused slot still reaches the solver. Validation is input-time only, so the wider bound costs nothing.

**Fortran/Python enum drift guard.** `generate_constants_fpp` silently skips compound registry keys (`fluid_pp(1)%eos` would produce the invalid identifier `fluid_pp(1)%eos_stiffened_gas`), so the `eos_*` constants are hand-written in `m_constants.fpp`. `test_eos_selector.py::test_fortran_and_python_enums_agree` compares them against `_EOS_NAMES` using `get_fortran_constants()`, which parses `m_constants.fpp`. Without that test nothing forces the two to agree.

## Open question for reviewers

Nothing in `src/` reads `%eos` yet, and the selector is fully determined by the compile-time `chemistry` flag, so there is exactly one legal value per build. That is deliberate — this establishes the interface and its validation — but it is a config knob that cannot be turned until a backend consumes it. The next PR in the series should be the thing that reads `%eos`, not more interface.

## Series context

Tracking issue #1638 records the plan from #1659:

1. Centralize thermodynamic operations *(not landed; #1663 was closed on performance grounds)*
2. **Add EOS selection and integrate the existing Pyrometheus path** — this PR
3. Add Mie-Gruneisen, then JWL
4. Add reaction models incrementally

Note this is step 2 landing before step 1, which is why nothing reads the selector yet.

Related: #1714 (`eos_state` refactor) gives the interface a dispatch seam; #1708 tracks the remaining EOS duplication. A constraint established while investigating #1714: `m_thermochem.f90` is generated for **every** build, but non-chemistry builds get a dummy `h2o2.yaml` mechanism (`toolchain/mfc/run/input.py`), and `num_species` from it sizes `rhoYks(1:num_species)` throughout. So `ideal_gas_mixture` is a whole-build mode, not a per-fluid peer backend — the fypp `chemistry` split must stay, and any future runtime dispatch covers only the non-chemistry backends.

---

# Purpose in the series

Tracking issue #1638 (JWL / multi-EOS). Plan from #1659:

1. Centralize thermodynamic operations — *not landed; #1663 closed on performance grounds*
2. **Add EOS selection and integrate the Pyrometheus path** — **this PR**
3. Add Mie-Gruneisen, then JWL
4. Add reaction models incrementally

This is step 2 arriving before step 1, which is why nothing in `src/` reads `%eos` yet. It establishes the vocabulary — a stable per-fluid enum, validated at input time — that later backends dispatch on. #1714 supplies the dispatch seam.

**Do not merge more interface after this.** The next PR in the series should be the first thing that actually reads `%eos`; otherwise the selector accumulates as a knob nobody can turn.

---

## Testing

## GPU

## Benchmarking

## GitHub mechanics


src/common/m_constants.fpp
```diff
@@ -114,6 +114,13 @@ module m_constants
     integer, parameter :: BC_NO_SLIP_WALL = -16
     integer, parameter :: BC_DIRICHLET = -17
 
+    ! fluid_pp(:)%eos selector; values must match _EOS_NAMES in toolchain/mfc/params/definitions.py
+    integer, parameter :: eos_stiffened_gas = 1
+    integer, parameter :: eos_ideal_gas_mixture = 2
+    integer, parameter :: eos_mie_gruneisen = 3
+    integer, parameter :: eos_jwl = 4
+    integer, parameter :: eos_table = 5
```
sbryngelson: These constants are hand-written directly above the `! AUTO-GENERATED from "names" in toolchain/mfc/params/definitions.py` include at lines 128-130, and `generate_constants_fpp` was patched to skip compound keys so it will not emit them.

The skip itself is necessary — `fluid_pp(1)%eos_stiffened_gas` is not a valid Fortran identifier, and the generator's `{param}_{name}` scheme would produce ten invalid copies. No objection to the mechanism.

The issue is that this creates an undocumented exception to auto-generation. `CLAUDE.md` says Fortran declarations are auto-generated and routes exceptions through `.claude/rules/common-pitfalls.md`, whose "Still manual" list enumerates the known ones (derived-type members, default assignments, `CASE_OPT_EXTRA_LINES`, multi-variable declarations, MPI broadcast residue). Compound-key enum constants are not among them.

Please add this case to the "Still manual" list in `.claude/rules/common-pitfalls.md`, so the next person adding a struct-member enum knows the generator will silently skip it. Worth also amending the comment at line 128 to note that not all enumerated constants below are generated.

toolchain/mfc/case_validator.py
```diff
@@ -747,6 +757,27 @@ def check_stiffened_eos(self):
...
 
+    def check_eos(self):
+        """Restricts the per-fluid EOS selector to the currently supported adapters"""
+        num_fluids = self.get("num_fluids")
+        chemistry = self.get("chemistry", "F") == "T"
+
+        if num_fluids is None:
+            return
+
+        eos_stiffened_gas, eos_ideal_gas_mixture = 1, 2
```
sbryngelson: This makes a third hand-synced copy of the mapping, after `eos_*` in `m_constants.fpp` and `_EOS_NAMES` in `definitions.py`. The comment this PR adds at `m_constants.fpp:117` states the invariant explicitly ("values must match `_EOS_NAMES`"), but nothing forces these literals to follow.

This file already imports `CONSTRAINTS` (line 20) and reads the `names` dict this way for `recon_type` at line 323. Prefer deriving the two values from `CONSTRAINTS["fluid_pp(1)%eos"]["names"]` rather than restating them.

Acknowledged that hardcoding enum integers is the prevailing style elsewhere in this file, so this is a preference, not a rule violation — but here the values carry a documented cross-file invariant that the other cases do not.

toolchain/mfc/case_validator.py
```diff
@@ -747,6 +757,27 @@ def check_stiffened_eos(self):
...
+
+        for i in range(1, num_fluids + 1):
```
sbryngelson: This loop stops at `num_fluids`, but the directly analogous `check_stiffened_eos` just above (line 742) deliberately covers one extra fluid-property slot when `bubbles_euler` is set:

```python
bub_fac = 1 if (bubbles_euler) else 0
for i in range(1, num_fluids + 1 + bub_fac):
```

The default assignment loops to `num_fluids_max`, so `fluid_pp(num_fluids+1)%eos` is settable and reaches the solver unvalidated. The Fortran `s_check_eos` has the same gap (`do i = 1, num_fluids`).

No functional consequence today since nothing reads `%eos`, and `chemistry` is mutually exclusive with `bubbles_euler` — but please make the two checks consistent now rather than leaving the asymmetry for whoever wires up the first real backend.

toolchain/mfc/params/definitions.py
```diff
@@ -887,6 +895,7 @@ def _load():
         px = f"fluid_pp({f})%"
         for a, sym in [("gamma", r"\f$\gamma_k\f$"), ("pi_inf", r"\f$\pi_{\infty,k}\f$"), ("cv", r"\f$c_{v,k}\f$"), ("qv", r"\f$q_{v,k}\f$"), ("qvp", r"\f$q'_{v,k}\f$")]:
             _r(f"{px}{a}", REAL, math=sym)
+        _r(f"{px}eos", INT, math=r"\f$\mathrm{EOS}_k\f$")
```
sbryngelson, line 899: Registering `eos` here leaves the comment three lines up stale: `# Members present in physical_parameters: gamma, pi_inf, Re, cv, qv, qvp, G.` no longer lists every registered member.

That comment came from #1553 ("Fix three pre-existing multi-rank broadcast defects; deregister dead parameters") specifically as a drift guard between the Python registry and the Fortran type — the next line warns that registering a member absent from the type crashes the namelist read. Since this PR adds `eos` to `physical_parameters`, please add it to the list.

Noting fairly that the comment was already incomplete on master (it omits `non_newtonian`, `K`, `nn`, `tau0`, `hb_m`, `mu_min`, `mu_max`, `mu_bulk`), so this is partly pre-existing. Fixing just your own addition is fine; fixing the rest is welcome but not required.

sbryngelson (dismissed):
Thanks for this. The plumbing checks out — `_r(f"{px}eos", INT, ...)` registration means the MPI broadcast is auto-generated correctly as `MPI_INTEGER` for all three targets, the namelist binds `fluid_pp` as a derived type so `fluid_pp(1)%eos` resolves, and the compile-time `chemistry` parameter makes the pre-namelist-read default correct. The Fortran and Python checkers agree.

Four things to address before merge, mostly around the enum having several hand-synced copies. Details inline.

Separately, worth noting for context rather than as a change request: nothing in `src/` currently reads `%eos`, and both checkers leave exactly one legal value per build since it is fully determined by `chemistry`. That is defensible as step-2 groundwork, but it is the part I would most want justified given the guidance against config knobs with one correct value.

sbryngelson:
### Code review

Found 1 issue:

1. Out-of-bounds access in `s_check_eos`: the loop bound `num_fluids + merge(1, 0, bubbles_euler)` reaches `fluid_pp(11)` when `num_fluids = 10` and `bubbles_euler = T`, but `fluid_pp` is `dimension(num_fluids_max)` with `num_fluids_max = 10`. `num_fluids` is constrained to `{"min": 1, "max": NF}` with `NF = 10` (`toolchain/mfc/params/definitions.py`), and `check_bubbles_euler` only forces `num_fluids = 1` for `model_eqns = 4` — so `num_fluids = 10, bubbles_euler = T, model_eqns = 2` passes every validator and then indexes one past the end. Debug builds abort on bounds check; release builds read adjacent memory.

This came in with `ff17a039`, mirroring the `bub_fac` idiom from `check_stiffened_eos`. That idiom is safe in Python only because `self.get(f"fluid_pp({i})%eos")` returns `None` for an out-of-range key and the loop body skips it; Fortran dereferences a fixed-size array instead. No other Fortran code indexes `fluid_pp` past `num_fluids`. Clamping the bound (or dropping the `merge` term, since the defaults loop already initializes all `num_fluids_max` slots) resolves it.

https://github.com/MFlowCode/MFC/blob/1a7222f6546c02abd6951912ba6f1602e51495ab/src/common/m_checker_common.fpp#L41-L43

https://github.com/MFlowCode/MFC/blob/1a7222f6546c02abd6951912ba6f1602e51495ab/src/common/m_constants.fpp#L23-L25

sbryngelson (dismissed):

sbryngelson:
I don't understand why we can't just do exactly what is done now, but use ideal gas as an option (leaving chemistry alone) for when \Pi_\infty = 0?

fahnab666:
> I don't understand why we can't just do exactly what is done now, but use ideal gas as an option (leaving chemistry alone) for when \Pi_\infty = 0?

@sbryngelson 

That's a valid approach. I kept the change small on purpose: eos defaults to stiffened_gas (eos=1) when it isn't set, and that path already covers both stiffened gas and ideal gas. With pi_inf = 0 the stiffened-gas EOS reduces to an ideal gas, with no chemistry involved. eos=2 (ideal_gas_mixture) is the Pyrometheus multi-species mixture specifically, which is the only reason it is tied to chemistry.

Happy to do it your way: require an explicit eos in the case file instead of defaulting it, and leave chemistry as its own independent switch. The ideal-gas case then stays exactly as you describe, stiffened_gas with pi_inf = 0, decoupled from chemistry.

sbryngelson:
I pushed three follow-ups to this branch directly (`c9252fd`, `b3efac1`) rather than round-tripping them — all small, all in response to the second read-through. Happy to revert any of them if you disagree.

**Reserved enum values removed.** `choices` listed all five values, so the generated hint contradicted the error:

```
fluid_pp(1)%eos selects an equation of state that is not yet implemented;
only 'stiffened_gas' and 'ideal_gas_mixture' are available
Valid values: 1 (stiffened-gas), 2 (ideal-gas mixture), 3 (Mie-Grueneisen), 4 (JWL), 5 (tabulated)
```

A user reads the second line, sets `jwl`, and gets rejected again. Dropping `mie_gruneisen`/`jwl`/`table` fixes that and removes three of the five hand-synced constants in `m_constants.fpp`. Add each value when its backend lands.

**Both checkers now cover every slot.** They stopped at `num_fluids + bub_fac`, but `_emit_fluid_pp` broadcasts `i = 1, num_fluids_max` and the defaults are assigned over the same range, so `fluid_pp(3)%eos` with `num_fluids = 2` was accepted unvalidated. Confirmed against the validator before the change. No consequence today, but it becomes a silent wrong-EOS path for whoever wires the first backend, which is the whole point of this PR.

**Tests added.** `test_eos_selector.py` — the enum, name resolution, and each `check_eos` branch. The one worth your attention is `test_fortran_and_python_enums_agree`: `generate_constants_fpp` skips compound keys, so the `eos_*` constants are hand-written and only a comment asserted they match `_EOS_NAMES`. `get_fortran_constants()` already parses `m_constants.fpp`, so the guard is a direct comparison.

Writing those tests caught a mistake in my own first pass: I had dropped the membership check on the assumption that the `choices` constraint would cover out-of-enum integers. It does not — `choices` is enforced by `validate_constraints`, a different layer from `CaseValidator` — so `eos: 99` passed. The explicit check is restored with a comment saying why it is not redundant.

Verification: `params_tests` 181 passed; `ruff check` / `ruff format --check` clean; `ffmt --check src/` clean. CI will confirm the rest.

On scope: I have left the "one legal value per build" point as a note in the description rather than a change request. It is a real cost — a case parameter that cannot be set to anything other than what it already is — and it is only worth paying if the first backend follows closely. I am fine merging on that basis, but the next PR in the series should be the thing that reads `%eos`, not more interface.

fahnab666:
@sbryngelson Do you have any recommendation of fix. need to be done before it merge?

sbryngelson:
i will take care of it and you can look at my commits after

sbryngelson:
this pr depends on #1714 merging, which hasn't happened yet (it's too slow). draft until then.

sbryngelson:
closing due to a complete refactor in #1762

Issue #1708, opened 2026-08-08
EOS algebra is duplicated across call sites; needs a single source of truth


sbryngelson:
## Summary

The stiffened-gas EOS algebra has canonical definitions in `src/common/m_variables_conversion.fpp` — `s_compute_pressure`, `s_compute_internal_energy`, `s_compute_speed_of_sound` — but several call sites open-code the same expressions instead of calling them. Two have already drifted into real bugs (#1706, #1707). This issue tracks making the algebra DRY, which is prerequisite groundwork for the multi-EOS interface in #1638: a second EOS cannot be introduced safely while the first one has several independent definitions.

## Pressure inversion

Canonical, `m_variables_conversion.fpp:71`:

```fortran
pres = (energy - dyn_p - pi_inf - qv)/gamma
```

Open-coded copies:

| Site | Expression | Status |
|---|---|---|
| `simulation/m_pressure_relaxation.fpp:311` | `(E - dyn_pres - qv_mix - pi_inf)/gamma` | correct; operand order differs |
| `simulation/m_data_output.fpp:1558` | `((E - KE)/(1 - alf) - pi_inf - qv)/gamma` | correct; duplicates the bubble branch |
| `simulation/m_data_output.fpp:1621` | `((E - KE)/(1 - alf) - pi_inf - qv)/gamma` | correct; duplicates the bubble branch |
| `simulation/m_bubbles_EL.fpp:374` | `(E - dynP - pi_inf)/gamma` | **wrong — omits `qv`** (#1706) |

## Specific total enthalpy at sound-speed call sites

`s_compute_speed_of_sound` requires an `H` that includes `qv` (it subtracts `qv/rho` internally). The 27 call sites split into two conventions:

- Solver paths derive it from conserved energy — `H_L = (E_L + pres_L)/rho_L` — which is correct and cannot drift.
- Four diagnostic sites open-code the closed form instead. `post_process/m_data_output.fpp:1289` includes `qv` and is correct; `simulation/m_data_output.fpp:1297,1379,1438` omit it and are **wrong** (#1707).

A fifth site, `s_compute_cson_from_pinf` in `simulation/m_bubbles_EL.fpp:844-846`,
reconstructs the energy rather than reading it:

```fortran
E    = gamma*pinf + pi_inf + 0.5_wp*rhol*dot_product(vel, vel)   ! no qv
H    = (E + pinf)/rhol
[...]

PR #1709, opened 2026-08-08
Fix missing qv term in the Lagrange bubble initial pressure

### LINKED ISSUE #1706: bug: Lagrange bubble initial gas pressure omits the qv term from the stiffened-gas inversion
## Summary

`s_locate_bubbles`/particle initialization in `src/simulation/m_bubbles_EL.fpp` open-codes the stiffened-gas pressure inversion and omits the `qv` (heat of formation) term. The resulting liquid pressure seeds the initial Lagrangian bubble gas pressure, so the error is in the physics, not in output.

## Location

`src/simulation/m_bubbles_EL.fpp:374`

```fortran
call s_convert_to_mixture_variables(q_cons_vf, cell(1), cell(2), cell(3), rhol, gamma, pi_inf, qv, Re)
dynP = 0._wp
do i = 1, num_dims
    dynP = dynP + 0.5_wp*q_cons_vf(eqn_idx%cont%end + i)%sf(cell(1), cell(2), cell(3))**2/rhol
end do
pliq = (q_cons_vf(eqn_idx%E)%sf(cell(1), cell(2), cell(3)) - dynP - pi_inf)/gamma
```

The canonical inversion in `s_compute_pressure` (`src/common/m_variables_conversion.fpp:71`) is

```fortran
pres = (energy - dyn_p - pi_inf - qv)/gamma
```

`qv` is missing from the `m_bubbles_EL` expression.

## Evidence that this is accidental

`qv` is an output argument of the `s_convert_to_mixture_variables` call on the line above, and it is never read anywhere else in that scope. The value is computed and discarded.

## Impact

`pliq` is too large by `qv/gamma`. It flows directly into the initial bubble gas pressure:

```fortran
gas_p(bub_id, 1) = pliq + 2._wp*(1._wp/Web)/bub_R0(bub_id)
```

so every Lagrangian bubble starts from a wrong internal pressure whenever `qv /= 0`.

Dormant when `qv = 0`, which is the default. It triggers for any case that sets `fluid_pp(i)%qv`, i.e. the phase-change and reactive-burn configurations — `examples/2D_phasechange_bubble`, `examples/3D_phasechange_bubble`, `examples/1D_exp_tube_phasechange`, `examples/2D_shocktube_phasechange`, `examples/1D_reactive_burn` all set it.

## Suggested fix

Route the site through `s_compute_pressure` rather than adding `- qv` to the open-coded expression, so the definition cannot drift again. See the tracking issue for the broader deduplication.

sbryngelson:
Fixes #1706.

## The defect

`src/simulation/m_bubbles_EL.fpp` open-codes the stiffened-gas pressure inversion and drops the `qv` (heat of formation) term:

```fortran
call s_convert_to_mixture_variables(q_cons_vf, cell(1), cell(2), cell(3), rhol, gamma, pi_inf, qv, Re)
...
pliq = (q_cons_vf(eqn_idx%E)%sf(cell(1), cell(2), cell(3)) - dynP - pi_inf)/gamma
```

The canonical inversion, `s_compute_pressure` in `m_variables_conversion.fpp:71`, is

```fortran
pres = (energy - dyn_p - pi_inf - qv)/gamma
```

The omission looks accidental rather than a deliberately different quantity: `qv` is an output argument of the `s_convert_to_mixture_variables` call on the line above, and it is never read anywhere else in that scope. It is computed and discarded.

## Impact

`pliq` is too large by `qv/gamma`, and it seeds the initial bubble gas pressure:

```fortran
gas_p(bub_id, 1) = pliq + 2._wp*(1._wp/Web)/bub_R0(bub_id)
```

So every Lagrangian bubble starts from the wrong internal pressure. This is physics, not diagnostic output.

Dormant when `qv = 0`, which is the default. It triggers for any case setting `fluid_pp(i)%qv` — the phase-change and reactive-burn configurations.

## Why it survived

No test exercised the combination. Exactly one golden case sets `bubbles_lagrange`, and it leaves `qv` at zero; the cases that set `qv` are phase-change and reactive-burn, none of which use Lagrangian bubbles. The suite could not have caught this.

## Test

Adds a 2D one-way-coupled Lagrange bubble case with `fluid_pp(1)%qv = 0.01`, restricted to a single configuration so it contributes one golden (`F428FDC0`).

The golden captures `beta`, the Lagrangian void fraction, alongside the conservative variables. `beta` follows the bubble radius, which is driven by the initial gas pressure, so the error is observable even under one-way coupling.

## Verification

```
new case, fix reverted:   1 failed  (tolerance mismatch)
new case, fix applied:    1 passed
all bubble tests:        64 passed, 0 failed
```

Built with GNU 15.2.0, MPI, on macOS. The negative result is the one that matters — without it, a green suite would prove nothing here, since the pre-existing cases are all `qv = 0`.

No existing golden moved: the change is confined to a path only reachable with `bubbles_lagrange`, and no existing lag case has nonzero `qv`.

## Behaviour change worth flagging

The correction shifts `pliq` by `-qv/gamma`, and `gamma > 0` always, so the direction follows the sign of `qv`. For `qv > 0` the initial bubble pressure drops, which brings two existing guards in the same routine within reach:

```fortran
if (gas_mg(bub_id) <= 0._wp) call s_mpi_abort("The initial mass of gas inside the bubble is negative. ...")
if (pv*(massflag) > gas_p(bub_id, 1)) call s_mpi_abort("Lagrange bubble initially located in a region with pressure below the vapor pressure.")
```

A case with `qv > 0` and Lagrangian bubbles that previously ran can now abort at startup. That is the correct outcome — it was running on an inflated bubble pressure that masked a setup the model rejects — but it converts a silently wrong answer into a hard stop, so it is a user-visible change rather than a pure numerical shift. For `qv < 0` (as in the phase-change examples, which use `qv = -1.167e6` for fluid 1) the pressure rises and the guards become less reachable.

`gas_p` also feeds the initial gas mass and the bubble natural frequency in the same routine, so both are corrected by the same change.

## Adjacent issue found while reviewing

`s_add_bubbles` also assigns the module-level global `pref` from the last-added bubble's `gas_p` (line 383). `pref` is a documented user-settable case parameter that the four-equation EOS branch reads. This PR changes the value it gets overwritten with, but the overwrite itself is pre-existing and looks wrong independently. Filed as #1710 rather than addressed here.

## Context

First of three PRs against #1708. The root cause is that the EOS algebra has several independent definitions, and `s_compute_pressure` is awkward enough to call — it demands a `num_species` array and an `intent(inout)` temperature — that call sites open-code the expression instead. This PR fixes only the bug; deduplication follows separately so that any behavior change lands in a PR whose purpose is to change behavior.

---

# Handover notes

Context for anyone picking this up on another machine.

## Branch and commits

```
repo    MFlowCode/MFC
branch  fix/lagrange-bubble-qv     (base: master)
  174775f  Fix missing qv term in the Lagrange bubble initial pressure
  9191c9a  Add golden for the qv_nonzero Lagrange bubble regression case
```

Two commits deliberately: the golden in the second is generated from a **clean checkout of the first**, so `golden-metadata.txt` records a real committed SHA rather than a dirty tree. Copilot flagged the dirty-tree provenance on the original single-commit version; that is why it is split. Preserve this if you amend — regenerate the golden only from a committed, clean tree.

## Environment

```bash
./mfc.sh build -j 16          # builds pre_process, simulation, post_process
./mfc.sh test  -j 12          # full golden suite (~5 min on 18 cores)
./mfc.sh test --only F428FDC0 # this PR's case alone
```

Verified with GNU 15.2.0 (gfortran), MPI on, no GPU, macOS arm64. A first `./mfc.sh build` bootstraps a Python venv and builds hdf5/silo; budget ~20 min.

Committing runs a pre-commit hook that executes the full CI lint-gate (7 checks). Use `--no-verify` only when you have already run it.

## The exact defect

`src/simulation/m_bubbles_EL.fpp`, in `s_add_bubbles`:

```fortran
call s_convert_to_mixture_variables(q_cons_vf, cell(1), cell(2), cell(3), rhol, gamma, pi_inf, qv, Re)
...
pliq = (q_cons_vf(eqn_idx%E)%sf(...) - dynP - pi_inf)/gamma      ! qv missing
```

Canonical inversion is `s_compute_pressure` in `src/common/m_variables_conversion.fpp`:
`pres = (energy - dyn_p - pi_inf - qv)/gamma`.

`qv` is an output argument of the call one line above and is read nowhere else in that scope — computed and discarded. That is the evidence the omission is accidental rather than a deliberately different quantity.

## Why `pliq` must be the full thermodynamic pressure

Settled by internal consistency, not by the Maeda paper (which I did not read — it is paywalled JCP; `docs/documentation/equations.md` section 6.2 documents the model):

`s_get_pinf` in the same file — "Compute the bubble driving pressure p_inf", the Maeda & Colonius (2018) subgrid closure — interpolates `q_prim_vf(eqn_idx%E)%sf`. That slot is filled at `m_variables_conversion.fpp` by `s_compute_pressure` **with `qv_K` passed**. So at every step after t=0 the pressure driving the bubble includes `qv`. Initialising from an inversion that omits it left each bubble out of equilibrium with its own driving pressure by `qv/gamma`.

Only `s_add_bubbles` needs the change: the restart path (`s_restart_bubbles`) reads `gas_p` straight from the restart file rather than recomputing it.

## Test case: why it is shaped the way it is

Added in `toolchain/mfc/test/cases.py`, inside `alter_lag_bubbles`, guarded to `ndims == 2 and couplingMethod == 1 and adap_dt == "F"` so it contributes exactly one golden (`F428FDC0`).

Two constraints that are easy to get wrong:

1. **The golden must capture `beta`.** Under one-way coupling the bubbles do not feed back into the Eulerian field, so the conservative variables alone would not move. It works because `beta`, the Lagrangian void fraction, follows bubble radius and therefore the initial gas pressure. If you restructure the case, confirm `beta` is still in `tests/F428FDC0/golden.txt`.
2. **Do not add `qv` to the existing lag-bubble stack.** That would churn every existing lag golden. A new label is what keeps it to one.

The case label feeds the golden UUID — `crc32(sha1(str(trace)))` at `toolchain/mfc/test/case.py`. Renaming the label changes the directory name. Avoid `!` in labels (history expansion in interactive bash); `qv_nonzero` replaced an earlier `qv!=0` for that reason.

## How the fix was verified, and a trap

```
F428FDC0 with fix reverted:  1 failed (tolerance mismatch)
F428FDC0 with fix applied:   1 passed
all bubble tests:           64 passed, 0 failed
full suite:                627 passed, 0 failed
```

**Trap:** once the fix is committed, `git stash push -- src/simulation/m_bubbles_EL.fpp` has nothing to stash and silently leaves the fixed binary in place — the "reverted" run then reports a pass that looks identical to a real one. Revert with `git checkout master -- src/simulation/m_bubbles_EL.fpp` instead, rebuild, run, then `git checkout HEAD -- <file>` to restore.

This matters because no pre-existing golden covers the path: only one case sets `bubbles_lagrange` and it has `qv = 0`, while the `qv /= 0` cases are phase-change and reactive-burn without Lagrangian bubbles. A green suite proves nothing here without the negative check.

## Related work

- #1706 — the issue this closes.
- #1708 — tracking issue for the EOS duplication this is one symptom of. Nine hand-written copies of two expressions; this is one of the four wrong ones.
- #1710 — `pref` overwrite in the same routine. Already resolved: #1713 (merged) deleted the global.
- #1714 — `eos_state` refactor; covers the enthalpy half of #1708.
- The pressure-inversion half of #1708 is still open: `m_pressure_relaxation.fpp`, `m_data_output.fpp` (x2), and this file all open-code the same inversion.

---

# Purpose in the series

A correctness fix, and the first concrete evidence for #1708.

The EOS algebra has nine hand-written copies of two expressions. Four are wrong. This is one of them, and the only one that affects **physics** rather than diagnostics — the others are output paths. It stands alone as a bug fix, but the reason it exists is structural: `s_compute_pressure` is awkward enough to call (it demands a `num_species` array and an `intent(inout)` temperature) that call sites open-code the one line instead, and then drift.

Fixing the instance without fixing the pressure is deliberate — the deduplication is tracked separately so any behaviour change lands in a PR whose stated purpose is to change behaviour.

---

## Testing

## GPU

## Benchmarking

## GitHub mechanics


sbryngelson:
Both Copilot points were valid and are addressed in the force-push (`b3431b4` → `9191c9a`). Details, since I checked them rather than just applying them.

**Label `qv!=0`.** Correct, and worse than cosmetic. The case UUID is `crc32(sha1(str(trace)))` (`toolchain/mfc/test/case.py:149`), so the label is not merely display text — it determines the golden directory name. Separately, `!` is history expansion in interactive bash, so `./mfc.sh test --only "...qv!=0..."` misbehaves when pasted into a shell. Checking the conventions: `=` in labels has ample precedent (`riemann_solver=1`, `weno_order=5`, `adap_dt=T`), but `qv!=0` was the only label in the entire file containing `!`. Renamed to `qv_nonzero`, which changed the UUID `43EA05B4` → `F428FDC0`; the old golden directory is removed and the new one added.

**Dirty working tree.** Also correct, and the recorded SHA `0c9a1d43` was not even the commit that ended up in the PR. Restructured into two commits so the provenance is real: the source fix and test case land first, then the golden is generated against that clean checkout, then committed. The metadata now reads

```
Git:  174775fb74510109ef017ea97c98fc9cee1eb5ca on fix/lagrange-bubble-qv (clean)
```

which is the parent commit in this PR.

**One thing the restructure caught.** After committing the fix, my original verification method silently stopped working — `git stash push -- src/simulation/m_bubbles_EL.fpp` had nothing to stash, so the "fix reverted" run was actually testing the fixed binary and reported a pass. Redone with `git checkout master -- <file>`, which reverts the source properly:

```
F428FDC0, fix reverted:   1 failed  (tolerance mismatch)
F428FDC0, fix applied:    1 passed
all bubble tests:        64 passed, 0 failed
```

Worth stating explicitly because a regression test that cannot fail is worse than no test, and the false pass looked identical to a real one.

**Also updated the description** with two things I had missed on the first pass: the fix can newly trigger the vapour-pressure and negative-gas-mass aborts for `qv > 0` cases that previously ran, and `s_add_bubbles` overwrites the user-supplied global `pref` from the last bubble's pressure — pre-existing, now filed as #1710.

sbryngelson:
The automated review flagged the `qv` term as dimensionally inconsistent, arguing it should be the density-weighted `rhol*qv` rather than bare `qv`. I checked it rather than applying it, and the finding is incorrect — the existing form is right.

The argument assumes `qv` here is a *specific* energy (energy/mass). It is not. `qv` is an output of `s_convert_to_mixture_variables`, which returns it already density-weighted (`src/common/m_variables_conversion.fpp:246`):

```fortran
qv_K = qv_K + alpha_rho_K(i)*qvs(i)
```

`qvs(i)` is the specific heat of formation and `alpha_rho_K(i)` is a partial density, so `qv_K` is a volumetric energy — the same units as `E`, `dynP`, and `pi_inf`. Subtracting it directly is dimensionally correct; multiplying by `rhol` would introduce a spurious extra density factor.

This is also exactly what the canonical inversion does. `s_compute_pressure` in the same file (line 99) is:

```fortran
pres = (energy - dyn_p - pi_inf - qv)/gamma
```

which is term-for-term what this patch now computes, and matching it is the stated point of the change. Had the suggested `rhol*qv` been applied, this call site would have disagreed with `s_compute_pressure` for every case with `qv /= 0` — reintroducing the same class of defect in the opposite direction, and in a regime with no golden coverage before this PR.

The `rhol*qv` form would only be correct if `qv` were specific, which is true of `qvs(i)` but not of the aggregated `qv` returned here.

sbryngelson:
Cross-platform check on the new golden. `F428FDC0` had only ever been generated and verified on one platform, and `.claude/rules/common-pitfalls.md` records the Lagrange-bubble / IB paths as where AMD-only failures have appeared before (a platform-only NaN being the signature), so a second compiler and vendor seemed worth the 14 minutes.

Ran the full Lagrange family — all 28 cases, including `F428FDC0` (`2D -> Lagrange Bubbles -> One-way Coupling -> qv_nonzero`) — on an MI250X (gfx90a) node with the latest AFAR drop (`therock-afar-23.2.1-gfx90a-7.13.0`, `amdflang`), OpenMP target offload, `OMP_TARGET_OFFLOAD=MANDATORY` so a silent host fallback would abort rather than quietly pass:

```
✓ ALL TESTS PASSED
   28 passed · 0 failed · 640 skipped · 668 total · 14m 7s
```

So the golden reproduces within tolerance on AMD/amdflang as well as the platform it was generated on, and the `qv` fix does not disturb any of the other Lagrange configurations.

Issue #1707, opened 2026-08-08
bug: simulation probe sound speed omits qv from enthalpy, disagreeing with post-process


sbryngelson:
## Summary

Three `s_compute_speed_of_sound` call sites in `src/simulation/m_data_output.fpp` build the specific total enthalpy `H` without the `qv` term while still passing the real `qv` as the final argument. Because the routine subtracts `qv/rho` internally, the sound speed comes out too low whenever `qv /= 0`. The equivalent post-process path builds `H` correctly, so simulation and post-process report different sound speeds for identical state.

## The contract

`s_compute_speed_of_sound` (`src/common/m_variables_conversion.fpp:1180`) computes, in the default stiffened-gas branch (line 1227):

```fortran
c = (H - 5.e-1*vel_sum - qv/rho)/gamma
```

Since `E = gamma*p + pi_inf + qv + rho*KE`, the specific total enthalpy `H = (E + p)/rho` includes `qv`. The internal `- qv/rho` then cancels it, leaving the correct stiffened-gas result `c^2 = ((gamma+1)*p + pi_inf)/(rho*gamma)`. **`H` must therefore include `qv`.**

Every solver call site satisfies this implicitly by deriving `H` from the conserved energy, e.g. `src/simulation/m_riemann_solver_hll.fpp:255`:

```fortran
H_L = (E_L + pres_L)/rho_L
```

Those paths are correct and unaffected.

## The defect

`src/simulation/m_data_output.fpp`, lines 1297, 1379, 1438:

```fortran
call s_compute_speed_of_sound(pres, rho, gamma, pi_inf, ((gamma + 1._wp)*pres + pi_inf)/rho, alpha, 0._wp, &
                              & 0._wp, c, qv)
```

`H` is open-coded as `((gamma + 1)*pres + pi_inf)/rho` — no `qv` — but the real `qv` is still passed. The internal subtraction then has nothing to cancel against:

```
c^2 = ((gamma+1)*p + pi_inf - qv)/(rho*gamma)      instead of
c^2 = ((gamma+1)*p + pi_inf)/(rho*gamma)
```

`src/post_process/m_data_output.fpp:1289` gets it right:

```fortran
H = ((gamma + 1._wp)*pres + pi_inf + qv)/rho
call s_compute_speed_of_sound(pres, rho, gamma, pi_inf, H, adv, 0._wp, 0._wp, c, qv)
```

## Impact

[...]

PR #1712, opened 2026-08-08
Keep every column of probe output in golden packs

### LINKED ISSUE #1711: Golden packer keeps only the last column of probe output, hiding probe regressions
## Summary

The golden packer treats every `.dat` file under `D/` as a spatial field of the form `<x> [<y> <z>] <value>` and keeps only the last column of each row. Probe output is not a spatial field — it is a multi-column time series — so all but one of its columns are silently discarded. Probe regressions are therefore invisible to the test suite.

## Mechanism

`toolchain/mfc/packer/pack.py:139-144`:

```python
# Every line is <x> <y> <z> <value> (<y> and <z> are optional). So the
# number of dimensions is the number of doubles in the first line minus 1.
ndims = len(_extract_doubles(content.split("\n", 1)[0])) - 1
# We discard all <x> <y> <z> values and only keep the <value> ones.
doubles = _extract_doubles(content)[ndims :: ndims + 1]
```

`lag_bubble` files are already special-cased above this, which is the same underlying problem recognised once.

## Effect on probe files

`s_write_probe_files` emits different column sets by configuration:

| Configuration | Columns written | Kept by the packer |
|---|---|---|
| 1D, general (`m_data_output.fpp:1519`) | `nondim_time, rho, vel(1), pres` | `pres` only |
| hypoelastic (line 1515) | `nondim_time, rho, vel(1), vel(2), pres, tau_e(1..3)` | `tau_e(3)` only |
| 3D (line 1524) | `nondim_time, rho, vel(1), vel(2), vel(3), pres, gamma, pi_inf, qv, c, accel` | `accel` only |

So in 3D the golden validates the acceleration magnitude and nothing else — density, velocity, pressure, and the sound speed are all dropped. Confirmed against the committed goldens: `tests/5CAA4E68/golden.txt` and `tests/FBB296DA/golden.txt` each store 50 values for a 50-row `probe1_prim.dat`, i.e. exactly one column.

## Why it matters now

#1707 is a defect in the probe sound speed. It cannot be given a regression test through the normal golden path, because `c` is never the last column in any configuration. A case can be constructed that exercises the defective code, produce visibly wrong output, and still pass its golden.

More generally, any probe regression that does not happen to land in the last column is undetectable.

## Suggested fix

Special-case probe files the way `lag_bubble` files already are, keeping every column (skipping any header). This will change the content of the three existing probe goldens — `tests/5CAA4E68`, `tests/AE9A7D73`, `tests/FBB296DA` — which will need regenerating. That regeneration is the point: it is coverage those cases should have had.

Worth checking whether other non-field `.dat` outputs under `D/` are mis-parsed the same way.

sbryngelson:
Fixes #1711.

## The defect

`toolchain/mfc/packer/pack.py` treats every `.dat` file under `D/` as a spatial field of the form `<x> [<y> <z>] <value>`, infers dimensionality from the first line, and keeps only the last column of each row:

```python
ndims   = len(_extract_doubles(content.split("\n", 1)[0])) - 1
doubles = _extract_doubles(content)[ndims :: ndims + 1]
```

Probe output is not a spatial field. It is a multi-column time series whose columns are distinct physical quantities, and the set varies by configuration:

| Configuration | Columns written |
|---|---|
| 1D, general (`m_data_output.fpp:1519`) | `nondim_time, rho, vel(1), pres` |
| bubbles (line 1490) | `nondim_time, rho, vel(1), pres, alf, R, Rdot, nR, nRdot` |
| hypoelastic (line 1515) | `nondim_time, rho, vel(1), vel(2), pres, tau_e(1..3)` |
| 3D (line 1524) | `nondim_time, rho, vel(1..3), pres, gamma, pi_inf, qv, c, accel` |

Under the field interpretation only the final column survived. In 3D that means the golden validated the acceleration magnitude and discarded density, velocity, pressure, `gamma`, `pi_inf`, `qv`, and the sound speed.

`lag_bubble` files were already special-cased for exactly this reason. Probe files are the remaining case.

## Change

Retain every column for probe files. No header line to skip — `s_write_probe_files` writes data rows only.

## Effect on existing goldens

| Golden | Before | After |
|---|---:|---:|
| `tests/5CAA4E68` (1D exp_bubscreen) | 50 values | 450 |
| `tests/FBB296DA` (1D bubblescreen) | 50 values | 450 |
| `tests/AE9A7D73` (1D poly_bubscreen) | 1 value | 9 |

Nine columns per row instead of one. The added values are what was being dropped: density, velocity, pressure, void fraction, and the bubble radius/velocity moments. That widening is the point of the change — it is coverage these cases should always have had.

The regenerated metadata records `(dirty)`. That is inherent to regenerating *tracked* goldens: writing them dirties the tree before the metadata is stamped. Newly added goldens do not have this problem.

## Verification

The three probe cases pass against their regenerated goldens. The change is gated on `"probe" in short_filepath`, so non-probe files take the original code path unchanged and no other golden should move. A full local suite run confirms it: **627 passed, 0 failed**.

An earlier run of the same suite reported the three probe cases failing with "Variable count didn't match". That did not reproduce — the three pass in isolation, pass on consecutive repeat runs, and pass in the clean full-suite run above. It matches non-reproducible flakiness seen on an unrelated branch this session (two chemistry cases, likewise green individually), and appears to be local contention at high `-j` rather than anything in this change. Worth knowing it has been seen, in case CI shows it.

## Why now

#1707 is a defect in the probe sound speed. It could not be given a regression test through the normal golden path, because `c` is never the last column in any configuration — a case could exercise the defective code, emit visibly wrong output, and still pass. This PR is a prerequisite for testing that fix, and stands on its own regardless.

Note that even with this change, observing `c` specifically requires a 3D probe case; the 1D writes do not emit it at all.

---

# Handover notes

## Branch and commits

```
repo    MFlowCode/MFC
branch  fix/packer-probe-columns    (base: master)
  0367641  Keep every column of probe output in golden packs
  c185dc8  Regenerate the three probe goldens with full column coverage
```

Split into two commits so the goldens are regenerated from a **clean checkout** of the packer change. Note the regenerated metadata still records `(dirty)` — that is unavoidable when regenerating *tracked* goldens, because writing them dirties the tree before the metadata is stamped. Newly added goldens do not have this problem. Do not chase it.

## Environment

```bash
./mfc.sh build -j 16
./mfc.sh test  -j 12
./mfc.sh test --only 5CAA4E68 AE9A7D73 FBB296DA          # the three probe cases
./mfc.sh test --generate --only 5CAA4E68 AE9A7D73 FBB296DA
```

GNU 15.2.0, MPI, no GPU, macOS arm64.

## The mechanism, precisely

`toolchain/mfc/packer/pack.py`, in `compile()`:

```python
ndims   = len(_extract_doubles(content.split("\n", 1)[0])) - 1
doubles = _extract_doubles(content)[ndims :: ndims + 1]
```

Every `.dat` under `D/` is assumed to be `<x> [<y> <z>] <value>`, so only the **last column of each row** survives. `lag_bubble` files were already special-cased above this for the same reason; probe files were the remaining case. The fix adds an `elif "probe" in short_filepath` branch keeping all columns. `s_write_probe_files` writes data rows only — there is no header to skip.

## Why this was needed

#1707 is a defect in the probe sound speed. It could not be given a regression test through the normal golden path, because `c` is never the last column in any configuration:

| configuration | columns written | kept before this PR |
|---|---|---|
| 1D general (`m_data_output.fpp`) | `nondim_time, rho, vel(1), pres` | `pres` |
| bubbles | `nondim_time, rho, vel(1), pres, alf, R, Rdot, nR, nRdot` | `nRdot` |
| hypoelastic | `nondim_time, rho, vel(1), vel(2), pres, tau_e(1..3)` | `tau_e(3)` |
| 3D | `nondim_time, rho, vel(1..3), pres, gamma, pi_inf, qv, c, accel` | `accel` |

A case could exercise the defective code, emit visibly wrong output, and still pass its golden. I confirmed this empirically: a purpose-built `qv /= 0` probe case passed with the bug present.

**Even after this PR, observing `c` requires a 3D probe case** — the 1D writes do not emit it at all. Anyone writing the #1707 regression test needs 3D plus `qv /= 0`.

## Golden impact

| golden | before | after |
|---|---:|---:|
| `tests/5CAA4E68` (1D exp_bubscreen) | 50 values | 450 |
| `tests/FBB296DA` (1D bubblescreen) | 50 values | 450 |
| `tests/AE9A7D73` (1D poly_bubscreen) | 1 value | 9 |

Nine columns per row instead of one. The widening is the point.

## Verification, and a flake to expect

```
three probe cases:  3 passed
full suite:         627 passed, 0 failed
```

An earlier full-suite run reported these same three cases failing with `Variable count didn't match`. It did not reproduce: they pass in isolation, pass on two consecutive repeat runs, and pass in a clean full-suite run. It matches non-reproducible flakiness seen on unrelated branches in the same session (two chemistry cases, likewise green individually) and appears to be local contention at high `-j`. If CI shows it, re-run before investigating.

## Related work

- #1711 — the issue this closes.
- #1707 — blocked on this; needs a 3D probe case with `qv /= 0`.
- #1714 — `eos_state` refactor, which subsumes the #1707 code fix via the interface.

---

# Purpose in the series

Test-infrastructure prerequisite, not a feature.

#1707 is a defect in the probe sound speed that **could not be given a regression test** — `c` is never the last column of probe output, and the packer kept only the last column. A case could exercise the defective code, emit visibly wrong output, and pass its golden. I confirmed that empirically before writing this.

So this exists to make a class of regression testable at all. The widened goldens are the point, not a side effect: three cases went from validating one column to validating nine.

Note #1714 fixes the #1707 code path via the interface, so this is no longer strictly blocking that fix — but it remains the only way a probe regression becomes visible to the suite.

---

## Testing

## GPU

## Benchmarking

## GitHub mechanics


sbryngelson:
Ordering note: #1716 removes integral output entirely (unused since it was added; confirmed with its author).

Once both land, the `or "integral" in short_filepath` clause added here becomes dead and should be dropped, along with the mention of `int_pres`/`max_pres` in the surrounding comment. Harmless if left — the match simply never fires — but it will read as referring to a feature that no longer exists.

The two PRs do not conflict and can merge in either order. If #1716 goes first I will drop the clause from this branch before merge; if this goes first I will drop it in a follow-up.

PR #1714, opened 2026-08-09
Pass an eos_state to s_compute_speed_of_sound

sbryngelson:
Closes the enthalpy half of #1708. Also subsumes #1707.

## The problem

`s_compute_speed_of_sound` took ten positional arguments, and the contract between two of them was invisible at the call site: `H` must include `qv`, because the routine subtracts `qv/rho` internally (`m_variables_conversion.fpp`, default stiffened-gas branch).

Nothing stated that. Of the five sites that open-coded `H`:

- one was correct (`post_process/m_data_output.fpp`),
- three were wrong (`simulation/m_data_output.fpp` — #1707),
- one was correct *only because two separate omissions cancelled* (`s_compute_cson_from_pinf` in `m_bubbles_EL.fpp`), which is a trap for the next person who "fixes" it.

## The change

`type(eos_state)` in `m_derived_types` carries the scalars the routine needs, with two ways to build one:

- **`s_eos_state`** derives `H` from the other members, so it cannot disagree with `qv`. The #1707 defect is unrepresentable through this path.
- **`s_eos_state_roe`** takes `H` explicitly, for the Roe-averaged Riemann paths, the chemistry Roe branch and the relativistic branch — all of which legitimately pass an `H` that is *not* the exact state enthalpy.

The call becomes `(state, adv, c)`. All 24 call sites across nine files are converted, and the state is added to the `private()` clause of every GPU parallel loop that builds one.

`adv` stays a separate argument rather than a component. A derived-type component cannot have a runtime extent, and `num_fluids` is a `parameter` only under case optimization, so `dimension(num_fluids)` does not compile in a general build:

```
Error: Variable 'num_fluids' cannot appear in the expression at (1)
```

Padding to `num_fluids_max` would compile but put ten reals into a per-cell private struct on device.

## The safety guarantee is uneven — read this before approving

There are three ways to build a state, and they do not all enforce the invariant:

| path | sites | enforces `H` includes `qv`? |
|---|---:|---|
| `s_eos_state` | 4 | **yes** — derives `H`, cannot disagree |
| `s_eos_state_roe` | 4 | no — caller supplies `H`, which is the point for Roe paths |
| intrinsic `eos_state(...)` | 16 | no — every component is caller-supplied |

The 16 hot Riemann sites use the intrinsic form for performance (below), so at those sites a wrong `H` is still *writable*. What makes them safe in practice is that they all pass an `H` derived from the conserved energy (`H_L = (E_L + pres_L)/rho_L`), which carries `qv` implicitly and cannot drift — the same property that kept them correct before this PR.

So the accurate claim is narrower than "the bug is now impossible": the defect is unrepresentable **at the sites that had it** (#1707, all diagnostic, all now using `s_eos_state`), and the sites that never had it keep the construction that made them correct. If you would rather trade the ~5% on HLLC for uniform enforcement, switching those 16 back to `s_eos_state_roe` is a one-line change per site.

A related hazard worth knowing: the **type's component order is `rho, pres, ...` while the procedure constructors take `pres, rho, ...`**. The intrinsic sites are positional and follow the type. Getting that backwards compiles cleanly and silently swaps two doubles; it is caught only by the goldens.

## Performance

The first cut cost +5–8% on the Riemann benchmarks. Profiling showed it was **not** inlining loss — `s_compute_speed_of_sound` is a real, non-inlined call both before and after. It was the constructor call added on top of it. Object counts for `m_riemann_solver_hllc` (arm64, GNU 15.2):

| build | total `bl` | relocs -> `speed_of_sound` | relocs -> `eos_state_roe` |
|---|---:|---:|---:|
| baseline | 179 | 27 | 0 |
| constructor call | 206 | 27 | 27 |
| structure constructor | 179 | 27 | 0 |

Nine call sites x 3 relocations = exactly the 27 added calls, one per site. Building the state at the 16 hot Riemann sites with Fortran's **intrinsic structure constructor** builds it in place and restores the baseline call count exactly. The procedure constructors remain at the diagnostic sites, where deriving `H` is the point and call cost is irrelevant.

Benchmarks after that fix (idle machine, versus the mean of two baseline runs; measured baseline noise +/-3.5%):

| case | delta |
|---|---:|
| 5eq_rk3_weno3_hll | -0.12% |
| 5eq_rk3_weno3_lf | -2.12% |
| hypo_hll | -1.00% |
| ibm | -0.57% |
| igr | -0.09% |
| viscous_weno5_sgb_acoustic | -1.26% |
| **5eq_rk3_weno3_hllc** | **+5.63%** |

Six of seven sit at or below baseline. **HLLC is a genuine outstanding cost** — consistent across two independent runs and above both baseline samples, though the baseline spread there is wide enough that the magnitude is soft (call it +4-7%). The call count is identical to baseline, so it is not call overhead; the likely cause is materialising the state itself, since HLLC has nine call sites, more than any other solver. Reviewers should weigh that against the interface guarantee.

## Relationship to #1707

The three probe sites now use `s_eos_state`, so the #1707 correction falls out of the interface rather than being a separate patch. That changes the reported probe sound speed where `qv /= 0`. No golden observes it, because the packer keeps only the last column of probe output (#1711, fixed in #1712).

## Verification

```
build (GNU 15.2, MPI):   clean
full suite:              627 passed, 0 failed
goldens regenerated:     none
precheck (7/7):          pass
```

A scan confirms no GPU parallel loop uses a state without declaring it `private`.

## Still open in #1708

This covers the enthalpy and sound-speed half. The four open-coded copies of the **pressure inversion** remain (#1709 fixes the one that was wrong); routing them through a single definition is follow-up work.

---

# Handover notes

## Branch and commits

```
repo    MFlowCode/MFC
branch  refactor/eos-state          (base: master, rebased after #1713 merged)
  936c676  Pass an eos_state to s_compute_speed_of_sound
  102c54c  Build hot-path eos_states with the intrinsic structure constructor
```

Originally built on top of #1713 and rebased with
`git rebase --onto origin/master <removal-sha> refactor/eos-state`
once that merged. Re-verified after the rebase: build clean, 627/0.

## Environment

```bash
./mfc.sh build -j 16
./mfc.sh test  -j 12
./mfc.sh bench --mem 1 -o out.yaml     # ~12 min, 7 cases
```

GNU 15.2.0, MPI on, no GPU, macOS arm64 (18 cores). **This branch has never been built for GPU or with case optimization** — both are worth checking in CI, since the type is used inside `[seq]` device routines and `private()` clauses.

## Where things live

| what | where |
|---|---|
| `type eos_state` | `src/common/m_derived_types.fpp`, next to `qbmm_idx_info` |
| `s_eos_state`, `s_eos_state_roe` | `src/common/m_variables_conversion.fpp`, immediately before `s_compute_speed_of_sound` |
| converted routine | same file, `s_compute_speed_of_sound(s, adv, c)` |
| call sites | 24 across `m_riemann_solver_{hll,hllc,hlld,lf}`, `m_cbc`, `m_data_output` (sim + post), `m_time_steppers`, `post_process/m_start_up` |

Component order is `rho, pres, gamma, pi_inf, qv, vel_sum, H, c_c` — note the **constructors take `pres` before `rho`**, the type does not. The structure-constructor call sites are positional and follow the *type* order; getting this backwards compiles fine and silently swaps two doubles.

## Two constructors, and when to use which

- `s_eos_state(...)` derives `H`. Use wherever the exact state enthalpy is wanted. This is what makes #1707 unrepresentable.
- `s_eos_state_roe(..., H [, c_c])` takes `H`. Required for Roe-averaged Riemann paths, the chemistry Roe branch, and the relativistic branch, which pass an `H` that is deliberately not the exact enthalpy. Do not "simplify" these to the derived form — it changes the scheme.
- At the 16 hot Riemann sites the state is built with Fortran's **intrinsic structure constructor** (`eos_s_L = eos_state(...)`) rather than a call, for performance (below). Keep it that way.

## Performance: the whole story

First cut cost +5-8% on the Riemann benchmarks. It was **not** inlining loss — `s_compute_speed_of_sound` is a real, non-inlined call both before and after (checked with `nm -u` on the object). It was the constructor call stacked on top.

Reproduce the object-level check:

```bash
O=build/staging/cpu-*/CMakeFiles/simulation.dir/fypp/simulation/m_riemann_solver_hllc.fpp.f90.o
objdump -d "$O" | grep -cE '^\s+[0-9a-f]+:.*\bbl\b'    # total call instructions
objdump -r "$O" | grep -c speed_of_sound               # relocations
objdump -r "$O" | grep -c eos_state_roe
```

| build | `bl` | relocs speed_of_sound | relocs eos_state_roe |
|---|---:|---:|---:|
| baseline | 179 | 27 | 0 |
| constructor call | 206 | 27 | 27 |
| structure constructor | 179 | 27 | 0 |

**Benchmark methodology matters.** Run `./mfc.sh bench` in the **foreground on an idle machine**. I once ran it in the background while the pre-commit precheck was running at `-j 12` and got a bogus +50% on `igr` — a case that executes none of the converted code (`run_time_info = F` in every benchmark case, and no case sets `cfl_dt`, so both per-cell loops using the state are switched off). Baseline-vs-baseline noise on this machine is +/-3.5%; run the baseline twice before trusting any delta.

## Outstanding: HLLC ~+4-7%

Real, consistent across two independent runs, above both baseline samples. Call count is identical to baseline so it is not call overhead; the likely cause is materialising the state at HLLC's nine call sites, more than any other solver. Not yet chased. Ideas, cheapest first:

1. More baseline samples on HLLC alone to narrow the error bar — the current estimate rests on two samples per side and the baseline spread is wide (68.82 vs 66.42).
2. Trim the struct. `c_c` is read only in the chemistry branch and `H` only in some branches; a smaller state means fewer stores per site.
3. Check whether `-flto` or NVHPC/Cray change the picture; this was measured on gfortran only.

## Verification performed

```
build:                clean
full suite:           627 passed, 0 failed
goldens regenerated:  none
precheck (7/7):       pass
GPU private() audit:  no loop uses a state without declaring it private
```

The `private()` audit is worth redoing if you add call sites — a CPU run cannot catch a missing `private()`, it is a silent device race. Script it by finding `$:GPU_PARALLEL_LOOP(` directives (**excluding** `END_GPU_PARALLEL_LOOP`, which contains the same substring and will produce false positives), joining `&` continuations, and checking every `eos_s*` used in the body appears in the directive text.

## Related work

- #1708 — parent issue. This closes the enthalpy/sound-speed half. The **pressure-inversion half is still open**: four open-coded copies at `m_pressure_relaxation.fpp`, `m_data_output.fpp` (x2), `m_bubbles_EL.fpp`.
- #1707 — subsumed here by construction. Its standalone branch (`fix/probe-sound-speed-qv`, unopened) has a focused regression test that is still worth keeping, but needs #1712 plus a 3D probe case to be observable.
- #1712 — packer fix; prerequisite for any probe-based regression test.
- #1713 — merged; this was rebased off it.

---

# Purpose in the series

Closes the enthalpy half of #1708 and gives #1700's selector something to dispatch through.

The underlying problem is that MFC has nine hand-written copies of two EOS expressions — four wrong, one right only by a cancellation no reader can see locally. Adding a second equation of state under #1638 multiplies that: every open-coded site is a place the new backend silently does not reach.

This PR attacks the enthalpy/sound-speed half by making the invisible contract (`H` must include `qv`) explicit in a type. **The pressure-inversion half remains open** — four copies at `m_pressure_relaxation.fpp`, `m_data_output.fpp` (x2) and `m_bubbles_EL.fpp`.

Read the "safety guarantee is uneven" section before approving; the enforcement is not uniform across the 24 sites, deliberately.

---

## Testing

## GPU

## Benchmarking

## GitHub mechanics


sbryngelson:
Pushed `c319903`, addressing all three review findings.

**Doc fixes.** The stale `!> Compute the speed of sound ...` line above `s_eos_state` is moved down to `s_compute_speed_of_sound`, which it actually describes and which was left undocumented. The `eos_state` type comment referred to `f_eos_state` / `f_eos_state_roe`; the constructors are subroutines, so those are now `s_eos_state` / `s_eos_state_roe`.

**The `optional :: c_c` concern.** The review asked for this to be verified on the offload backends rather than assumed. Checking it first, the argument turns out to be dead: no caller anywhere passes `c_c`. All four call sites (`m_cbc.fpp:665`, `m_data_output.fpp:207`, `m_time_steppers.fpp:681`, `post_process/m_start_up.fpp:537`) omit it, and the hot Riemann paths never call `s_eos_state_roe` at all — they build the state with the intrinsic structure constructor, which is where a non-zero `c_c` actually comes from:

```fortran
eos_s_avg = eos_state(rho_avg, pres_R, gamma_avg, pi_inf_R, qv_avg, vel_avg_rms, H_avg, c_sum_Yi_Phi)
```

So rather than verifying that an OPTIONAL dummy behaves correctly in a device routine, the dummy and its `present()` check are removed. `s%c_c = 0._wp` already covers every caller. That eliminates the pattern the review flagged instead of relying on it being safe.

**Verification on AMD flang.** Built and ran the full suite on an MI250X (gfx90a) node with the latest AFAR drop (`therock-afar-23.2.1-gfx90a-7.13.0`, `amdflang`), OpenMP target offload, `OMP_TARGET_OFFLOAD=MANDATORY` so a silent host fallback would abort rather than pass:

```
✓ ALL TESTS PASSED
   627 passed · 0 failed · 32 skipped · 659 total · 103m 15s
```

This is the post-fix tree. An earlier run on the pre-fix head also passed every test it could build (608), with 19 failures that were solely a missing `gpu-mp-chem-*` build variant, not numerical.

sbryngelson:
Superseded by #1762, which takes the opposite approach: rather than bundling the scalars into a type, it removes them.

`H`, `|u|^2` and `qv` cancel out of the sound-speed relation for a real state, so they were never needed as arguments. #1707 becomes unrepresentable rather than fixed.

The type is also what costs the performance. Measured on an MI210 (gfx90a, AFAR amdflang, OpenMP offload), `5eq_rk3_weno3_hllc`:

| | scratch | VGPR | AGPR | ns/gp/eq/rhs |
|---|---|---|---|---|
| master | 28 / 236 / 108 B | 134 | 2 | 2.720 |
| this PR | 4880 / 5072 / 4960 B | 394 | 138 | 3.266 (+20%) |
| with `value` | identical | 394 | 138 | 3.202 |

A derived-type dummy on a `declare target [seq]` routine forces a memory ABI on an address-taken aggregate, and register promotion collapses across the whole enclosing loop, not just the call. Declaring it `value` does not help - the code object is byte-identical. It is not the constructor call either; that was removed in the second commit here and recovered nothing.

Recorded in `.claude/rules/common-pitfalls.md` on branch `eos/gpu-guards` so the next attempt does not repeat it.

PR #1716, opened 2026-08-09
Remove integral output

sbryngelson:
Removes integral output — `integral_wrt`, `num_integrals`, and the 30 `integral(i)%{x,y,z}{min,max}` parameters.

## Why

It writes region-integrated pressure to `D/integral<i>_prim.dat` for Euler-Euler bubble runs. Nothing exercises it: no example, test, or benchmark sets `integral_wrt`, and it has had no coverage since it was added roughly nine years ago. Confirmed with the maintainer (its author) that it has no users.

Unlike the four-equation removal, this touches **no hot path** — the output was cold-guarded behind `if (integral_wrt)`. The value here is removed surface area, not simplified code. `130 lines` of `182` deleted are the output block itself.

## What was removed

**Fortran**

- the 118-line output block in `s_write_probe_files` and the file-open block in `s_open_probe_files` (`simulation/m_data_output.fpp`)
- `type integral_parameters` (`common/m_derived_types.fpp`)
- the fypp MPI broadcast loop over `integral(j)%{xmin..zmax}` (`simulation/m_mpi_proxy.fpp`)
- the default assignments (`simulation/m_global_parameters.fpp`)

**Toolchain**

- registry entries, constraints and descriptions for `integral_wrt`, `num_integrals`, and the 30 `integral(i)%*` parameters
- `check_probe_integral_output` becomes `check_probe_output`, keeping only the probe `fd_order` rule; `lint_docs.py`'s skip-set entry renamed to match
- `case.md` entries

**Locals that die with it:** `int_pres` and `max_pres` (used only in the removed block), plus `rad`, `thickness` and `trigger`, whose declarations were annotated "For integral quantities".

## Verification

```
build (GNU 15.2, MPI):   clean
full suite:              627 passed, 0 failed
goldens regenerated:     none
toolchain unit tests:    170 passed
precheck (7/7):          pass
```

No golden moved, which is expected — no test enabled the feature. That is also the reason it could be removed with this little risk.

## Ordering note for #1712

#1712 extends the golden packer to keep every column of **probe and integral** output. Once both PRs land, the `or "integral" in short_filepath` clause in `toolchain/mfc/packer/pack.py` becomes dead and should be dropped. It is harmless if left — the match simply never fires — but it will read as referring to a feature that no longer exists.

Merge order does not otherwise matter; the two do not conflict.

---

# Handover notes

## Branch

```
repo    MFlowCode/MFC
branch  remove/integral-output   (base: master, at b40e08f)
  ca7840d  Remove integral output
```

## Environment

```bash
./mfc.sh build -j 16
./mfc.sh test  -j 12
cd toolchain && python -m pytest mfc/params_tests -q
```

GNU 15.2.0, MPI on, no GPU, macOS arm64. Not built for GPU or with case optimization — worth confirming in CI, though the removal touches no GPU-declared state beyond deleting the `integral` broadcast.

## How the removal was scoped

Start from `grep -rn "integral_wrt\|num_integrals\|integral(" src/ toolchain/ docs/`. Two false positives to keep:

- `m_thinc.fpp`'s `f_mthinc_volume_integral` — unrelated volume integral in the THINC reconstruction.
- any `integral` hit inside `params_tests/` fixtures (there were none, but check before deleting).

The output block was located by matching `if (integral_wrt .and. bubbles_euler) then` and walking to its balanced `end if` — 118 lines, ending at the subroutine's last statement. If you redo this, count `if (...) then` / `end if` at matching depth rather than relying on indentation.

After removing the block, check for locals that were only used inside it. `int_pres` and `max_pres` dropped to zero occurrences; `rad`, `thickness` and `trigger` were left with exactly one (their declarations), which is the signature of an orphan.

## What was deliberately not done

No deprecation period. The feature is undocumented beyond two table rows, has no test coverage, and its author confirms no users. A case setting `integral_wrt` will now fail validation with the standard unknown-parameter error, which names the offending key.

---

# Purpose

Surface-area reduction, same argument as #1713 but weaker and worth stating honestly.

`model_eqns = 4` was removed because it was an unverifiable physics model that forced branches through the pressure inversion, energy assembly, primitive recovery, HLLC, IBM, acoustic source and both init paths — removing it deleted 500 lines and simplified hot code.

Integral output is **not** that. It is a cold-guarded diagnostic writer costing ~20 references across two files, with no hot-path branching. I argued against removing it on those grounds: it is a documented, registered, opt-in feature, and removal breaks any external user silently.

What changed the decision is a fact only the maintainer could supply — he wrote it about nine years ago and confirms it has no users. With that settled the case is clear, but the precedent should not be read as "untested implies removable". The reasoning was: untested **and** no hot-path cost **and** authorial confirmation of zero users.

---

## Testing

## GPU

## Benchmarking

## GitHub mechanics


sbryngelson:
Pushed `2b618b2` with the three review findings fixed, plus four further references to the removed feature that the review did not catch:

- `m_global_parameters.fpp` — the dead `do i = 1, num_probes_max / end do` shell
- `m_data_output.fpp` — the now-unset `npts` declaration
- `descriptions.py` — the six `integral(i)%...` regex descriptions
- `docs_gen.py` — the `"integral": "Integral region parameters"` doc-section label (the only functional one of the extras; the rest are comments/labels)
- `definitions.py` — `TAG_DISPLAY_NAMES["probes"]` read `"Probe/integral"`, and the `NPR` comment listed `integral`
- `descriptions.py` — `FEATURE_DESCRIPTIONS["probes"]` read `"Probe points and integral regions"`
- `m_global_parameters.fpp:201`, `m_mpi_proxy.fpp:212` — comments naming `integral`

A repo-wide grep for `integral_wrt`, `num_integrals`, and `integral(` now returns only the unrelated THINC volume integrals. Net −12 LOC.

Verified on an MI250X (gfx90a) node with the latest AFAR drop (`therock-afar-23.2.1-gfx90a-7.13.0`, `amdflang`), OpenMP target offload:

```
✓ ALL TESTS PASSED
   627 passed · 0 failed · 32 skipped · 659 total · 125m 2s
```

PR #1670, opened 2026-07-23
Combustion: condensed-phase reactive burn (5-/6-equation, Arrhenius kinetics) + IBM burning surfaces

sbryngelson:
## Description

Adds two combustion capabilities to the solver, along with example cases and regression tests. The work targets solid-fuel / condensed-phase and immersed-boundary combustion (internal ballistics, hybrid rockets, bluff-body flame holding, shock–flame interaction).

**1. Condensed-phase reactive burn** (`src/simulation/m_reactive_burn.fpp`, `src/simulation/m_pressure_relaxation.fpp`)
A programmed, pressure-driven burn on the multi-fluid model that converts a reactant fluid to a product fluid, releasing chemical energy — a lightweight route to detonation/deflagration without finite-rate kinetics.

It runs on **both** the 5-equation (pressure-equilibrium) and 6-equation (Saurel/Pelanti phasic-energy) models. For `model_eqns = 3` the reactant→product `qv` release is carried through the phasic-pressure relaxation: `s_equilibrate_pressure` / `s_correct_internal_energies` now thread the formation energy `alpha_rho_k*qv_k`, making the phasic and mixture pressures qv-consistent with `s_compute_pressure` (a nonzero `qv` otherwise inflated the phasic pressure by `rho_k*qv_k/gamma_k`). For `qv = 0` the new terms are exactly zero, so every existing 6-equation golden is bit-identical. An optional **Arrhenius** term gives temperature-driven ignition instead of a pure pressure switch.

Parameters are grouped in a `rburn` derived type (following `chem_params`):

| Parameter | Meaning |
|---|---|
| `reactive_burn` | Enable the burn |
| `rburn%k` | Rate coefficient [1/s] |
| `rburn%pign` | Ignition-pressure threshold [Pa] |
| `rburn%pref` | Reference pressure for the pressure drive [Pa] |
| `rburn%n` | Pressure-drive exponent |
| `rburn%ta` | Activation temperature [K]; `0` = pure pressure-driven, `>0` multiplies the rate by `exp(-rburn%ta/T)` with `T` the reactant phasic temperature |

[...]

PR #1717, opened 2026-08-09
Move input-only checks to case_validator, delete dead params_tests, enforce the split

sbryngelson:
## Description

Three related changes. Net **−1,363 lines** (409 added, 1,772 removed).

### 1. Move input-only checks out of Fortran

`case_validator.py` and `src/simulation/m_checker.fpp` had drifted into double entry — `reactive_burn` (#1670) added nine constraints to **both**, in the same PR.

| Check | Status |
|---|---|
| `reactive_burn` num_fluids / gamma / pi_inf / qv | new in Python |
| chemistry operator-split sub-stepping (4) | new in Python |
| synthetic-turbulence forcing zones (3) | new in Python (`check_synthetic_turbulence`) |
| `many_ib_patch_parallelism requires ib` | new in Python |
| `bf_spatial_support` 2D-only | new in Python |
| `dt <= 0`, `ib_state_wrt`, `chemistry + bubbles_euler/qbmm` | already in Python — Fortran duplicate dropped |
| `particle_cloud%packing_method` (2), `muscl_order`/`int_comp` | already in Python — Fortran duplicate dropped |

`src/simulation/m_checker.fpp`: **225 → 113 lines.** `s_check_inputs_time_stepping`, `s_check_inputs_particle_clouds`, and `s_check_inputs_synthetic_turbulence` are gone entirely, along with the now-unused `m_helper_basic` import and `muscl_order_first_order`.

**Fortran defaults are materialized where an unset value is meaningful.** `chem_params%reaction_substeps{,_max}` and `fluid_pp(i)%qv` default to `0`, so unset is treated as `0`. `gamma`/`pi_inf` default to the `dflt_real` sentinel, which `f_approx_equal` reports as equal, so the comparison is skipped when either is unset. `check_synthetic_turbulence` loops `d = 1..num_dims` to match the Fortran, so trailing components of a lower-dimensional case stay optional.

**`nv_uvm_igr_temps_on_gpu` deliberately stays in Fortran.** Its default is `3` and the check is `== 3 .and. igr_iter_solver == 2`, so it fires precisely when the user has *not* set the parameter. In Python an unset value reads as `None` and the check would silently never fire — moving it would look like a migration and act like a deletion. It is also inside `#ifdef __NVCOMPILER_GPU_UNIFIED_MEM`.

What remains in Fortran depends on state unavailable at validation time: MPI decomposition (`s_check_total_cells`, `s_check_inputs_fft`), per-rank `m`/`n`/`p` (WENO/MUSCL stencils), compiler conditionals (`s_check_amd`, `rdma_mpi`), and `num_species`, which Cantera fills in at runtime.

### 2. Delete the `params_tests` CLI island — −1,651 lines

`coverage.py`, `inventory.py`, `mutation_tests.py`, `negative_tests.py`, `runner.py`, `snapshot.py`. Together they collect **zero** pytest tests. Every one is imported only by `runner.py`, which nothing imports and which appears in no workflow, script, doc, or CMake file. Its whole history is incidental edits from unrelated feature PRs — #1713 had to update `mutation_tests.py` to drop `model_eqns = 4` earlier today.

It could not be promoted as-is: `negative_tests` decides whether the right error fired with a `matches >= len(key_terms) * 0.5` substring heuristic, and `runner verify` needs a `data/` baseline that is gitignored. The sibling `test_*.py` files, which supply all 171 real tests in the package, are untouched.

### 3. Enforce the split — `check_checker_input_constraints`

New check in `lint_source.py`. A `@:PROHIBIT` in `src/**/m_checker*.fpp` is allowed only inside a subroutine listed in `RUNTIME_CHECKER_SUBROUTINES`, or on a line marked `! lint: runtime-check <reason>`. Otherwise it fails with a pointer to `case_validator.py`.

This is what stops the next feature PR from re-creating the double entry this one removes.

### Type of change

- Refactor

## Testing

- `./mfc.sh precheck` — all 7 stages pass (full pytest suite + all 155 example cases), run repeatedly including via the pre-commit hook
- Built `pre_process` / `simulation` / `post_process` — clean, no new warnings
- `./mfc.sh test --percent 4` — 24 passed, 0 failed, on both commits
- New `toolchain/mfc/test_case_validator.py` — **32 tests**, every migrated check in both directions
- Lint rule verified both ways: passes on the tree, and rejects an injected input-only `@:PROHIBIT` with the right message; the `! lint: runtime-check` escape hatch suppresses it
- `check_synthetic_turbulence` verified against the real `examples/2D_synthetic_turbulence` case — accepts it, and rejects it once `synth_L(1,2)` is removed

`case_validator.py` had **no pytest coverage of its constraint checks**, and validating the example cases only exercises configurations meant to pass, so a check that stopped firing would go unnoticed. That matters more now that these have no Fortran backstop. Reverting `case_validator.py` fails 15 of the original 24 tests; the 9 that survive are exactly those covering checks that already existed in Python.

## Checklist

`physics_constraints.md` / `case_constraints.md` are generated from `case_validator.py` at docs-build time. A `PHYSICS_DOCS` entry was added for `check_synthetic_turbulence` — `lint_docs.py` caught its absence, which is the doc-coverage gate working as intended.

<details>

Validation-only; no kernel, data-movement, or numerics changes. The one GPU-conditional check (`nv_uvm_igr_temps_on_gpu`, under `#ifdef __NVCOMPILER_GPU_UNIFIED_MEM`) is deliberately left in place, and `s_check_inputs_nvidia_uvm` is on the lint rule's runtime allowlist.

</details>


sbryngelson:
Thanks — went through all four findings. Two were valid and are fixed in 94b8565; two are false positives, with evidence below.

## Fixed — Copilot, `test_case_validator.py`

**`errors_for` caught bare `Exception`.** Correct, and the consequence was real: a regression raising `KeyError` would be stringified and could satisfy an assertion. Now catches only `CaseConstraintError`; anything else propagates.

**`assertAccepts` only checked one message was absent.** Also correct — a case broken for an unrelated reason would still pass. It now asserts full validity (no violations at all) and takes only `params`. I checked every call site individually first; all seven were already fully valid configurations, so nothing had to be relaxed to accommodate the stronger form.

`assertRejects` also gained an explicit "did it fail at all" check, so a wrongly-accepted case now reports `expected validation to fail with X` instead of `X not found in ''`.

Verified the new assertions catch what the old ones missed:
- injecting a `KeyError` into `CaseValidator.validate` → test now **errors** (previously would have passed)
- `assertAccepts` on a fixture with an unrelated violation (`dt <= 0`) → now **fails**
- `assertRejects` on a case that validates cleanly → now **fails**

## Not a defect — Claude review, "lost validation" ×2

Both checks were **already in `case_validator.py` on `master`** before this PR. They were among the nine `reactive_burn` / MUSCL constraints that existed in *both* languages — which is the duplication this PR removes. Only the Fortran copy was deleted; the Python side is untouched by this diff.

| Claim | Where it lives on `master` |
|---|---|
| `rburn%ta` / `fluid_pp(1)%cv` | `case_validator.py:1571` onward, in `check_reactive_burn` |
| `muscl_order` / `int_comp` | `case_validator.py:431` |

`git diff master...HEAD -- toolchain/mfc/case_validator.py` contains no `+`/`-` line touching `rburn%ta`, `%cv`, or `int_comp`.

Confirmed by running the validator on this branch:

```
rburn%ta < 0                          -> CAUGHT
rburn%ta > 0 with fluid_pp(1)%cv = 0  -> CAUGHT
rburn%ta > 0 with cv unset            -> CAUGHT
muscl_order = 1 with int_comp = 1     -> CAUGHT
```

Message for the second: `reactive_burn with rburn%ta > 0 requires fluid_pp(1)%cv > 0 (the reactant temperature needs a physical heat capacity; cv = 0 silently disables the Arrhenius factor)` — the exact constraint the review reported as dropped.

Worth noting these two are *only* covered by `check_reactive_burn` / `check_interface_compression`, not by the new tests, since they predate this PR. Happy to add cases for them if that's wanted, though it widens the scope past the migration.

Issue #1512, opened 2026-06-01
misc tooling dedup: twin result dict, identical doc getters, hardcoded sched thresholds, redundant ORG_COLORS


sbryngelson:
A grab-bag of small cleanups across five tooling files. Four are clean good-first-issue dedups; (b) is a larger refactor that should be scoped separately.

**(a) fp_stability.py — 13-key result dict duplicated.** The identical default-result dict (`name, passed, max_dev, threshold, float_proxy, vprec, dd_sym_syms, dd_line_locs, cancellation_locs, mca_dev, mca_sigbits, float_max_locs`) appears at `toolchain/mfc/fp_stability.py:970-983` and again at `:1322-1335`, differing only in where `name`/`threshold` come from. → extract `_empty_result(name, threshold) -> dict`.

**(c) gen_case_constraints_docs.py — three identical getters + duplicated emoji map.** `get_model_name` (133), `get_riemann_solver_name` (140), `get_time_stepper_name` (147) have byte-identical bodies modulo the schema key string. Also the level→emoji map `{"Beginner": "🟢", "Intermediate": "🟡", "Advanced": "🔴"}` is duplicated at lines 159 and 267. → collapse to one `_named(param, value)` helper and hoist the emoji map to a module constant.

**(d) sched.py — `HEADLESS_THRESHOLDS` table is not used for its numbers.** The table at `sched.py:21-25` stores `(2*60, msg), (10*60, msg), (30*60, msg)`, but the consumer (143/148/153) re-hardcodes `elapsed >= 2 * 60` etc. and only indexes the table for the message text. → drive the comparisons from `HEADLESS_THRESHOLDS[i][0]` so the seconds live in one place.

**(e) user_guide.py — `ORG_COLORS` is entirely redundant.** All 8 entries (lines 56-65) map to `"yellow"`, the lookup uses `ORG_COLORS.get(org, "yellow")` (line 145), and `"CSCS"` (in `ORG_ORDER`) is missing from the dict. Net effect: every org renders yellow regardless. The missing CSCS key is harmless (falls back to "yellow"), so this is dead config, not a bug. → delete `ORG_COLORS` and inline `"yellow"`, or give orgs distinct colors if intended.

[...]

PR #1720, opened 2026-08-10
Tooling dedup: gen_case_constraints_docs.py (c), sched.py (d), user_guide.py (e)

AnaPcode:
## Description

Addresses (a), (c), (d), (e) from issue 1512, following the specific fixes the issue suggested for each file except for (a).

(a) fp_stability.py - no changes made. The file seems to have changed since the issue was filed (now 693 lines), and the cited line numbers (970-983, 1322-1335) no longer correspond to the described code. Found a _blank_result(name) helper already present, but couldn't locate a dict matching the original 13-key description anywhere in the file. Is this still an open issue, or can it be considered addressed?

(c) gen_case_constraints_docs.py - two separate dedups:

The level emoji map was duplicated in render_playbook_card (line 159) and generate_playbook (line 267). Moved it to a LEVEL_EMOJI constant, placed after PlaybookEntry, and pointed both call sites at it.

get_model_name, get_riemann_solver_name, and get_time_stepper_name (lines 133-151) had byte-identical bodies modulo the schema key string. Replaced all three with a single _named(param, value) helper and updated the three call sites: line 170 (model_eqns), line 222 (riemann_solver), and line 226 (time_stepper).

(d) sched.py - the seconds were hardcoded at each comparison in notify_long_running_threads, while only the message text indexed the HEADLESS_THRESHOLDS table. Comparisons now read the seconds from the table too.

(e) user_guide.py - all 8 entries in ORG_COLORS mapped to "yellow", and the lookup defaulted to "yellow" anyway, so every org rendered yellow regardless. Deleted the dictionary and inlined "yellow" directly at the lookup site. Chose not to give orgs distinct colors, since that's a user-facing change and I could not identify what information/benefit color-coding would add across the 9 orgs. May be worth its own feature-request issue if that's actually wanted.

All changes are mechanical and behavior-preserving, no functional or output changes are intended.

### Type of change

- Refactor

## Testing

These changes are cleanups in the toolchain Python files, with no build-system or Fortran impact, so I verified behavior preservation directly rather than through the test suite. For two of the files, gen_case_constraints_docs.py and user_guide.py, I compared generated output before and after my changes. To do this I created a temporary copy of the original master branch alongside my working copy, so I could run both versions side by side. sched.py is time-dependent, so I verified it by value instead.

gen_case_constraints_docs.py - I ran main(), the entry point that generates the case constraint documentation, and compared outputs from both the original file and the file with changes. Outputs were the same. The path covered both changes in this file: it calls render_playbook_card(), which uses _named, and both it and generate_playbook() use LEVEL_EMOJI.

user_guide.py - same comparison on _generate_clusters_content(). Outputs were identical, and all orgs still render yellow.

sched.py - prints only after minutes of real elapsed time, so I compared values rather than output. The table holds 120, 600, and 1800 seconds, the same numbers as the 2 * 60, 10 * 60, 30 * 60 literals they replaced.

I also used file hashes (a way to confirm two files are byte-for-byte identical) as an extra check on top of the direct comparisons for gen_case_constraints_docs.py outputs and user_guide.py outputs where both matched.

<details>
<summary>Commands used for testing</summary>

​```bash
cd ~/MFC
mkdir -p ~/mfc-verify
git worktree add /tmp/original_master master
​```

**1. gen_case_constraints_docs.py**

​```bash
cd /tmp/original_master && python3 -c 'import sys; sys.path.insert(0,"toolchain")
from mfc.gen_case_constraints_docs import main; main()' > ~/mfc-verify/docs_before.txt

cd ~/MFC && python3 -c 'import sys; sys.path.insert(0,"toolchain")
from mfc.gen_case_constraints_docs import main; main()' > ~/mfc-verify/docs_after.txt

diff ~/mfc-verify/docs_before.txt ~/mfc-verify/docs_after.txt && echo "gen_case_constraints_docs.py: NO CHANGE"
​```

**2. user_guide.py**

​```bash
cd /tmp/original_master && python3 -c 'import sys; sys.path.insert(0,"toolchain")
from mfc.user_guide import _generate_clusters_content as f; print(f())' > ~/mfc-verify/guide_before.txt

cd ~/MFC && python3 -c 'import sys; sys.path.insert(0,"toolchain")
from mfc.user_guide import _generate_clusters_content as f; print(f())' > ~/mfc-verify/guide_after.txt

diff ~/mfc-verify/guide_before.txt ~/mfc-verify/guide_after.txt && echo "user_guide.py: NO CHANGE"
​```

**3. sched.py**

​```bash
cd ~/MFC && python3 -c 'import sys; sys.path.insert(0,"toolchain")
from mfc.sched import HEADLESS_THRESHOLDS as H
print(H[0][0], H[1][0], H[2][0], "should be", 2*60, 10*60, 30*60)'
​```
Result:
​```
120 600 1800 should be 120 600 1800
​```

**Proof (hashes)**

​```bash
shasum ~/mfc-verify/*.txt
​```
Result:
​```
3016ca54d8cafc7fed2e41f49ac3fd4dd2cababd  ~/mfc-verify/docs_after.txt
3016ca54d8cafc7fed2e41f49ac3fd4dd2cababd  ~/mfc-verify/docs_before.txt
a22ed36dc99d289c79f9c9dae2486fe2b5a5f5d6  ~/mfc-verify/guide_after.txt
a22ed36dc99d289c79f9c9dae2486fe2b5a5f5d6  ~/mfc-verify/guide_before.txt
​```

**Cleanup**

​```bash
cd ~/MFC && git worktree remove /tmp/original_master
​```

</details>

Part of #1512


sbryngelson (commented):
Thanks for the careful writeup — the before/after output diffing with matching hashes is more verification than most refactor PRs get, and it's the right way to justify a "no behavior change" claim. Note that green CI doesn't mean much here: no workflow exercises these three toolchain modules, so your manual comparison is what's actually backing this.

All three changes are correct and behavior-preserving. Comments below are about whether each is the *right* fix, not whether it works.

### `gen_case_constraints_docs.py` — good, this part is ready

Both dedups are real. `get_value_label(param, value)` is already generic, so the three getters were byte-identical modulo the key string, and collapsing them is exactly right. Confirmed there were no callers of `get_model_name` / `get_riemann_solver_name` / `get_time_stepper_name` outside this file, so dropping those names is safe.

Two asks:

1. `_named` is too vague — `_named('model_eqns', ...)` reads worse at the call site than `get_model_name(...)` did. The only thing the helper adds over `get_value_label` is the `"Not specified"` fallback, so name it for that: `_value_label_or_default`, or `_schema_name`. Also drop the enumeration from the docstring ("i.e. model name, Riemann solver name, time stepper name") — that rots as soon as there's a fourth call site.

2. You missed a third copy of the level list. `generate_playbook()` still has:

   ```python
   for level in ["Beginner", "Intermediate", "Advanced"]:
   ```

   Now that `LEVEL_EMOJI` is an insertion-ordered dict with exactly those keys, that should be `for level in LEVEL_EMOJI:`. That removes the last duplication and makes the constant the single source of truth, which is the point of the change.

### `sched.py` — please take this further or drop it

The motivation is legitimate: the table drove only the message text (`"(>2min)"`) while the comparisons used separate `2 * 60` literals, so editing the table would silently desync the threshold from what it prints.

But this removes one of three duplications and costs readability doing it. `elapsed >= HEADLESS_THRESHOLDS[0][0]` is harder to read than `elapsed >= 2 * 60`, and the structural duplication is untouched: three unrolled blocks, three `notified_2m` / `notified_10m` / `notified_30m` fields on `WorkerThreadHolder`, and three `# 2 minutes`-style comments that are now themselves stale-able copies of the table. The index-to-flag coupling (`[0]` ↔ `notified_2m`) stays hardcoded and unenforced — insert or reorder a threshold and the flags silently mismatch, which is the same failure mode you set out to fix.

What (d) is actually reaching for is a loop over the table:

```python
for i, (secs, msg) in enumerate(HEADLESS_THRESHOLDS):
    if i not in holder.notified_headless and elapsed >= secs:
        cons.print(f"  {msg} [bold magenta]{case_uuid}[/bold magenta]  {case_trace}")
        holder.notified_headless.add(i)
```

with the three booleans collapsed into a single `notified_headless: set` field on the holder. That kills all three duplications and makes adding a fourth milestone a one-line table edit. Either do that, or revert this file — the current middle ground isn't worth the readability hit.

### `user_guide.py` — answering your question

Good catch on the dead lookup, and thanks for asking instead of guessing. Deleting it: go ahead. `ORG_COLORS` was aspirational and never wired up — the missing `CSCS` key while `ORG_ORDER` has it is the tell that nobody was maintaining it, precisely because it never affected output. If we want per-org colors later it's a two-line addition, and it should come with an actual rationale for what the colors encode. Your instinct not to invent one was right.

### On (a), `fp_stability.py`

You're correct that the file moved out from under the issue — the cited line numbers are stale. Leave it out of this PR; I'll re-scope that item on #1512 separately. Don't block on it.

---

Once the two `gen_case_constraints_docs.py` items are in and `sched.py` goes one way or the other, this is good to merge.

PR #1599, opened 2026-06-15
fix: extend patch validation loops to num_patches_max

SVS87:
## Description

The alteration rights, smoothing, and primitive variable validation loops were bounded by num_patches, making the inactive-patch else branches unreachable. Changed all three loop bounds to num_patches_max to match the geometry loop, so inactive patch slots are validated across all four passes.

Closes #1482

### Type of change (delete unused ones)

- Bug fix

## Testing

Ran ./mfc.sh test -j 4 locally on Ubuntu/WSL with GNU compiler and MPI. 578 tests passed, 0 failed. The fix is output-neutral for valid cases — the newly-reached inactive-patch checks only abort on misconfigured inputs that no passing test exercises.

## AI code reviews

Reviews are not retriggered automatically. To request a review, comment on the PR:
- `@claude full review` — Claude full review (also triggers on PR open/reopen/ready)
- Or add label `claude-full-review` — Claude full review via label

PR #1727, opened 2026-08-14
Run inactive-patch validation over all patch slots

### LINKED ISSUE #1482: Inactive-patch validation never runs: s_check_patches loops bound at num_patches instead of num_patches_max
In `src/pre_process/m_check_patches.fpp`, `s_check_patches` runs four per-patch validation passes. The first (geometry) loop correctly iterates over all *possible* patch slots and uses `i <= num_patches` to split active vs. inactive handling:

```fortran
! lines 42-94 — CORRECT
do i = 1, num_patches_max
    if (i <= num_patches) then
        ... active-patch geometry checks ...
    else
        @:PROHIBIT(patch_icpp(i)%geometry /= dflt_int, "Inactive patch defined. ...")
        call s_check_inactive_patch_geometry(i)
    end if
end do
```

The other three passes use the same `if (i <= num_patches) ... else ...` structure but loop only to `num_patches`:

```fortran
! lines 97-103 — alteration rights
do i = 1, num_patches
    if (i <= num_patches) then
        call s_check_active_patch_alteration_rights(i)
    else
        call s_check_inactive_patch_alteration_rights(i)   ! DEAD: i never exceeds num_patches
    end if
end do

! lines 106-115 — smoothing  (else -> s_check_unsupported_patch_smoothing)
! lines 118-124 — primitive variables (else -> s_check_inactive_patch_primitive_variables, DEAD)
```

Because `i` only ranges up to `num_patches`, the condition `i <= num_patches` is always true and every `else` branch is unreachable.

**This is a latent validation gap, not merely cosmetic dead code.** The inactive-check routines perform real validation that inactive patches (slots `num_patches+1 .. num_patches_max`) were not accidentally configured — e.g.:
- `s_check_inactive_patch_alteration_rights` (lines 404-415): forbids `alter_patch(0)` being cleared or any `alter_patch(1:)` enabled on inactive patches.
- `s_check_inactive_patch_primitive_variables` (lines 511-526): forbids `alpha_rho`, `rho`, `vel`, `pres`, `alpha`, `gamma`, `pi_inf` being set on inactive patches.

As written, these checks run only for the geometry pass, so a case that sets, say, `alter_patch` or primitive variables on a patch index beyond `num_patches` will silently pass validation instead of being rejected.

**Fix.** Change the loop bound from `num_patches` to `num_patches_max` in the three loops (lines 97, 106, 118), matching the geometry loop. Verify the smoothing loop's `else` predicate is correct for inactive patches (an inactive patch has `geometry == dflt_int`, so it falls into the `else`/unsupported branch as intended).

**Output-neutral for valid cases.** The newly-reached checks only abort on *misconfigured* inactive patches, which no passing test exercises. It tightens validation rather than changing solver output.

---
<sub>Filed from a repo-wide code-cleanliness review; locations verified against `master` @ `40dde5e`.</sub>

## Code references
- [`src/pre_process/m_check_patches.fpp:42-94`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/src/pre_process/m_check_patches.fpp#L42-L94) — geometry loop — correct bound
- [`src/pre_process/m_check_patches.fpp:97-124`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/src/pre_process/m_check_patches.fpp#L97-L124) — three loops bound at num_patches (dead else)
- [`src/pre_process/m_check_patches.fpp:404-415`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/src/pre_process/m_check_patches.fpp#L404-L415) — inactive alteration checks
- [`src/pre_process/m_check_patches.fpp:511-526`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/src/pre_process/m_check_patches.fpp#L511-L526) — inactive primitive-var checks

Mohit-Ak:
Three of the four per-patch validation loops in `s_check_patches` were bounded at
`num_patches` instead of `num_patches_max`, so their `else` branches — the ones
that validate the *inactive* patch slots — could never be reached.

The first loop (alteration of deprecated geometry numbers, L43) is already
correct: it runs `do i = 1, num_patches_max` and uses `if (i <= num_patches)` to
split active from inactive slots. The other three copy that same
`if (i <= num_patches) ... else ...` shape but only ever iterate to
`num_patches`, which makes the predicate unconditionally true. That left three
validators as dead code:

| loop | validator that was unreachable |
|---|---|
| L98 — alteration rights | `s_check_inactive_patch_alteration_rights` (L405) |
| L107 — smoothing | `s_check_unsupported_patch_smoothing` (L442) |
| L119 — primitive variables | `s_check_inactive_patch_primitive_variables` (L512) |

The fix widens those three bounds to `num_patches_max`. That's the whole change,
+3/-3 in one file.

The smoothing loop is worth a second look because it's structurally different —
its predicate is a geometry whitelist (`i > 1 .and. geometry == 2/3/4/...`)
rather than `i <= num_patches`. An inactive slot has `geometry == dflt_int`, so
once the bound is widened it falls through to the `else`/unsupported branch,
which is the intended behaviour. I verified that empirically rather than by
reading (see the `smooth` row below).

## Why this can't change results for any valid case

`m_global_parameters.fpp:230-268` initialises all `num_patches_max` (= 10, from
`m_constants.fpp:26`) slots to defaults in its own `do i = 1, num_patches_max`
loop — `alter_patch = .false.` with `alter_patch(0) = .true.`, `smoothen =
.false.`, `smooth_patch_id = i`, and `dflt_real` for every primitive variable.
Those defaults are exactly what the inactive-patch validators assert. So for a
well-formed case the newly-reachable checks are no-ops; the change only tightens
validation on misconfigured input and cannot move a golden file.

## How it was tested

I built a 1D Sod-like case with `num_patches: 2` and then set one forbidden
parameter on the *inactive* slot 3, one variant per broken loop, plus a clean
control. Run through `./mfc.sh run <case.json> -t pre_process -n 1`:

| variant | what it sets on patch 3 | before | after |
|---|---|---|---|
| `control` | nothing (valid case) | exit 0 | exit 0 |
| `alter` | `alter_patch(1) = T` | exit 0 — wrongly accepted | aborts: `Inactive patch 3: cannot have any alter_patch(i) enabled` |
| `smooth` | `smoothen = T`, `smooth_coeff = 0.5` | exit 0 — wrongly accepted | aborts: `Inactive patch 3: cannot have smoothen enabled` |
| `prim` | `pres`, `vel(1)`, `rho` | exit 0 — wrongly accepted | aborts: `Inactive patch 3: rho must not be set` |

Each bad variant produces a *different* message, which is what proves the three
loops independently rather than just showing that some check fires. The control
still passes in both directions, so nothing valid became newly rejected.

Both CI gates pass locally on the committed tree:

- `./mfc.sh lint` → `372 passed, 4 subtests passed in 9.74s`
- `./mfc.sh precheck -j 4` → all 7 checks OK (formatting, spelling, toolchain
  lint, source lint, doc refs, parameter docs, example cases)

## Note on prior work

PR #1599 took the same approach and was closed for author inactivity, not
because anything was wrong with it. This is a clean re-do with the reachability
matrix above as evidence.

Fixes #1482

### DIFF
--- a/src/pre_process/m_check_patches.fpp
+++ b/src/pre_process/m_check_patches.fpp
@@ -95,7 +95,7 @@ contains
         end do
 
         ! Constraints on overwrite rights initial condition patch parameters
-        do i = 1, num_patches
+        do i = 1, num_patches_max
             if (i <= num_patches) then
                 call s_check_active_patch_alteration_rights(i)
             else
@@ -104,7 +104,7 @@ contains
         end do
 
         ! Constraints on smoothing initial condition patch parameters
-        do i = 1, num_patches
+        do i = 1, num_patches_max
             if (i > 1 .and. (patch_icpp(i)%geometry == 2 .or. patch_icpp(i)%geometry == 3 .or. patch_icpp(i)%geometry == 4 &
                 & .or. patch_icpp(i)%geometry == 5 .or. patch_icpp(i)%geometry == 8 .or. patch_icpp(i)%geometry == 9 &
                 & .or. patch_icpp(i)%geometry == 10 .or. patch_icpp(i)%geometry == 11 .or. patch_icpp(i)%geometry == 12 &
@@ -116,7 +116,7 @@ contains
         end do
 
         ! Constraints on flow variables initial condition patch parameters
-        do i = 1, num_patches
+        do i = 1, num_patches_max
             if (i <= num_patches) then
                 call s_check_active_patch_primitive_variables(i)
             else


Mohit-Ak:
Closing: the Fortran fix is correct but causes widespread pre_process failures across the existing test suite — many test cases inherit a 3-patch BASE_CFG and only override num_patches to 1 or 2, leaving inactive slots with non-default values that the now-reachable validators reject. The fix needs a coordinated update to the test-case generator alongside the Fortran change. Will revisit with a complete fix.

PR #1732, opened 2026-08-18
ci(coverage): key map freshness on the refresh run, not on _meta.git_sha

sbryngelson:
## Description

The daily **Coverage Map Health** cron has failed every day since 2026-08-16 ([latest run](https://github.com/MFlowCode/MFC/actions/runs/32110851684/job/95629747994)) with:

```
Coverage map is STALE: built before the most recent coverage-relevant commit. Refresh workflow may be broken.
```

The refresh workflow is not broken. #1683 added two things in one commit that contradict each other:

1. `coverage_map_changed.py` — the refresh pushes a commit only when the rebuilt coverage *entries* differ; `_meta` (`built_at`, `git_sha`) is deliberately excluded from the comparison, so a no-op refresh pushes nothing.
2. `built_after_last_change()` — health requires the map's `_meta.git_sha` to descend from the last commit touching `src/**/*.fpp` or `toolchain/mfc/test/cases.py`.

`_meta.git_sha` only advances when a commit lands. So the first coverage-relevant commit whose coverage happens to be unchanged pins `git_sha` in the past **permanently**, and the health check reports a working refresh as broken. Nothing self-heals it; only a later commit that actually moves the entries would.

The committed map is at `git_sha: ae4b4c40` (#1414, Aug 11). Four coverage-relevant commits landed after it, each with a refresh run that printed `Coverage entries unchanged (675 tests) -> nothing to push`:

| Commit | Refresh run |
|---|---|
| `bb29773b7` (#1718) | [31889075305](https://github.com/MFlowCode/MFC/actions/runs/31889075305) ✅ unchanged |
| `3db0bd175` (#1722) | [31956587627](https://github.com/MFlowCode/MFC/actions/runs/31956587627) ✅ unchanged |
| `b7d788384` (#1686) | [31964706234](https://github.com/MFlowCode/MFC/actions/runs/31964706234) ✅ unchanged |
| `ca26c4cd4` (#1721) | [31986971385](https://github.com/MFlowCode/MFC/actions/runs/31986971385) ✅ unchanged |

(The `ModuleNotFoundError: No module named 'cantera'` tracebacks in the health log are not the cause of the STALE failure — those cases are printed under `Note: 16 case(s) could not be loaded` and are excluded from the check. They are a separate real bug, fixed below.)

### The fix

Record the freshness signal where it can actually be observed — on the refresh **run**, not in the map's content:

- `coverage-refresh.yml` moves `refs/coverage-map/verified` to the commit it rebuilt from on **every** successful refresh, in both branches of the commit guard. The push is non-fatal: the map is already pushed by that point, so failing this bookkeeping step must not red a refresh that worked — and it is not silent, because a ref that stops advancing drops the health check back to its wall-clock rule, which goes red within `MAX_AGE_DAYS`.
- `check_coverage_map_health.py` asks whether *that ref* descends from the last coverage-relevant commit. A missing ref (fork, or the window before the first refresh after this lands) reads as undeterminable and falls back to the wall-clock age rule, exactly as a shallow clone already does.
- The ref lives outside `refs/heads/` and `refs/tags/` on purpose: updating it fires no `push` workflow trigger. A branch or tag would start `homebrew.yml`, which filters on `paths:` alone with no ref filter.

### Also: the `cantera` failures are an interpreter bug, not a missing install

The existing code comment says those optional deps are "not installed in this lightweight job". That is wrong — every health run's init step logs `Downloading cantera (18.4MiB)` / `Installed 96 packages`.

`common.get_py_program_output` ran case files with a hardcoded `"python3"` taken from PATH. `coverage-health.yml` invokes `build/venv/bin/python3` **directly** instead of activating the venv, so PATH still resolved to the setup-python interpreter, which has no cantera. Every other entry point goes through `./mfc.sh`, which activates the venv and makes PATH's `python3` the venv's — which is why only this job ever saw it.

Case files now run under `sys.executable`. Under an activated venv that is the same binary, so no existing flow changes; it differs only where the venv interpreter is invoked by path.

This closes a gap rather than just quieting logs: those 16 chemistry cases were dropped from the `MIN_FRACTION` denominator, so coverage rot in the chemistry tests was invisible to the health check.

### Also: the refresh timeout

Refresh run [32095131563](https://github.com/MFlowCode/MFC/actions/runs/32095131563) (push of #1717, which changed `cases.py`) was **killed at 4h05m** by `timeout-minutes: 240`. Wall clock there is dominated by the phoenix SLURM queue, not the build — observed waits reach 6h03m. Raised to 480. That timeout is why the map is now also behind the current test list, independent of the bug above.

### Type of change

- Bug fix

## Testing

`ruff format --check`, `ruff check`, `typos`, and the full toolchain unit suite (`422 passed`) pass locally.

New unit tests in `test_coverage_unit.py` drive the two git predicates against a throwaway repository:

- `test_verified_sha_is_none_when_the_ref_was_never_pushed` — absent ref is undeterminable, not broken
- `test_verified_after_last_change_true_when_a_refresh_ran_since_the_change` — including a refresh *at* the change
- `test_verified_after_last_change_false_when_the_refresh_predates_the_change` — the genuine broken-refresh case
- `test_a_no_op_refresh_still_keeps_the_map_healthy` — the regression this PR fixes: an ancient `_meta.git_sha` with a current ref must read healthy

`check_coverage_map_health.py` gained an `if __name__ == "__main__":` guard so its predicates are importable; it was previously untestable top-level script code.

Two tests in the new `toolchain/mfc/test_common.py` cover the interpreter fix, both verified to fail against the previous implementation:

- `test_case_files_run_under_the_toolchain_interpreter` — the case file reports the interpreter that ran it
- `test_a_hostile_python3_on_path_is_not_consulted` — PATH is poisoned with a `python3` that exits 3, so the test cannot pass by coincidence on a machine where the two interpreters happen to match

**After merging:** dispatch `Coverage Map Refresh` once by hand to create `refs/coverage-map/verified`. Until it exists the health check falls back to the 10-day age rule, and the current map (built 2026-08-11) crosses that on 2026-08-21 — before the Monday cron. That run also brings the map up to date with #1717's test-list changes, which the timeout above prevented.

## Checklist

### DIFF
--- a/.github/scripts/check_coverage_map_health.py
+++ b/.github/scripts/check_coverage_map_health.py
@@ -11,54 +11,90 @@ from mfc.test.cases import list_cases  # noqa: E402  (returns the current test l
 MAX_AGE_DAYS = 10
 MIN_FRACTION = 0.80
 
+# coverage-refresh.yml moves this ref to the commit it just rebuilt the map from, on EVERY
+# successful refresh -- including the ones whose entries come out identical and therefore
+# push no commit at all.
+#
+# The map's own _meta.git_sha cannot answer the freshness question. It only advances when a
+# commit lands, and the refresh deliberately skips the commit when the coverage entries are
+# unchanged (a no-op commit on every push is what #1683 removed). So a coverage-relevant
+# commit whose coverage happens to be identical pinned git_sha in the past and made this
+# check fail every day thereafter, with no way to recover: a refresh that WAS working
+# looked broken. The question here is "has a refresh run since the last relevant commit",
+# and only the refresh itself can answer it.
+#
+# It lives outside refs/heads/ and refs/tags/ so that updating it fires no `push` workflow
+# trigger. actions/checkout does not fetch it; coverage-health.yml fetches it explicitly.
+VERIFIED_REF = "refs/coverage-map/verified"
+
 # Must mirror the `paths:` trigger of coverage-refresh.yml. A superset would raise false
 # alarms (a change that never triggers a refresh would look like a missed refresh).
 COVERAGE_RELEVANT_PATHS = [":(glob)src/**/*.fpp", "toolchain/mfc/test/cases.py"]
 
 
-def built_after_last_change(git_sha):
-    """Was the map built at or after the last coverage-relevant commit? None if unknowable.
+def verified_sha(cwd=None):
+    """Commit the last successful refresh verified the map against, or None if unknown.
+
+    None covers both "the ref was never pushed" (a fork, or the window before the first
+    refresh after VERIFIED_REF was introduced) and "the fetch did not bring it down". The
+    caller must read that as undeterminable and fall back to the wall-clock age rule, not
+    as a failure -- an absent ref is not evidence of a broken refresh.
+    """
+    rev = subprocess.run(["git", "rev-parse", "--verify", "--quiet", f"{VERIFIED_REF}^{{commit}}"], capture_output=True, text=True, check=False, cwd=cwd)
+    return rev.stdout.strip() or None
+
+
+def verified_after_last_change(git_sha, cwd=None):
+    """Did that refresh run at or after the last coverage-relevant commit? None if unknowable.
 
     This is the direct question the wall-clock age check only approximated: it stays quiet
-    over a genuinely quiet repo (no refresh needed, so no heartbeat commit needed either)
-    and fires on the next relevant push after a refresh breaks, rather than 10 days later.
+    over a genuinely quiet repo (no refresh needed, so no heartbeat needed either) and
+    fires on the next relevant push after a refresh breaks, rather than 10 days later.
     """
     if not git_sha:
         return None
-    last = subprocess.run(["git", "log", "-1", "--format=%H", "--", *COVERAGE_RELEVANT_PATHS], capture_output=True, text=True, check=False)
+    last = subprocess.run(["git", "log", "-1", "--format=%H", "--", *COVERAGE_RELEVANT_PATHS], capture_output=True, text=True, check=False, cwd=cwd)
     if last.returncode != 0 or not last.stdout.strip():
         return None  # shallow clone or no such commit -> fall back to the age rule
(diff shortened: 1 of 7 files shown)


sbryngelson:
Both review comments applied in 8a004dd, plus the root cause of the `cantera` noise.

**1. `$GITHUB_SHA` instead of `HEAD`** — agreed. Both satisfy the ancestry test (the bot commit is a child of $GITHUB_SHA), but only $GITHUB_SHA names the same thing in both branches of the guard: the source state the map was built against. `HEAD` silently meant a different commit depending on whether the entries moved.

**2. Stale comment above `actions/checkout`** — updated; it still explained the full-history requirement in terms of `_meta.git_sha`.

**3. The `cantera` errors were not a missing install.** I had described them as unrelated noise; that was only half right — they are unrelated to the STALE failure, but they are a real bug, and the existing code comment about optional deps "not installed in this lightweight job" is wrong. Every health run's init step logs:

```
Downloading cantera (18.4MiB)
 Downloaded cantera
Installed 96 packages in 2.24s
```

`common.get_py_program_output` ran case files with a hardcoded `"python3"` from PATH. `coverage-health.yml` invokes `build/venv/bin/python3` **directly** rather than activating the venv, so PATH still resolved to the setup-python interpreter, which has no cantera. Every other entry point goes through `./mfc.sh`, which activates the venv and makes PATH's `python3` the venv's — which is why only this job ever saw it.

Fixed at the source: case files now run under `sys.executable`. Under an activated venv that is the same binary, so no existing flow changes; it differs only where the venv interpreter is invoked by path, which is the broken case.

This closes a real gap rather than just quieting logs — those 16 chemistry cases were dropped from the `MIN_FRACTION` denominator, so coverage rot in the chemistry tests was invisible to the health check.

Both new tests in `toolchain/mfc/test_common.py` were verified to fail against the previous implementation and pass against this one; the second poisons PATH with a `python3` that exits 3, so it cannot pass by coincidence on a machine where the two interpreters happen to match. Full suite: 424 passed.

PR #1736, opened 2026-08-18
1536 ib neighborhood radius automatic size checking

### LINKED ISSUE #1536: ib_neighborhood_radius automatic size checking
A new case file parameter was recently added called `ib_neighborhood_radius`. It is responsible for making sure that the ib neighborhood size is large enough so that all IB patches are drawn fully. For now, it is set manually by the user, but this can cause some failures and confusion that are best to be avoided. 

To solve this, the neighborhood radius should now default to 0. If it is set to 0, then the code should initiate some sort of automatic size estimator. For each patch type, a maximum physical distance from the centroid should be computed. Then the maximum of that is the physical size of the neighborhood radius. That value should then be converted into an integer neighborhood radius that can be communicated to all processors. This way users have the option to let the code automatically handle this computation for them. They can also choose to set it manually if they want to avoid that start-up time or if they know that they can get away with something smaller than the theoretical limit.

Alternatively, it may be optimal to completely disable all of the IB communication when the neighborhood size is not set, and all particles can simply exist on all ranks. Very clean solution.

danieljvickers:
## Description

Immersed Boundary cases have one extra layer of tediousness due to the IB neighborhood radius needing ot be recomputed based upon the number of ranks. This causes significant problems when attempting to run the same case on  CPUs vs. GPUs (because one may use more CPUs for the same case) and in strong scaling scenarios.

To address this, this PR allows users to let MFC automatically compute the IB Neighborhood radius. The code computes the largest width required for any IB in the simulation and the largest processor width. These are used to approximate a size, with an additional 10% safety factor on the neighborhood size for safety.

This is not ideal, as it does not consider things like the orientation of the IB for reducing the size further. If particle forces are desired, it is always recommended to compute the IB neighborhood radius that one desires. However, for uniform distributions of small IBs, this will often produce optimal results with no modification to the case file.

To enable automatic radius checking, just set `ib_neighborhood_radius: 0` in the case file, which is also the new default. 

Closes #1536 

- New feature

## Testing

How did you test your changes?

2D ranks on GNU for increasing IB size.

## Checklist

<details>

</details>

## AI code reviews


danieljvickers:
Addressed AI comments. Comments 1 and 3 are valid. Correction 3 is not a concern on its own, but it helped me notice a nearby and more insidious issue with taking a floor that should have really been a ceiling. The airfoil response is sound.

The STL concern is nonsense, as the flags are for two different STL models. We should unify these, but it is still nonsense as written. I have ignored it..

PR #1754, opened 2026-08-23
test: smoke-test the LOC diff bot (do not merge)

sbryngelson:
Throwaway PR to verify the LOC diff bot from #1753 actually posts. **Do not merge — this will be closed.**

The change splits one `integer :: i, j, k` declaration into two lines: semantically identical, +1 counted line.

Expected sequence:
1. This push -> a sticky comment showing `m_global_parameters.fpp  744  +1` and `**total** 45478 +1`.
2. A follow-up push reverting it -> the comment is deleted (zero diff, nothing to say).

Then the PR gets closed.

PR #1755, opened 2026-08-23
test: smoke-test the LOC bot from a fork (do not merge)

sbryngelson:
Second smoke test for the bot from #1753, this time **from a fork** — the case #1754 could not cover. **Do not merge — will be closed.**

Two things under test:

1. **Fork PRs get commented at all.** This is the entire reason the workflow uses `pull_request_target`; under `pull_request` a fork's `GITHUB_TOKEN` is read-only and the comment would fail. It also exercises fetching `refs/pull/N/head` for a fork head.

2. **Merge-base counting.** This fork's master is 32 commits stale, and its `src/` tree differs from upstream master by **-1403 lines**. Counting against master's *tip* would blame this PR for roughly -1403 lines of other people's deletions. Counting against the merge base should report **+1** — the one declaration line this PR actually splits.

Expected comment: `m_global_parameters.fpp +1`, `**total** +1`. Anything near -1403 means the merge-base logic is broken.

### DIFF
--- a/src/simulation/m_global_parameters.fpp
+++ b/src/simulation/m_global_parameters.fpp
@@ -681,7 +681,8 @@ contains
     !> Initialize the global parameters module
     impure subroutine s_initialize_global_parameters_module
 
-        integer :: i, j, k
+        integer :: i, j
+        integer :: k
         integer :: fac
 
         #:if not MFC_CASE_OPTIMIZATION


sbryngelson:
Verified: fork PRs get commented (pull_request_target working as intended), and the merge-base counting reported +1 rather than the ~-1403 that comparing against master's tip would have produced on this 32-commit-stale fork. Closing — not for merge.

PR #1759, opened 2026-08-25
Work around amdflang whole-image device codegen instability

sbryngelson:
## What

Adds `-attributor-max-pi-accesses=16384` to the offload link for LLVMFlang OpenMP-offload builds, and documents two amdflang-specific traps in `gpuParallelization.md`.

## Why

amdflang generates device code for the whole image at link time. Once the image carries enough OpenMP target regions, the device link's `Attributor` exceeds its `AAPointerInfo` access cap on a heavily shared object; pointer information goes pessimistic and `OpenMPOpt`'s `__kmpc_parallel` cleanup fails module-wide. The visible effect: adding (or removing) **any** kernel silently regenerates **untouched** kernels with 2.4–4.5× worse ISA — register spills and +512 B LDS in every kernel, image-wide. Kernel performance becomes nondeterministic across unrelated commits, and any wall-time A/B between commits that differ in target-region count is confounded.

Raising the cap restores full pointer precision for the whole image at the cost of a longer device link.

## Evidence

- Deterministic from source; partition-independent (`-flto-partitions=1` reproduces); reproduces on two AFAR drops sharing flang 23.0.0git.
- In-link A/B on identical objects: default cap → slow-class ISA (e.g. WENO scratch 28→140 B, uniform LDS bump); cap 16384 (or 0) → fast-class ISA restored.
- A full solver run with the flag: byte-exact MPI communication totals and wall parity vs an unflagged fast-class build.
- A minimal reproducer does not trigger it — the ~1000-kernel image is required — so the workaround is applied at the build level.

Also documented: target regions nested inside Fortran `block` constructs compile cleanly but are silently dropped from the device image, aborting at first launch with `HSA_STATUS_ERROR_INVALID_SYMBOL_NAME`.

## Status

Draft until the flagged amdflang gfx90a build re-verification completes on our cluster (the flag itself has been verified via env-injected `FFLAGS`; this PR moves it into the build system). A compiler bug report to AMD is being prepared separately.

### DIFF
--- a/cmake/MFCTargets.cmake
+++ b/cmake/MFCTargets.cmake
@@ -203,7 +203,19 @@ exit 0
                         -fopenmp-assume-threads-oversubscription
                         -fopenmp-assume-teams-oversubscription
                         -fopenmp-assume-no-nested-parallelism)
-                    target_link_options(${a_target} PRIVATE -fopenmp --offload-arch=gfx90a -flto-partitions=${MFC_BUILD_JOBS})
+                    # attributor-max-pi-accesses: amdflang generates device code for the WHOLE
+                    # image at link time, and once the image carries enough target regions the
+                    # device link's Attributor exceeds its AAPointerInfo access cap on a
+                    # heavily-shared object. Pointer information then goes pessimistic and
+                    # OpenMPOpt's __kmpc_parallel cleanup fails module-wide: UNTOUCHED kernels
+                    # regenerate with 2.4-4.5x worse ISA (register spills, +512 B LDS in every
+                    # kernel) whenever ANY kernel is added or removed anywhere in the code.
+                    # Raising the cap restores full pointer precision for the whole image and
+                    # makes kernel quality independent of unrelated edits, at the price of a
+                    # longer device link. See docs/documentation/gpuParallelization.md
+                    # ("AMD flang known issues") for the failure signature.
+                    target_link_options(${a_target} PRIVATE -fopenmp --offload-arch=gfx90a -flto-partitions=${MFC_BUILD_JOBS}
+                        "SHELL:-Xoffload-linker -mllvm -Xoffload-linker -attributor-max-pi-accesses=16384")
                 endif()
             endif()
 
--- a/docs/documentation/gpuParallelization.md
+++ b/docs/documentation/gpuParallelization.md
@@ -829,6 +829,42 @@ LIBOMPTARGET_JIT_SKIP_OPT=1
 - If set, the image will only be passed through the backend.
 - The backend is invoked with the `LIBOMPTARGET_JIT_OPT_LEVEL` flag.
 
+## AMD flang (amdflang) Known Issues
+
+### Whole-image device codegen instability (worked around in the build)
+
+amdflang generates device code for the whole image at link time. Once the image carries
+enough OpenMP target regions, the device link's `Attributor` pass exceeds its
+`AAPointerInfo` access cap on a heavily shared object; pointer information goes
+pessimistic and `OpenMPOpt`'s `__kmpc_parallel` cleanup then fails for the whole module.
+The visible effect: adding (or removing) ANY kernel anywhere silently regenerates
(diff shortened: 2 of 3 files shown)


sbryngelson:
Independent data that may support the image-wide diagnosis here.

While measuring #1762 on an MI210 (gfx90a, AFAR amdflang `therock-afar-23.2.1`, OpenMP offload, **without** this PR's flag), I dumped static kernel resources from the code object for all 471 kernels. Five unrelated kernels share a byte-identical signature:

| kernel | scratch | VGPR | AGPR |
|---|---:|---:|---:|
| `m_viscous::compute_viscous_stress_cylindrical_boundary` | 17760 B | 400 | 144 |
| `m_boundary_common::populate_beta_bc_direction` | 3864 B | 400 | 144 |
| `m_qbmm::mom_inv` | 3648 B | 400 | 144 |
| `m_bubbles_el::enforce_el_bubbles_boundary_conditions` | 2440 B | 400 | 144 |
| `m_bubbles_el::compute_bubble_el_dynamics` | 2376 B | 400 | 144 |

Five independent problems landing on exactly 400/144 seems unlikely; a common cause fits better, and the register counts are consistent with the pessimised pointer information described here. `m_riemann_solver_hlld` is separately at 21440 B / 470 VGPR / 214 AGPR — 684 registers against gfx90a's 512 unified per SIMD, i.e. one wave per SIMD.

If it is useful, these are reproducible with a small tool on branch `eos/gpu-guards` (`toolchain/mfc/kernel_resources.py`) that reads `.private_segment_fixed_size`/`.vgpr_count`/`.agpr_count` from the AMDGPU ELF inside `.llvm.offloading` and diffs two builds. Comparing before/after this PR's flag on those five kernels would confirm or rule out the shared cause quickly.

One caveat on my own numbers: they were taken with kernel counts held constant on both sides, precisely because of the effect this PR documents. Anyone diffing kernel resources across a change that adds or removes kernels should not trust the result.

sbryngelson:
The `Case Opt | Frontier (AMD) (gpu-omp)` failure here is a drop-version problem, not a problem with the workaround. `-attributor-max-pi-accesses` does not exist in the AFAR drop CI currently builds with:

```
ld.lld: error: -mllvm: ld.lld: Unknown command line argument '-attributor-max-pi-accesses=16384'.
ld.lld: Did you mean '--attributor-max-iterations=16384'?
```

CI is on `therock-23.1.0-gfx90a-7.12.0-bb5005b6`. That flag landed in the 23.2.x drops:

| drop / toolchain | has `attributor-max-pi-accesses` |
|---|---|
| `therock-23.1.0-…-7.12.0` (CI today) | no |
| `therock-afar-23.2.0-…-7.13.0` | **yes** |
| `therock-afar-23.2.1-…-7.13.0` | **yes** |
| `/opt/rocm-7.2.0`, `/opt/rocm-7.13.0` | no |

Side-by-side link of the same source with the same flags, only the drop differing:

```
=== therock-23.1.0-gfx90a-7.12.0-bb5005b6 ===
ld.lld: error: -mllvm: ld.lld: Unknown command line argument '-attributor-max-pi-accesses=16384'.
ld.lld: Did you mean '--attributor-max-iterations=16384'?
clang: error: ld.lld command failed with exit code 1
rc=1

=== therock-afar-23.2.1-gfx90a-7.13.0-7357b5084b ===
rc=0
```

That is the exact `-Xoffload-linker -mllvm -Xoffload-linker -attributor-max-pi-accesses=16384` spelling this PR adds, accepted end-to-end through the offload linker on 23.2.1.

This explains the gap in the PR description between "verified via env-injected `FFLAGS`" and CI failing: the verification ran against a 23.2.x drop, while CI is still on 23.1.0. Note both are flang 23.0.0git, so the version string alone doesn't distinguish them.

#1768 moves `famd` to `therock-afar-23.2.1`. Once that lands, rebasing this PR on top should clear the failure. Until then it will fail on the missing flag regardless of whether the workaround itself is correct.

sbryngelson:
drafting until #1768 merges

PR #1765, opened 2026-08-26
[CI probe — do not merge] up/mega x master integration

sbryngelson:
Integration probe for #1628: that PR is conflicting with master, so its CI matrix cannot run. This branch is up/mega with upstream master merged in (9 conflicted files resolved; the m_rhs resolution keeps the flat standard sweep path and hosts master's dual-pass HLLD on its vf fields, allocated only under hypo_nc_mode_dual_pass). Purpose: run the full compiler matrix on the merged code to surface cross-compiler issues before the real merge. Compiles with AMD flang GPU offload locally. Not intended to merge — the real integration lands through #1628.


sbryngelson:
Pushed the fix for the GNU lane's finding: `allocated()` on a `scalar_field%sf` POINTER member (gfortran rejects, AMD flang tolerated) — now `associated()`. Same fix landed on up/mega as c37dea4c. This is the probe working as intended.

PR #1802, opened 2026-09-02
File-Per-Process IB markers

danieljvickers:
For some reason, I never added file-per-process for IB markers, and it is a slowdown constaint for some of the IB cases we are running. Adding this here. I tested this on the `2D_mibm_shock_cylinder` case on 2 ranks and verified the visual output was correct.

Even though this is marked as "draft", it is ready for merge.


danieljvickers:
@sbryngelson Do you still want us to keep PRs as draft when they are done? This and the restart changes would both be useful for me to get merged. So it would be helpful to have them bumped up in terms of priority for merge if possible

sbryngelson:
noted. i'll prioritize it and move it out of draft soon enough

sbryngelson (approved):

sbryngelson (approved):

PR #1804, opened 2026-09-02
ci: take gpu-l40s out of rotation and seed frontier10202 as excluded

sbryngelson:
Two GPU resources accounted for most of the non-code failures in [run 33553417354](https://github.com/MFlowCode/MFC/actions/runs/33553417354). Both are addressed here; neither touches solver code.

## `gpu-l40s` out of rotation

It has been failing jobs for weeks, and it is also the partition the selector kept choosing and then starving on — four Phoenix lanes submitted, never started, cancelled after 3–5.5 h.

Two independent causes, both fixed:

1. **`mix` counted as available.** A `mix` node is partially allocated and may have no free GPU. CI logged `Selected GPU partition: gpu-l40s (1 idle/mix nodes)` and then queued until the job timeout. Only fully idle nodes count now.
2. **The fallback also pointed at l40s**, so removing it from the priority list alone would have sent jobs straight back whenever nothing was idle. Fallback is now `gpu-a100`.

Priority is now `gpu-h200 gpu-h100 gpu-a100 gpu-v100`.

## `frontier10202` seeded into the exclude list

That node produced **all 183** GPU memory-access faults in the run, across **43 distinct tests**; three exhausted their retries (`851F7AE2`, `12ECE133`, `05A8C23C`). The same lanes passed on `frontier10181/82/93/97/10203/06/07/10` with **zero** faults, and the same tests passed under `gpu-acc` on other nodes in that very run.

It is seeded into `node_exclude` exactly the way phoenix already seeds its two nodes, so the preflight added in #1797 can still append more at run time — this does not replace or bypass that mechanism.

Its faults are intermittent (379 of 382 tests still passed on it), which is why `syscheck` can clear the node and a static seed is still worth having. **Drop it once OLCF has looked at the node.**

## Not covered here

The third failure class in that run was the self-hosted runner `frontier-22` losing communication with GitHub (three jobs, no steps executed). That is a runner-host problem, not something the submit scripts can influence.

## Verification

Both scripts pass `bash -n`. `submit-slurm-job.sh` still renders exactly one `#SBATCH --exclude=` directive (the dynamic one) — the seed feeds it rather than competing with it.

Fallback choice is worth a second opinion: I picked `gpu-a100` without knowing which Phoenix partition has the most nodes.


sbryngelson:
Thanks — pushed 1341e445 for the two that hold up, and here is why I am not acting on the other two.

**Fixed: `^idle` over-matches.** Correct and important. `sinfo`'s `%t` suffixes a state to flag it (`*` not responding, `$` reserved for maintenance, `~` powered down), and this cluster emits them right now — `drain*`, `down*`, `alloc$`, `drain$` are all live in `sinfo` output as I write this. So `idle*` and `idle$` are reachable, and the unanchored match would have counted unusable nodes as available, reintroducing the exact starvation this PR removes. Now anchored `^idle$`.

**Fixed: stale `GPU_PARTITION_MIN_NODES` doc.** It still said the threshold counted "idle/mix" nodes, which stopped being true one commit earlier. Corrected.

**Not changing: `node_exclude=` overwrites a previously seeded value.** I checked, and there is nothing to overwrite. `node_exclude=""` is initialised once at line 38, then assigned only inside the mutually-exclusive `case "$cluster"` branches — phoenix at 147, frontier at 158 — and no environment or argument path pre-populates it. The additive path is later, at line 317, where a preflight fault merges a node in via `node-exclude.sh merge`. My line is identical in form to the phoenix seed directly above it; if it were unsafe, that one would be too.

**Not changing: counting only idle under-reports availability.** This is a real trade-off rather than an oversight, and I took the conservative side deliberately. Basing the decision on free GRES would be more precise, but the failure being fixed here is not "we picked a slightly slower partition" — it is four lanes that submitted, never started, and were cancelled after 3–5.5 h because a `mix` node with no free GPU was counted as available. The two errors are not symmetric: being too strict costs a step down a priority list that is tried in order, while being too loose costs the entire job. If GRES-aware selection is wanted it is worth doing on its own, with the queue-time data to show it helps.

PR #1809, opened 2026-09-03
Evaluate Mie-Gruneisen coefficients per cell along the 5-equation path

sbryngelson:
Stacked on #1805 (the backend) and overlapping #1808 (the ideal-gas cleanup); review the last three commits. Third step of #1638.

**What changes.** When any fluid's equation of state is state dependent, the mixture loop in `s_compute_mixture_coefficients` evaluates `s_eos_coefficients` per phase at that phase's own density; the stiffened-gas arithmetic is untouched, and the existing 27-case gate is bit-identical. `s_compute_speed_of_sound` takes the partial densities as an optional trailing argument and, under the same flag, mixes frozen per-phase moduli with the derivative term `-rho_i Pi_i'/Gamma_i` applied at that one site, so `f_bulk_modulus` and its twenty other callers do not change. The Riemann kernels, the cell-state helper and the post-process energy integral pass the densities through; everything that cannot is refused by the validator (5-equation model, HLL/HLLC/LF, no bubbles, hypoelasticity, IBM, IGR, relativity, MHD, chemistry, acoustic source, probes, post-process `c`, characteristic BCs or Wood's law). Those fences come out one by one in follow-ups; the 6-equation model needs the phasic isentrope generalized first and is its own PR.

**Parameter ownership.** A Mie-Gruneisen fluid has neither `gamma`, `pi_inf` nor `qv` and an ideal gas has no `pi_inf`; the validator refuses them, the suite's base fluid and every example drop the dead values, and all 744 test definitions validate. The reference-curve maths moves into `toolchain/mfc/eos.py`, one definition shared by the unit tests and the validation runners.

**Validation** (both registered as convergence-kind tests):
- `1D_mg_impact`: symmetric impact of two slabs; shock speed and plateau density match the Hugoniot `u_s = c0 + s u_p` to 1.4e-3 across four closing speeds (bound 5e-3).
- `1D_mg_acoustic`: a small simple-wave pulse travels at the analytic `c` to 1.9e-4 (bound 1e-3). This is the one observable that isolates the derivative term: without it the speed is 24% off.
- Golden `5AC2F65D`: a Mie-Gruneisen fluid beside an ideal gas, so one kernel carries both paths.

Also deletes the `gamma = gammas(1)` block in both output routines, which was assigned and never read.


sbryngelson:
Superseded by #1811, which carries this PR's content unchanged plus the rest of the family, the review round and the Cray OpenACC fix; consolidated so there is one PR to review and one CI to watch.

PR #1810, opened 2026-09-03
JWL, and state-dependent EOS through Wood's law, characteristic BCs, probes and post-process output

sbryngelson:
Stacked on #1805 and #1809 (and overlapping #1808); review the last two commits. Fourth step of #1638.

**JWL.** One more `case` in `s_eos_coefficients`: `p_ref = A e^{-R1 V} + B e^{-R2 V}`, `V = rho0/rho`, with `Gamma_G = omega`. The curve is itself an isentrope, so `de_ref/drho = p_ref/rho^2` and the derivative pair is two lines; the dispatch tail now works in `rho`-derivatives so the conversion is family-agnostic. Six parameters (`jwl_a, jwl_b, jwl_r1, jwl_r2, jwl_omega, jwl_rho0`), synthetic values only. The validator's selector is family-generic: each family owns its parameters, a state-dependent fluid has no `gamma`, `pi_inf` or `qv`, and `R1 > R2 > 0`.

**Fences lifted.** Wood's law is an N-fluid loop over per-phase moduli (bit-identical for two stiffened gases); the characteristic-BC coefficient derivative applies the per-phase chain rule `d(alpha Pi)/dt = Pi dalpha/dt + Pi'(dalpha_rho/dt - rho dalpha/dt)` (the alpha cancels, no division); probes and the post-process `c` output receive the partial densities. Still refused: bubbles, hypoelasticity, IBM, IGR, relativity, MHD, chemistry, acoustic sources, and the 6-equation model, whose relaxation step is written on the closed-form stiffened-gas isentrope and needs the phasic isentrope generalized first (its own PR).

**Verification.**
- 27-case stiffened-gas gate: bit-identical.
- Isentropic release (`1D_jwl_release`): every released cell sits on the closed-form JWL isentrope through the initial state to 7.5e-5 at N = 200, 9e-7 at N = 800 (bound 1e-3).
- Hugoniot and acoustic-speed validations still pass; the acoustic pulse now comes from a rectangle patch tracked by its centroid (2.5e-4 to 1.6e-6 over N = 100..400), because an analytic IC is compiled in and every distinct one costs the suite a full rebuild.
- Goldens: JWL beside an ideal gas, Mie-Gruneisen under Wood's law, Mie-Gruneisen with characteristic walls.
- 109 toolchain unit tests, all 748 test definitions validate.


sbryngelson:
Superseded by #1811, which carries this PR's content unchanged plus the rest of the family, the review round and the Cray OpenACC fix; consolidated so there is one PR to review and one CI to watch.

PR #1812, opened 2026-09-03
[CI probe, do not merge] CCE OpenACC EOS helper diagnostics

sbryngelson:
Temporary probe for the Frontier CCE OpenACC NaNs seen on #1811. Runs one lane on seven targeted tests and prints each EOS helper evaluated on the device next to its host value at startup. Will be closed once the failing construct is identified.


sbryngelson:
Closing: this probe has served its purpose. It answered two questions that the full suite could not, because both needed one Frontier lane running one test repeatedly.

**Cray OpenACC.** Three cycles narrowed the NaNs in the EOS helpers to an array element passed by reference into a device routine containing a loop. Recorded in #1815 and reproduced minimally in sbryngelson/compiler-bugs.

**Cray OpenMP.** Repointed at the reacting Roe HLLC shocktube, three passes of one binary gave three different answers:

    0.07593517165078
    0.07593517172084
    0.07593517170526

That is a data race, not compiler noise. The cause was `c_sum_Yi_Phi`, written inside the HLLC kernel and missing from the pure-fluid private list while present in the hypoelastic one. It predates #1811 and is fixed on master in #1823.

No code from this branch is intended for merge.

PR #1821, opened 2026-09-04
Add heterogeneous reacting surface boundary conditions

rocfire11:
## Contribution Policy

We do not accept pull requests generated primarily by AI without genuine understanding or real-world usage context.

All contributions are expected to demonstrate:
- A clear understanding of the codebase
- Alignment with product direction
- Thoughtful reasoning behind changes
- Evidence of real-world usage or hands-on experience with the problem

If these expectations are not met, we would prefer to implement the changes ourselves rather than spend time reviewing low-effort submissions.

---

## Acknowledgement

- [ X] I confirm this PR meets the above expectations and reflects my own understanding and real-world context.

---

_PR template credit: junegunn_

## Summary

This PR adds heterogeneous reacting surface boundary conditions for immersed
boundaries. The implementation was developed for reacting carbon-particle
simulations in which heterogeneous surface chemistry is coupled to the
compressible gas-phase species equations.

The surface treatment supports:

- zero-normal-gradient temperature (`thermal_bc = 0`)
- prescribed surface temperature (`thermal_bc = 1`)
- coupled reacting-surface energy balance (`thermal_bc = 2`)
- heterogeneous species-flux boundary conditions
- Cantera-based heterogeneous surface mechanisms through
  `surface_cantera_file` and `surface_phase`

For a reacting surface, the species boundary condition balances diffusive
transport, Stefan mass flux, and heterogeneous surface production. One species
equation is replaced by the mass-fraction closure. When `thermal_bc = 2`, the
surface temperature is included as an additional Newton unknown and the
conductive and heterogeneous reaction heat fluxes are balanced.

## Motivation

The immediate application is heterogeneous oxidation/gasification of carbon
particles using an immersed-boundary representation. The implementation is
intended to remain general with respect to the number of gas species and the
Cantera surface mechanism rather than hard-coding a particular carbon
mechanism.

## Testing

The implementation has been tested using an 11-species reduced GRI-based gas-phase 
mechanism together with a compatible heterogeneous carbon surface mechanism, including:

- prescribed surface temperature
- coupled surface species/energy solution
- MPI CPU/GPU simulations on HiPerGator

Both prescribed-temperature and coupled-energy carbon cases run successfully 
and produce the expected surface reaction products, oxygen consumption, 
thermal field, and reacting wake.

The branch was rebased from current MFC master before the surface changes were 
introduced, and ./mfc.sh precheck and the simulation build pass.


sbryngelson:
requires me to merge #1852 before this works on amd compilers

will make 'ready to review' when that merges. work on this pr can continue if needed (or ideally a follow up pr later)

sbryngelson:
Review of `94ac09d`. `format` and `precheck` are clean, so none of this is lint.

Worth saying first, because it shapes how to read the rest: the parts most likely to be *silently* wrong are right. Species indexing agrees end to end - `gas_index` is built from the main mechanism's `species_names`, which is the ordering `m_thermochem`, `Ys_IP` and `eqn_idx%species%beg` all use. The Stefan closure `mdot_s = sum(W_k omega_k)` correctly recovers the carbon mass crossing from the solid (reaction 5: 2*28.01 - 31.998 = 24.02 = 2 W_C). The `Y_k * sum_BG` diffusive-flux correction is the standard zero-sum form, and dropping species `num_species` for `sum(Y) - 1` is legitimate because `sum_k R_k` vanishes at the root. `norm = gp%levelset_norm` points into the fluid, so the blowing sign is right. The GPU private clause is complete. `molecular_weights` is copied into a kernel local before being passed down, which is exactly what the nvfortran 23.11/24.1 trap requires. And `s_add_cloud_particle` sets all three new `patch_ib` members - the single thing our pitfalls file says is most often missed.

The findings below are about the guards around that, not the physics.

## 1. `Ys_g = 2*Ys_s - Ys_IP` goes negative, and your own example reaches it

`src/simulation/m_ibm.fpp:316`. The mirrored extrapolation is unclamped, so whenever the surface consumes a species faster than diffusion resupplies it, `Ys_s < Ys_IP/2` and the ghost mass fraction is negative. It lands in a conserved variable at line 461.

Worked through with your mechanism, your `Twall = 1200 K`, Cantera mixture-averaged diffusivities, on the 25x25 grid the Example suite caps this case to (`d ~ 2.8e-4 m`):

| species | k (m/s) | Da | Ys_s/Ys_IP | Ys_g/Ys_IP |
| --- | --- | --- | --- | --- |
| OH | 57.2 | 48 | 0.020 | **-0.959** |
| O | 118.1 | 97 | 0.010 | **-0.980** |
| O2 | 0.032 | 0.041 | 0.961 | +0.921 |

R1 and R2 have `Ea = 0` and `k ~ sqrt(T)`, so the radical channels are diffusion-limited at *any* surface temperature - this is not a high-temperature corner. `sum(Ys_g)` is still exactly 1 because the products compensate, so no global check catches it. The alpha-QSS clip at `m_chemistry.fpp:271,288` hides the symptom, and only when `reaction_substeps > 0`.

**Fix:** clamp the mirror, or drop to first order where the mirror would leave the physical range. The second keeps the surface value exact:

```fortran
if (2._wp*Ys_s(q) - Ys_IP(q) < 0._wp) then
    Ys_g(q) = Ys_s(q)          ! first-order ghost; the mirror would be unphysical
else
    Ys_g(q) = 2._wp*Ys_s(q) - Ys_IP(q)
end if
```

Renormalising afterwards is not enough on its own - the sum is already 1.

## 2. `T_g = 2*T_s - T_IP` has the same problem, and it bites `thermal_bc = 1` too

`src/simulation/m_ibm.fpp:296` (inert) and `:317` (reacting). `T_g` feeds `alpha_rho_IP(1) = alpha_IP(1)*pres_IP*mw_g/(gas_constant*T_g)` and `get_mixture_energy_mass(T_g, ...)`. A 1200 K wall under a 2500 K flame - the carbon-combustion case - gives `T_g = -100 K`: negative ghost density, and NASA polynomials evaluated at negative `T` (the `a6/T` term). The Newton solve clamps `T_s` internally to `[200, 5000]`; nothing clamps the extrapolation. Same one-line treatment as above, with a floor rather than zero.

## 3. `thermal_bc` and `Twall` are silently ignored without chemistry

The whole block at `m_ibm.fpp:288` sits inside `if (chemistry .and. patch_ib(patch_id)%inj_species == 0)`, but `m_checker.fpp:104` was loosened from `if (ib .and. chemistry)` to `if (ib)`, and `docs/documentation/case.md:366,409` documents them as general IB thermal boundary conditions. So an isothermal cylinder in a plain Navier-Stokes IB run validates, documents as working, and is silently adiabatic. `thermal_bc` is read nowhere else in `src/` - I grepped. Either gate the parameters on `chemistry` or lift the thermal branch out of the conditional; the docs should match whichever you pick.

## 4. The generator mis-emits sticking and reversible surface reactions

`toolchain/mfc/run/input.py:236-247`. `append_reaction_rate` guards only on `hasattr(rate, "pre_exponential_factor")`. Checked against the Cantera 3.1.0 in the project venv: `ct.StickingArrheniusRate(0.1, 0, 0).pre_exponential_factor` returns `0.1`, so a sticking reaction passes the guard and its dimensionless sticking probability is emitted as an Arrhenius `A` - wrong by 1e4 to 1e12. Separately `reaction.reversible` is never consulted, so a mechanism written with `<=>` (Cantera's default for interface reactions) silently loses its reverse branch, and `rate.coverage_dependencies` is dropped.

Your own mechanism uses irreversible `=>` with explicit orders and no sticking, so none of this affects your results - it is a trap for the next user. The generator already raises cleanly for surface-site species and unsupported thermo; three more `raise`s in the same style would close it.

## 5. Two documented device-routine portability traps

`s_surface_species_residual` (`m_ibm.fpp:1705`) and `s_surface_energy_residual` (`:1740`) are `GPU_ROUTINE(parallelism='[seq]')` and call `get_species_mass_diffusivities_mixavg` / `get_mixture_thermal_conductivity_mixavg` / transitively `get_species_enthalpies_rt` from inside. `.claude/rules/common-pitfalls.md` records this exactly: calling `get_species_*` from inside a `GPU_ROUTINE` rather than from the kernel gave CCE OpenMP a runtime `Memory access fault by GPU node-N` on the first step, while every other backend ran. The build stays clean and only a case that reaches the path shows it, so NVIDIA and CPU lanes passing is not evidence. Every existing call site (`m_chemistry.fpp:409-416`) evaluates these in the `GPU_PARALLEL_LOOP` body and passes the arrays down - same restructuring here.

(I checked and discarded the related `ftn-7066` concern: `num_species` is an `integer, parameter` in the pyrometheus output, so those bounds are constants, not device globals.)

## 6. Newton non-convergence is a silent BC switch, and it will make the golden flaky

`m_ibm.fpp:1875,1898` return `converged = .false.`, consumed at `:321` and `:454`, and the ghost point silently reverts to an inert zero-flux surface - no counter, no warning, no output field. Beyond the diagnostic problem, `converged` is a discontinuous branch driven by a strict `norm_trial < norm_R` with no Armijo slack, so a one-ulp difference between compilers flips a ghost point between two O(1)-different boundary conditions. `tests/F52F0D4C` is an Example-tolerance golden on top of GRI-11 kinetics; we have already had to skip `1D_propellant_flame`, `2D_hybrid_slab` and `1D_flamelet` for subtler drift than this. Expect it to be flaky across lanes.

Related: the Example test is not step-capped. `cases.py:3177` caps `t_step_stop` only `if "t_step_stop" in case`, and this example uses `cfl_adap_dt` with `t_stop`/`t_save`, so the golden is the full run - ~1e3 steps of GRI-11 with a 30-iteration Newton (12 residual evaluations each, each O(Ns^2)) per ghost point per RK stage. Several adaptive-dt examples are already in `casesToSkip` for this. A small step-capped dedicated test would serve better than the full-run Example golden.

## 7. Smaller

- No Python-side validation: `case_validator.py` is untouched, so `./mfc.sh validate` accepts a bad `thermal_bc` and a real run does a full pre_process before aborting. The IB block at lines 906-944 already validates `airfoil_id`/`model_id`/`burn_rate_exp`; these belong there with `PHYSICS_DOCS` entries.
- Build slug: `build.py:310` correctly keys the gas mechanism on `.source` - the identical argument applies to `surface_cantera_file`/`surface_phase`, which are not hashed at all, so two cases differing only in surface mechanism share a binary.
- `input.py:118-122` has a broad `except Exception`: if the local surface file exists but fails to parse, the error prints dim and a same-named file in `MFC_MECHANISMS_DIR` loads instead - silently a different mechanism.
- `d = abs(gp%levelset)` has no floor and `1/d` is in both residuals. The NaN is contained (every line-search comparison is false for NaN, so it exhausts and returns `converged = .false.`) but that lands in finding 6. `max(d, small)` is clearer than relying on NaN comparison semantics.
- `dx = fd_eps_Y*max(abs(Ys_s(j)), 1._wp)` (`:1841`) - mass fractions are <= 1 so the `max` is always 1 and the scaling is dead. `min` was probably intended.
- The pivot test `pivot_value <= epsilon(1._wp)` (`:1935`) is absolute on a matrix with O(1e7) entries; it will never fire. A relative test against the column norm would.
- Convention: the scalars use second-order mirroring while the Stefan velocity is added once to `vel_g` (`:426`), matching the first-order `v_blow` convention. If reconstruction sees the wall-normal velocity as ~`v_stefan/2`, the convected mass flux is half what the species BC assumed. May well be deliberate, but the two halves of one surface condition using different ghost conventions deserves a comment.
- `get_mixture_molecular_weight(Ys_IP, ...)` and `Xs_IP` are recomputed inside every residual evaluation though `Ys_IP` never changes; the transport coefficients could be frozen at the outer Newton level too.

## Suggested order

1 and 2 are the ones that produce wrong answers rather than crashes, and both are a few lines. 3 is a one-line gate. 4 is three `raise`s matching the ones already there. 5 is the restructuring `m_cbc`/`m_ibm` already use. 6 and 7 argue together for a cheap step-capped test instead of the full-run Example golden.

Separately: your Frontier AMD CPU build failures are not yours to fix - they are the 11-species mechanism against the `dimension(10)` literal under the `USING_AMD` guard, which #1852 raises. I cherry-picked #1852 onto `94ac09d` locally and the tree builds clean on MI210 with `F52F0D4C` passing, so that lane clears when #1852 lands. The `gpu-omp [2/2]` lane was a bad node (`syscheck` failed three times on `frontier10212`).

rocfire11:
As a newbie, is there any required of me right now before merging with master branch? Your steps 3-7 will be addressed in a separate PR since I do not want to corrupt this one.

sbryngelson:
you can leave this here as is, thanks @rocfire11

sbryngelson:
## Ghost-state reconstruction at the reacting surface

I reviewed this and pushed the fixes to your branch as ad6893a (one commit on top of b857850). Most are small — a hardcoded `dimension(10)` for `Ys_s`, the surface mechanism missing from the build hash, sticking-coefficient rates passing a `hasattr` guard and being emitted as plain Arrhenius, a few checker gaps. The one that changes results is the ghost-state reconstruction, so here is what changed and why.

### The problem

The ghost value at a reacting surface is set by the linear mirror

```
phi_g = 2*phi_s - phi_IP
```

which makes the midpoint of the ghost/image-point pair reproduce the surface value. Its `-1` coefficient on `phi_IP` is unconditional, so it goes negative whenever the surface value drops below half the image-point value. Two places that bites:

- **Species.** A carbon surface consumes O and OH essentially to zero. With `Y_s ≈ 1e-9` and `Y_IP ≈ 1e-3`, the mirror gives `Y_g ≈ -1e-3`. I instrumented the PR head and ran the `2D_ibm_reacting_surface` Example test as-is: **114,377 ghost-point updates write a negative mass fraction**, 90% of them O, most of the rest OH. The existing golden for that test encodes those values.
- **Temperature.** With `thermal_bc = 1`, `T_g = 2*Twall - T_IP`. A 300 K wall in 1500 K gas gives `T_g = -900 K`, hence negative ghost density via `rho = p*W/(R*T)`. The Example never hits this because its wall is hotter than the freestream, but a wall of 210 K against 298 K gas gives `T_g = 122 K`, which is below the 200 K floor of every NASA polynomial in the mechanism. The run doesn't crash — it silently produces a wall layer that is 64% too dense (max density 2.87 vs 1.75 with the fix; the ratio matches `T_g` 201/122 to 0.2%, so the excess is exactly the too-cold ghost temperature).

### What I changed

Rather than clamp the result (which breaks `sum(Y) = 1` and hides the excursion), the ghost state is now a convex blend of the mirror back toward the surface value:

```
phi_g(theta) = phi_s + theta*(phi_s - phi_IP),    theta in [0, 1]
```

`theta = 1` is the existing second-order mirror; `theta = 0` is the first-order Dirichlet ghost `phi_g = phi_s`. Each ghost point takes the largest `theta` that keeps the state physical:

- **Composition** uses one `theta_Y` shared by every species, the largest value keeping every `Y_g(k) >= 0`. Because both endpoints satisfy `sum(Y) = 1`, every blend between them does too — exactly, with no renormalization.
- **Temperature** uses its own `theta_T`, the largest value keeping `T_g` inside `[T_surface_min, T_surface_max] = [200, 5000] K`, i.e. the range the thermodynamic fits are valid over (now a parameter in `m_constants`, shared with the Newton solve and the checker). Bounding on positivity alone is too weak: a 300 K wall in 1500 K gas would still admit `T_g = 30 K`.

The two factors are decoupled deliberately. My first version shared one `theta`, and on your Example a trace radical at `1e-9` dragged the thermal mirror from 2090 K down to 1205 K. That was wrong: nothing couples the temperature reconstruction to the composition.

Wherever the mirror is already physical, `theta = 1` and nothing changes.

### What it costs

On the Example, `theta_Y` is ≈ 0.006 or 0 at essentially every ghost point on the burning surface — the composition ghost is first-order there, always. That is the honest answer: the surface *is* a sink for those radicals, and their gradient is the steepest thing in the problem. Two things make the downgrade cheap:

- The linear mirror was only ever first-order in the wall *flux* (Ezra et al., IJHMT 2025 — it is second-order in the field, first-order in the Nusselt number).
- A boundary closure one order below the interior scheme retains the interior convergence rate (Gustafsson 1975; Svärd & Nordström 2014).

This is the same construction as the Zhang–Shu / Hu–Adams–Shu positivity limiter — blend toward a known-admissible state rather than clip — applied to a wall ghost instead of a face flux. `phi_g = phi_s` is Gibou et al. (2002) Eq. 17, and is what they also fall back to where the linear form is ill-behaved.

What it does *not* do: make the boundary strictly conservative. Başkaya et al. (Comput. Fluids 2024) trace this same cold-wall, steep-gradient regime in ablation to ghost-cell mass-conservation error that shows up as spurious blowing, and only a flux-based cut-cell boundary fixed it for them. That's a different discretization, not a tuning of this one, and out of scope here — but worth knowing about if you push into colder walls.

### Effect on tests

- The `2D -> Example -> ibm_reacting_surface` golden (`F52F0D4C`) is regenerated. It changes because the old values encoded negative O/OH ghosts; the first failing variable is a species (`cons.10`), and density (`cons.1`) is unchanged, as expected since `theta_T = 1` on this case.
- Added `2D -> IBM -> Reacting Surface -> Cold Wall` (`Twall = 210 K`) to pin the temperature-limited branch, which the Example never reaches.
- 64/64 2D IBM + chemistry cases pass; the 3D IBM cases don't reach this code without `chemistry = T`.

### Also in the branch, briefly

- `sum(Y) = 1` now displaces the *most abundant* species' balance, re-picked each Newton iteration (what Cantera's `solveSP` does), instead of the last species in the mechanism — which for the shipped GRI-11 subset is H2O2, a trace radical. Surface CHEMKIN's manual warns the displaced species absorbs every other balance's roundoff.
- Ghost points with zero levelset distance are skipped and counted rather than dividing by `d`; NaN residuals were defeating the pivot test, so the solve reported success.
- Non-converged and ill-posed surface solves are counted and printed at the end of the run (same pattern as `s_report_pressure_relaxation`). Previously a failed solve silently fell back to an inert wall.
- Newton tolerance loosened `1e-8 -> 1e-6`. The forward-difference Jacobian makes convergence linear near the root; Cantera runs the same `1e-7` step against `1e-4`.

Feel free to reword, split, or push back on any of it. Review and fixes were done with Claude Code.

rocfire11:
Thanks, Spencer. I went through the changes carefully and they look good to me. In particular, I agree with using separate species and temperature blending factors; this avoids trace species unnecessarily limiting the temperature reconstruction. The dynamic choice of the species equation replaced by the mass-fraction closure also makes sense, as do the added diagnostics for failed/ill-posed surface solves and the additional mechanism checks.

I also agree with keeping the present ghost-state treatment within the scope of this PR. I will explore other limiters

The additional cold-wall test is helpful, and I am satisfied with the current changes. Thanks for the detailed review and fixes.

sbryngelson (dismissed):

sbryngelson:
@rocfire11 pr approved so benchmarking will run and it will check self-hosted frontier/amd runners - and nvidia gpu. shold check out by my measurements. if anything looks awry in these changes (now or in future) do flag it immediately. AI is not a full surrogate for the domain experts @rocfire11

sbryngelson:
Thanks for going through it. One correction to flag, since you approved partly on the strength of the cold-wall test and it was broken when you read it.

That test ran to `t = 2e-5`, and its golden did not survive a change of compiler. Generated under nvhpc 25.11, it missed GNU and **every other nvhpc release** by ~1e0 relative in energy — 22 red CI lanes, none of them an actual regression. My fault, not something in your code: I verified it on one compiler locally and assumed that generalized.

A cold wall is exactly the state you cannot run long and compare tightly. It pins the ghost temperature against the 200 K floor of the NASA fits, and that state then feeds back through the stiff surface and gas kinetics, so small differences in how each compiler evaluates them compound.

Fixed in 45c77c0 by stopping the case at one step. It still pins the branch it exists for — 193 ghost updates take the limited path, at `theta_T = 0.102` — but the answer is now set by the reconstruction rather than by accumulated kinetics. Golden regenerated; both reacting-surface goldens pass locally.

Worth knowing for the limiters you mentioned exploring: any test you add for a cold or strongly-reacting wall will have this property. Keep it short, or it will pin your compiler rather than your physics.

Nothing needed from you — just did not want you approving a state that has since changed.

sbryngelson:
Opened three follow-up issues from reading through this, none of which need anything from you on this PR:

- #1891 — move the surface Fortran codegen upstream into pyrometheus. The gas phase is generated in one line by pyrometheus; the surface phase is ~400 hand-written lines in MFC's toolchain doing the same job, including a re-implementation of NASA polynomial evaluation for non-gas species. Both bugs I found in that file (sticking coefficients emitted as Arrhenius, dropped coverage dependencies) are the failure mode of maintaining a second, partial chemistry generator. Pyrometheus has no interface/surface support today, so this is a real upstream feature.
- #1892 — extract the surface Newton solver out of `m_ibm.fpp`, which is now 2220 lines. The solver takes `(pres, T, d, Ys)` and returns a surface state; nothing in it knows what a ghost point is. Costs nothing on GPU — `m_ibm` already makes cross-module device calls from inside the ghost loop — and makes the solve unit-testable, which is where the missing coverage is.
- #1893 — tracker with the ordering.

To be clear about scope: this is not an argument that this PR should have been split. The remaining ~1400 lines are the example, its mechanisms, goldens and parameter plumbing, which are the right size and in the right place. The two extractions are worth doing because they are what make the feature testable without running a full 2-D reacting simulation — and #1891 in particular deletes MFC code rather than adding any.

sbryngelson:
Ran this on a GPU, which nothing in the test suite does — every reacting-surface golden is CPU-only, so the offload path had never been executed.

**It works, and the answer is bit-identical to CPU.**

| check | result |
|---|---|
| `--gpu acc` build (nvfortran 25.11) | clean |
| device routines emitted | `!$acc routine seq` under `#if MFC_OpenACC` |
| run on an A100 | exit 0, 47 steps, no NaN |
| surface Newton solve | no failures reported — every ghost point converged |
| CPU vs GPU fields | **bit-for-bit identical**, 10,752 cells, all 32 field files at the final step |

Case: the `2D_ibm_reacting_surface` example at `cells_per_D = 16`, `Tstar = 0.05`, `parallel_io = F`, single rank, one GPU.

Worth checking because `s_solve_surface`, `s_blend_ghost_state` and `s_pick_bath_species` are all `GPU_ROUTINE(parallelism='[seq]')` called from inside the ghost-point parallel loop, and that pattern has two documented traps in this codebase: CCE rejects such a call when it is nested inside another `routine seq`, and nvfortran 23.11/24.1 segfault when a `parameter` array from `m_thermochem` is passed into a declare-target routine. Neither bites here — `W_species` is copied from `molecular_weights` into a local before the call, which is the form that works.

**Scope.** This is one backend of three. It says nothing about CCE OpenACC, Cray or AMD flang OpenMP offload, or `--gpu mp` on nvfortran — and CCE is the one with the sharpest edges. But it is one lane verified where there were none, and the bit-identical result means the surface solve is not accumulating any device/host divergence.

Happy to run `--gpu mp` on nvfortran too if that is useful; the AMD and Cray lanes need CI.

sbryngelson (dismissed):

sbryngelson (approved):

PR #1822, opened 2026-09-04
ci: run the NVHPC container lanes without MPI

sbryngelson:
Every failure the NVHPC container lanes have produced has been an **MPI test**.

`post_process` segfaults under `mpirun` on the GitHub-hosted runners — exit 139, then `h5dump` cannot open `p0/0.silo` because the file was never written. Observed on:

| test | UUID | seen on |
|---|---|---|
| `3D -> 2 MPI Ranks` | `CE232828` | #1816, #1800 |
| `MPI Consistency -> 3D -> Viscous` | `0090B316` | #1800, TerminalOutput |

It is intermittent, spans NVHPC 24.9 / 24.11 / 25.11, and the build always succeeds — the crash is at run time. The same tests pass on the GNU and Intel lanes and on the self-hosted clusters that run MPI at scale, so this is specific to running MPI in these containers on GitHub's runners.

## The change

Add `--no-mpi` to the NVHPC **cpu** build and test steps. The harness already skips `ppn > 1` cases when built without MPI (`toolchain/mfc/test/test.py`), and both failing tests are `ppn=2`, so this drops exactly those and keeps the rest of `--test-all`.

That preserves what these lanes exist for: compile coverage across 15 NVHPC releases, where breaks have reached CI before (the `GPU_DECLARE` ordering break and two ICEs). MPI runtime coverage is unaffected — the GNU, Intel and self-hosted lanes all still run it.

Only the `cpu` lanes change; the `gpu` lanes are build-only and never ran tests.

## What this does not do

It does not diagnose the segfault. A bisect against #1762 was inconclusive — it did not reproduce locally under NVHPC 24.1, which is not one of the affected versions, so that result says nothing either way. If the crash is a real `post_process` bug rather than runner flakiness, this hides it on these lanes; it would still be caught by the GNU, Intel and self-hosted MPI lanes, which exercise the same code paths.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -289,12 +289,21 @@ jobs:
           CHANGED_FILES: ${{ needs.file-changes.outputs.changed_files }}
 
       # ── NVHPC build + test (via docker exec into long-lived container) ──
+      # --no-mpi: every failure these lanes have produced has been an MPI test.
+      # post_process segfaults under mpirun on the GitHub-hosted runners --
+      # "3D -> 2 MPI Ranks" and "MPI Consistency -> 3D -> Viscous" -- while the
+      # same tests pass everywhere else, including the self-hosted clusters that
+      # actually run MPI at scale. The harness skips ppn>1 cases without MPI
+      # (test.py), so this drops exactly those and keeps the rest of --test-all,
+      # which is what these lanes exist for: compile coverage across 15 NVHPC
+      # releases. MPI runtime coverage stays on the GNU, Intel and self-hosted
+      # lanes.
       - name: Build (NVHPC)
         if:   matrix.nvhpc && matrix.target == 'cpu'
         run: |
           docker exec nvhpc bash -c '
             source /etc/nvhpc-env.sh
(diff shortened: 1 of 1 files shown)


sbryngelson:
Both taken, in `28f72b26`.

**Duplicated flags** — real hazard, not just style: if the build and test steps drifted, a no-MPI build would be tested with MPI and `ppn>1` cases would run against a binary that has none. They now come from a single `MFC_NVHPC_TEST_FLAGS`.

One detail worth recording, since the obvious implementation would not have worked: a job-level `env:` does not reach inside the container. These commands run under `docker exec nvhpc bash -c '...'` with single quotes, so `$(nproc)` and anything else expands *inside* the container, not on the runner. The variable is therefore passed at `docker run` time alongside the existing `-e CC`/`-e FFLAGS`, where the long-lived container picks it up for every subsequent `docker exec`.

YAML anchors were the other suggestion, but GitHub Actions does not support them in workflow files.

**Comment length** — fair, trimmed from nine lines to the essentials with the incident detail left in this PR. I kept the one non-obvious consequence inline (why the two steps must agree), since that is the thing a future editor could break without noticing.

Issue #57, opened 2023-01-14
Some of the example cases do not run to completition


sbryngelson:
Some example cases have CFL errors before completing. For example, `2d-shock-bubble` and `3d-sph-bubble-collapse`.

sbryngelson:
@anandrdbz

sbryngelson:
closed in lieu of https://github.com/MFlowCode/MFC/issues/98

PR #1825, opened 2026-09-05
ci: resolve CodeQL code-scanning alerts in workflows

sbryngelson:
Clears all 27 open [code-scanning alerts](https://github.com/MFlowCode/MFC/security/code-scanning). Both classes live in GitHub Actions workflows — no Fortran or toolchain source is touched.

## `missing-workflow-permissions` — 26 alerts, medium

No workflow declared a `permissions:` block, so every job inherited the repository-default `GITHUB_TOKEN` scope. That's read-only under current settings, but it's settings-dependent and would silently become write if the org default ever changed.

I checked what each flagged job actually does with the token: **none of the 26 write to the repo via `GITHUB_TOKEN`**. The three jobs that do publish use their own secrets instead — `docs.yml` clones `secrets.DOC_PUSH_URL`, `homebrew-release.yml` uses `secrets.TAP_REPO_TOKEN`, `coverage.yml` uses `secrets.CODECOV_TOKEN`. So a workflow-level read default is sufficient everywhere:

```yaml
# Least-privilege default: no job in this workflow writes to the repo.
permissions:
  contents: read
```

Applied to `bench`, `cleanliness`, `convergence`, `coverage-health`, `coverage`, `docs`, `formatting`, `fp-stability`, `homebrew-release`, `homebrew`, `lint-toolchain`, `pmd`, `spelling`, `test-toolchain-compat`, `test`. All 21 workflows now declare permissions at workflow or job level, and every file still parses as YAML.

## `untrusted-checkout/high` — 1 alert, high (#57)

`claude-code-review.yml` runs privileged — `pull_request_target` + `issue_comment`, so it holds `CLAUDE_CODE_OAUTH_TOKEN` along with `pull-requests: write` and `issues: write` — and it fetched fork commits into that privileged workspace via `git fetch origin pull/<n>/head`.

That step turned out to be **dead code**: nothing read `FETCH_HEAD`, and the `pr_head_ref` output was written to `$GITHUB_OUTPUT` and never consumed. The review already sources its diff and per-file context through `gh` API calls that treat fork content as data rather than as checked-out code. So the step and the unused output are removed outright rather than sandboxed — the flagged pattern goes away with no behavior change.

A comment on the remaining `actions/checkout` records why the base repo is the only thing checked out, so the fetch doesn't get reintroduced later.

## Follow-ups, not addressed here

Two things worth a separate look — neither is flagged by CodeQL and neither is fixed in this PR:

- **Prompt-injection surface remains in `claude-code-review.yml`.** Claude reads an attacker-authored diff while running `--dangerously-skip-permissions` with `Bash` allowed, in a job holding write scopes. `claude-code-action` gating on the triggering actor's write access is the main protection. Since the review comment is already posted by a *separate* step, the job could be split — review at `contents: read`, publish at `pull-requests: write` — so the model's step holds no write token at all.
- ~~**`bench.yml`'s `pull_request_review` path may be a functional bug.**~~ **Retracted — I checked and this was wrong.** On `pull_request_review`, `actions/checkout` resolves to the PR merge ref (`refs/remotes/pull/<n>/merge`), not the base branch, so `Clone - PR` clones the PR as intended. Verified against run logs, and confirmed end-to-end: PR #1772's approval at `18:32:41Z` triggered a bench run 4 seconds later in which all six Phoenix/Frontier jobs ran for 34–175 minutes. No bug here.

https://claude.ai/code/session_01XPqfEaUBG7ZaZVzeMWnKHd

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -11,8 +11,11 @@ concurrency:
   cancel-in-progress: true
 
 # Least-privilege default: no job in this workflow writes to the repo.
+# pull-requests: read is required by dorny/paths-filter, which reads the PR's
+# changed-file list through the API on pull_request events.
 permissions:
   contents: read
+  pull-requests: read
 
 jobs:
   file-changes:
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -11,11 +11,8 @@ concurrency:
   cancel-in-progress: true
 
 # Least-privilege default: no job in this workflow writes to the repo.
-# pull-requests: read is required by dorny/paths-filter, which reads the PR's
-# changed-file list through the API on pull_request events.
 permissions:
   contents: read
-  pull-requests: read
 
 jobs:
   file-changes:
@@ -23,6 +20,13 @@ jobs:
     if: >
       github.event_name != 'pull_request_review' ||
       github.event.review.user.type != 'Bot'
+    # Job-level permissions replace the workflow default outright rather than
+    # merging with it, so contents must be restated here. paths-filter reads the
+    # PR's changed-file list via pulls.listFiles; this is the only job that needs
+    # it, so it is granted here instead of workflow-wide.
+    permissions:
+      contents: read
+      pull-requests: read
     runs-on: 'ubuntu-latest'
     outputs:
       checkall: ${{ steps.changes.outputs.checkall }}
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -10,6 +10,10 @@ concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}${{ github.event_name == 'pull_request_review' && format('-review-{0}', github.run_id) || '' }}
   cancel-in-progress: true
 
+# Least-privilege default: no job in this workflow writes to the repo.
+permissions:
+  contents: read
+
 jobs:
   file-changes:
     name: Detect File Changes
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -58,25 +58,16 @@ jobs:
             exit 1
           fi
 
-          PR_HEAD_REF="$(gh pr view "$PR_NUMBER" --repo "${{ github.repository }}" --json headRefName --jq .headRefName)"
-
           echo "pr_number=$PR_NUMBER" >> "$GITHUB_OUTPUT"
-          echo "pr_head_ref=$PR_HEAD_REF" >> "$GITHUB_OUTPUT"
 
+      # Base repo only. This job is privileged (pull_request_target / issue_comment:
+      # it holds secrets and a write-capable token), so the PR head is never checked
+      # out or fetched here. The diff and per-file context are pulled through the gh
+      # API in later steps and treated purely as data.
       - name: Checkout base repo
         uses: actions/checkout@v5
         with:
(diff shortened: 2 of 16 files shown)


sbryngelson:
Addressed in 5f4be21 — Copilot's finding was correct, and its list of six files was exactly right.

Declaring a `permissions:` block sets every unlisted scope to `none`, so `contents: read` alone dropped `pull-requests` for the six workflows whose `file-changes` job runs `dorny/paths-filter@v4`. That action defaults to `token: ${{ github.token }}` and reads the changed-file list via `pulls.listFiles` on `pull_request` events. `test.yml` additionally passes `list-files: shell`, which consumes that list directly. None of the six override `token`, so all of them take the API path.

Those six now declare:

```yaml
permissions:
  contents: read
  pull-requests: read
```

I re-audited the other nine while I was in there, and they stay at `contents: read`:

- None of the 15 modified workflows reference `GITHUB_TOKEN` or `GH_TOKEN` at all.
- No SARIF upload, no release action, no `gh` write calls.
- The two workflows that push (`docs.yml` → `DOC_PUSH_URL`, `homebrew-release.yml` → `TAP_REPO_TOKEN`) authenticate with their own secrets, so `GITHUB_TOKEN` scope doesn't affect them. Same for `coverage.yml` via `CODECOV_TOKEN`.
- `upload-artifact@v6` / `download-artifact@v7` use `ACTIONS_RUNTIME_TOKEN`, not `GITHUB_TOKEN`, and `docs.yml` downloads from its own run (`needs: build`), so no `actions: read` is required.

Verified 1:1: the set of workflows using `paths-filter` and the set granted `pull-requests: read` are identical.

One note that reinforces the `bench.yml` follow-up in the description rather than anything in this diff: `bench.yml`'s `file-changes` job also fires on `pull_request_review`, and `paths-filter` documents support for `push` / `pull_request` / `pull_request_target`, expecting an explicit `base`/`ref` otherwise. That's pre-existing and untouched here, but it's a second reason the review-triggered benchmark path deserves its own look.

sbryngelson:
Tightened the scoping in 84cf49a (and mirrored on #1826 so the two files stay identical).

Previously `pull-requests: read` sat at the workflow level in all six `paths-filter` workflows, which handed it to every job in those files — the multi-hour NVHPC container lanes, the self-hosted Frontier and Phoenix runs, the benchmark jobs. None of them touch the PR API.

Only `file-changes` needs it: it's the sole job running `dorny/paths-filter`, it finishes in about ten seconds, and it holds no other privilege. So the workflow default returns to `contents: read` and the grant moves into the job:

```yaml
jobs:
  file-changes:
    name: Detect File Changes
    permissions:
      contents: read
      pull-requests: read
```

The subtlety worth recording, and the reason `contents: read` is restated inside the job block: **job-level permissions replace the workflow default outright rather than merging with it.** A bare `pull-requests: read` there would have silently revoked `contents` and broken the job's own `actions/checkout`. That's the same failure mode as the original finding on this PR — declaring a permissions block sets every unlisted scope to `none` — just one level down.

Effective result across the six workflows, verified by parsing each file and resolving job-level against workflow-level:

| Job | Effective permissions |
|---|---|
| `file-changes` | `contents: read`, `pull-requests: read` |
| every other job | `contents: read` |

Applied to `bench`, `cleanliness`, `convergence`, `coverage`, `fp-stability`, `test`. The other nine workflows in this PR have no `paths-filter` job and are unchanged at `contents: read`.

PR #1826, opened 2026-09-05
ci: cap NVHPC lane concurrency with max-parallel

sbryngelson:
The 30 NVHPC lanes dispatch all at once and occupy 30 standard-runner slots for roughly two hours, leaving every other workflow queued behind them. Minutes are free on this repo; **concurrency is the scarce resource**.

## The change

`strategy.max-parallel` is the direct control — a matrix entry queued behind it holds no runner. The catch is that it applies to an *entire job matrix*, and the NVHPC entries currently share the `github` job with the ubuntu/macOS lanes, so capping in place would have throttled those too.

So the NVHPC entries move into their own `nvhpc` job that carries the cap:

```yaml
  nvhpc:
    name: NVHPC ${{ matrix.nvhpc }} (${{ matrix.target }})
    strategy:
      fail-fast: false
      max-parallel: 6
      matrix:
        nvhpc: ['23.11', '24.1', ... '26.3']
        target: [cpu, gpu]
```

Peak slot usage for these lanes goes **30 → 6**. The 30 hand-written `include:` entries also collapse to a two-axis product.

## What is deliberately unchanged

No compiler coverage is dropped. All 15 releases from 23.11 through 26.3 still build; `cpu` still runs `--test-all`; `gpu` still builds acc then omp plus the case-optimized ICE guard. This trades wall-clock for slots and nothing else.

- **Job names are byte-identical** (`NVHPC 25.5 (cpu)`), so no check-name churn and no branch-protection impact.
- **No job `needs:` the `github` job**, so the split adds no ordering constraints.
- I verified all 15 step bodies are byte-identical to `master` apart from the intended edits: dropping the now-vacuous `matrix.nvhpc` / `!matrix.nvhpc` guards, and moving each `# ── NVHPC ... ──` section comment along with the steps it describes.

Side benefit: this staggers the nvcr.io image pulls that the retry loop in `Pull NVHPC container` exists to survive — that step's own comment notes "~30 matrix jobs hit it at once".

## Tuning

`max-parallel: 6` is a starting point, not a derived optimum. Raise it for faster feedback, lower it to free more slots. Timings from a recent master run, if useful for tuning:

| Lane | Fixed overhead | Build | Test | Total |
|---|---|---|---|---|
| cpu | ~7 min | 11.7 min | 102.4 min | ~120 min |
| gpu | ~5 min | 90 min (acc+omp) + 13.8 (case-opt) | — | ~110 min |

## Not included

Two larger reductions I looked at but left out, since this PR is deliberately scoped to the concurrency cap:

- ~~The cpu lane spends 102 of its 120 minutes running `--test-all`.~~ **Now included — see "cpu test step" below.**
- The `cuda_multi` image tag is what forces the `Free disk space` step; a single-CUDA tag would be far smaller. That depends on MFC's actual CUDA requirements, which I haven't verified.

## Also included: `--test-all` off the cpu test step

`--test-all` does two unrelated things (`toolchain/mfc/test/test.py`):

| Where | What it does | Keep? |
|---|---|---|
| Build (line 389) | adds `post_process` to the compiled binaries | **yes** — this is the compile coverage |
| Test (line 735) | re-runs `pre_process+simulation+post_process` a second time per case, plus silo processing | **no** — ~half the lane's wall clock |

Only the first is what the matrix exists for. The second duplicates runtime checking the GNU, Intel and self-hosted lanes already do.

So the flag moves out of the shared `MFC_NVHPC_TEST_FLAGS` and onto the build step alone:

```yaml
-e "MFC_NVHPC_TEST_FLAGS=--no-mpi"        # was: --test-all --no-mpi
# Build (NVHPC):  mfc.sh test --dry-run ... --test-all $MFC_NVHPC_TEST_FLAGS
# Test  (NVHPC):  mfc.sh test --max-attempts 3 ... $MFC_NVHPC_TEST_FLAGS
```

`post_process` still compiles on all 15 releases — no compiler coverage is lost — it just is not re-run here. `--no-mpi` stays shared, because that one genuinely must match between build and test, which is what the original comment was guarding.

The gpu lanes are untouched: they already pass `--test-all` explicitly and never referenced the env var.

Expected effect on the cpu lane is roughly halving the ~102 min test step rather than eliminating it, since the flag adds one extra pipeline run per case. I have not measured it — the first run on this branch will show the real number.

https://claude.ai/code/session_01XPqfEaUBG7ZaZVzeMWnKHd

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -257,7 +257,7 @@ jobs:
             -e "FFLAGS=-tp=px -Kieee -noswitcherror" \
             -e CFLAGS=-tp=px \
             -e CXXFLAGS=-tp=px \
-            -e "MFC_NVHPC_TEST_FLAGS=--test-all --no-mpi" \
+            -e "MFC_NVHPC_TEST_FLAGS=--no-mpi" \
             "$NVHPC_IMAGE" sleep infinity
 
       - name: Setup NVHPC
@@ -296,16 +296,25 @@ jobs:
       # MFC_NVHPC_TEST_FLAGS carries --no-mpi: post_process segfaults under
       # mpirun in these containers, and every failure the lanes have produced
       # has been an MPI test (see PR #1822). The harness skips ppn>1 cases
-      # without MPI, so the rest of --test-all -- the compile coverage these
-      # lanes exist for -- is unaffected, as is MPI coverage elsewhere. Set in
-      # one place so the build and test steps cannot disagree: a no-MPI build
-      # tested with MPI would run ppn>1 cases against a binary that has none.
+      # without MPI, so compile coverage is unaffected, as is MPI coverage
+      # elsewhere. It stays in one place because a no-MPI build tested with
+      # MPI would run ppn>1 cases against a binary that has none.
+      #
+      # --test-all is deliberately NOT shared, and sits on the build only.
+      # It does two separate things (toolchain/mfc/test/test.py): at build
+      # time it adds post_process to the set of compiled binaries, and at
+      # test time it re-runs pre_process+simulation+post_process a second
+      # time for every case. The first is the compile coverage these lanes
+      # exist for; the second is roughly half the lane's ~120 min wall clock
+      # and duplicates runtime checking the GNU, Intel and self-hosted lanes
+      # already do. So: build with it, test without it. post_process is
+      # still compiled on all 15 releases, just not re-run here.
       - name: Build (NVHPC)
         if:   matrix.target == 'cpu'
         run: |
           docker exec nvhpc bash -c '
             source /etc/nvhpc-env.sh
-            /bin/bash mfc.sh test -v --dry-run -j $(nproc) $MFC_NVHPC_TEST_FLAGS
+            /bin/bash mfc.sh test -v --dry-run -j $(nproc) --test-all $MFC_NVHPC_TEST_FLAGS
           '
 
       - name: Build (NVHPC GPU)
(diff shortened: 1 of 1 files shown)


sbryngelson:
Fixed in 4acb93c.

The alert is correct, and it's a merge-order artifact rather than anything wrong with the new job. This branch is based on `master`, which has no `permissions:` block in `test.yml`, so every job there still inherits the repository-default `GITHUB_TOKEN` — and adding a job makes CodeQL flag the file again. #1825 adds that block, but relying on it landing first would leave a window where merging this PR alone reintroduces the alert.

So this PR now declares it itself:

```yaml
permissions:
  contents: read
  pull-requests: read
```

The `nvhpc` job needs only `contents: read` (checkout, then a `docker pull` from nvcr.io that uses no GitHub token). `pull-requests: read` is there for the `file-changes` job's `dorny/paths-filter`, exactly as in #1825.

The block is byte-identical to the one #1825 adds to this same file, so the two branches merge cleanly in either order — I confirmed with `git merge-tree` against their merge base: zero conflict markers.

One thing I did *not* do, which is worth a look separately: this grants `pull-requests: read` workflow-wide, so it reaches the multi-hour NVHPC and self-hosted Frontier/Phoenix jobs that have no use for it. Tighter would be workflow-level `contents: read` with a job-level override on `file-changes` alone, since that's the only job running `paths-filter`. I left it consistent with #1825 rather than diverging the two PRs mid-review — happy to tighten both if you'd prefer.

PR #1837, opened 2026-09-10
Validate job options before writing the job script

### LINKED ISSUE #1511: build/run/state cleanups: stale job script on invalid options, duplicated build parse, twin make_options/make_slug, quit shadows builtin
Four small issues across the run-orchestration modules. (b) is a real (cosmetic) bug; the rest are hygiene.

**(b) run.py writes the job script before validating options — real bug.** In `toolchain/mfc/run/run.py`:
```python
# run.py:181-183
__generate_job_script(targets, case)
__validate_job_options()
__generate_input_files(targets, case)
```
`__validate_job_options()` (run.py:20-34) raises `MFCException` on e.g. `--no-mpi` with >1 rank, `nodes <= 0`, or a malformed `--email`. Because it runs *after* `__generate_job_script` has already written the `.sh` to disk, an invalid invocation aborts leaving a **stale job script**. Fix: move `__validate_job_options()` above `__generate_job_script(...)`. (Pure reorder; validation has no side effects.)

**(a) ninja/make progress parse duplicated 4× in build.py.** `_run_build_with_progress` parses the ninja regex and the make regex in its non-streaming branch (build.py:73-92) and again, nearly verbatim, in its streaming branch (184-210). → extract `_parse_progress(line) -> (completed, total, percent, action) | None` used by both branches.

**(c) `make_options` and `make_slug` share an identical loop body** — `toolchain/mfc/state.py:42-52` and `54-64`. Both build the same `--[no-]<cli_k>` / `--{v}-gpu` option list; `make_slug` only additionally sorts by key and `"_".join`s. → have `make_slug` call `make_options`.

**(d) `common.quit` shadows the builtin** — `toolchain/mfc/common.py:205`:
```python
def quit(sig):
    os.kill(os.getpid(), sig)
```
→ rename to `s_quit`/`kill_self` and update callers.

**Why:** (b) leaves confusing artifacts after a failed run; (a)/(c) are copy-paste that must be edited in lockstep; (d) is a readability/lint nicety.

**Scope:** recommend **splitting** — (b) as its own small bug-fix PR (clearest, most valuable), and (a)+(c)+(d) bundled as a "run-orchestration tidy-up" PR. Behavior preserved in all four.

---
<sub>Filed from a repo-wide code-cleanliness review; verified against `master` @ `40dde5e`.</sub>

## Code references
- [`toolchain/mfc/run/run.py:181-183`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/toolchain/mfc/run/run.py#L181-L183) — generate-before-validate (bug)
- [`toolchain/mfc/build.py:73-92`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/toolchain/mfc/build.py#L73-L92) — ninja/make parse (1 of 4 copies)
- [`toolchain/mfc/state.py:42-64`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/toolchain/mfc/state.py#L42-L64) — make_options / make_slug twins
- [`toolchain/mfc/common.py:205`](https://github.com/MFlowCode/MFC/blob/40dde5e8701faf09813b910f9ddf6651fe4c673c/toolchain/mfc/common.py#L205) — quit shadows builtin

Mohit-Ak:
`run()` currently does its option validation *after* it has already rendered and written the job script:

```python
__generate_job_script(targets, case)
__validate_job_options()
__generate_input_files(targets, case)
```

`__validate_job_options()` raises for `--no-mpi` with more than one rank, `nodes <= 0`, `tasks_per_node <= 0`, or a malformed `--email`. Because it runs late, any of those aborts the run but leaves a freshly written `<name>.sh` in the case directory — and, since `--clean` and `build()` also run before it, a rejected invocation has by then already wiped the previous run's outputs and possibly triggered a build.

This moves `__validate_job_options()` to the top of `run()`, right after the targets are resolved and before the case file is loaded (loading executes `case.py`, so it's a side effect too). Validation only reads `ARG(...)`, so the reorder has no other effect; valid invocations behave exactly as before.

**Testing.** Added `toolchain/mfc/run/test_run.py`, which drives `run()` under a sandboxed `state.gARG` with each rejected option combination and asserts that `build`, `case.clean()`, the job-script generator and the input-file generator were all *not* called, plus a control case that a valid invocation still reaches generation. On pristine `master` the four rejection tests fail (`AssertionError: job script was written before option validation`); with this change all five pass.

```
$ cd toolchain && python -m pytest mfc/run/test_run.py -q
5 passed
$ python -m pytest . -q          # the ./mfc.sh lint suite
375 passed, 8 warnings, 4 subtests passed in 8.68s
$ ruff check . && ruff format --check mfc/run/run.py mfc/run/test_run.py   # ruff 0.6.5
All checks passed! / 2 files already formatted
```

This is part (b) of #1511 only, per the issue's own suggestion to split it from the (a)/(c)/(d) tidy-ups.

Fixes #1511

### DIFF
--- a/toolchain/mfc/run/run.py
+++ b/toolchain/mfc/run/run.py
@@ -158,6 +158,14 @@ def __execute_job_script(qsystem: queues.QueueSystem):
 
 def run(targets=None, case=None):
     targets = get_targets(list(REQUIRED_TARGETS) + (targets or ARG("targets")))
+
+    # Reject invalid job options before anything has side effects: loading the
+    # case executes case.py, and build, --clean and the job script/input file
+    # generation all touch the build tree or the case directory. A run that is
+    # going to be refused should not leave a stale job script (or a wiped
+    # output directory) behind.
+    __validate_job_options()
+
     case = case or input.load(ARG("input"), ARG("--"))
 
     build(targets)
@@ -197,7 +205,6 @@ def run(targets=None, case=None):
             cons.print(f"  [dim]MPI: {ARG('nodes')} nodes × {ARG('tasks_per_node')} tasks/node = {ARG('nodes') * ARG('tasks_per_node')} total ranks[/dim]")
 
     __generate_job_script(targets, case)
-    __validate_job_options()
     __generate_input_files(targets, case)
 
     if verbosity >= 2:
--- a/toolchain/mfc/run/test_run.py
+++ b/toolchain/mfc/run/test_run.py
@@ -1,0 +1,120 @@
+"""Tests for run.run() option validation ordering (issue #1511).
+
+An invalid invocation -- ``--no-mpi`` with more than one rank, ``nodes <= 0``,
+a malformed ``--email`` -- must be rejected before run() touches the case
+directory. Previously the job script was rendered and written to disk first,
+so a rejected run left a stale ``<name>.sh`` behind (and, with ``--clean``, had
+already wiped the previous run's outputs).
+"""
+
+import types
+import unittest
+from contextlib import ExitStack
+from unittest.mock import Mock, patch
+
+from .. import state
+from ..common import MFCException
+from . import run as run_mod
+
+
+def _fake_target(name):
+    return types.SimpleNamespace(name=name)
+
+
+def _fake_case(clean):
+    return types.SimpleNamespace(params={}, clean=clean)
+
+
+class _StateSandbox(unittest.TestCase):
+    def setUp(self):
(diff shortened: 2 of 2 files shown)


sbryngelson:
seem unimportant

PR #1843, opened 2026-09-11
Make the IB force-reduction receive buffers device resident

### LINKED ISSUE #1840: Moving IB + MPI on Cray OpenMP offload: recv_forces_snap not in present table (s_communicate_ib_forces)
AI generated but was a real failure...

## Summary

With a moving immersed boundary and more than one MPI rank, the OpenMP-offload build on Frontier (Cray CCE) aborts at the first time step in `s_communicate_ib_forces` (`src/simulation/m_ibm.fpp`). The ranks that do **not** own the body fail with

```
ACC:            find_in_present_table failed for 'recv_forces_snap(:,:)' (0x14542ec0-0x1467f540) from .../src/simulation/m_ibm.fpp:1338
ACC: libcrayacc/acc_runtime.c:703 CRAY_ACC_ERROR - Variable not found in present table
srun: error: frontier10177: tasks 1-3: Exited with exit code 1
```

Master `33ad77a6`, `./mfc.sh build --gpu mp --case-optimization`, `source ./mfc.sh load -c f -m g` (cpe/25.03, rocm/6.3.1), 1 node, 4 ranks. The same case runs on the OpenACC build (`--gpu` without `mp`).

## Cause

`recv_forces_snap`, `recv_torques_snap`, `recv_ids`, `recv_ft` are module allocatables that are allocated with a plain `allocate` (`m_ibm.fpp:100`) rather than `@:ALLOCATE`, so they never enter the device present table. They are then handed to a kernel through `copyin='[recv_ft, recv_ids]'` and `copy='[forces, torques, recv_forces_snap, recv_torques_snap]'` (`m_ibm.fpp:1338`). Under the OpenMP macro layer on Cray this becomes a lookup of a non-present allocatable and the runtime aborts. The sibling send arrays `send_ids`/`send_ft` are `@:ALLOCATE`d and updated with `GPU_UPDATE`, so the receive side is just inconsistent with them.

## Reproducer

2D flat plate (`patch_ib(1)%geometry = 3`, `moving_ibm = 1`, analytic `vel`/`angular_vel` in `t`), `ib_state_wrt = "T"`, 4 ranks, OpenMP offload on Frontier. `examples/2D_tumbling_rectangle` with `-n 4` should show the same thing on the mp build.

## Fix that works for me

Make the four receive arrays device-resident and update them explicitly after the host writes:

```diff
-            allocate (recv_forces_snap(size(patch_ib), 3), recv_torques_snap(size(patch_ib), 3), recv_ids(size(patch_ib)), &
-                      & recv_ft(6, size(patch_ib)))
+            @:ALLOCATE(recv_forces_snap(size(patch_ib), 3), recv_torques_snap(size(patch_ib), 3), recv_ids(size(patch_ib)), &
+                & recv_ft(6, size(patch_ib)))
...
                 recv_forces_snap = 0._wp
                 recv_torques_snap = 0._wp
+                $:GPU_UPDATE(device='[recv_forces_snap, recv_torques_snap]')
...
                         call MPI_UNPACK(ib_force_recv_buf, buf_size, unpack_pos, recv_ft, 6*recv_count, mpi_p, MPI_COMM_WORLD, ierr)
-                        $:GPU_PARALLEL_LOOP(private='[i, j]', copyin='[recv_ft, recv_ids]', copy='[forces, torques, &
-                                            & recv_forces_snap, recv_torques_snap]')
+                        $:GPU_UPDATE(device='[recv_ids, recv_ft]')
+                        $:GPU_PARALLEL_LOOP(private='[i, j]', copy='[forces, torques]')
...   (same two-line change at the second unpack site in the back-propagation phase)
...
-            deallocate (recv_forces_snap, recv_torques_snap, recv_ids, recv_ft)
+            @:DEALLOCATE(recv_forces_snap, recv_torques_snap, recv_ids, recv_ft)
```

With this the 4-rank moving-IB case runs on the mp build. Happy to open a PR if that shape looks right.

@danieljvickers this is in the neighborhood force reduction, so tagging you.

sbryngelson:
Fixes #1840.

`s_communicate_ib_forces` hands `recv_forces_snap`, `recv_torques_snap`, `recv_ids` and `recv_ft` to GPU kernels via `copy`/`copyin`, but they are allocated with a plain `allocate` and so are never present on the device. Under Cray OpenMP offload this aborts at the first time step for any moving immersed boundary run on more than one rank:

```
ACC: find_in_present_table failed for 'recv_forces_snap(:,:)' from .../m_ibm.fpp:1338
ACC: libcrayacc/acc_runtime.c:703 CRAY_ACC_ERROR - Variable not found in present table
srun: error: frontier10177: tasks 1-3: Exited with exit code 1
```

The send-side twins (`send_ids`, `send_ft`) are already `@:ALLOCATE`d and pushed with `GPU_UPDATE`, so this only makes the receive side consistent with them: `@:ALLOCATE`/`@:DEALLOCATE`, an explicit `GPU_UPDATE(device=...)` after each host write, and the corresponding `copy`/`copyin` clauses dropped.

**Testing.** Frontier, `--gpu mp` (cpe/25.03, rocm/6.3.1), 2D moving flat plate with `ib_state_wrt`, 4 ranks: aborts at step 1 on master, runs to completion with this patch. An OpenACC build of the same case was unaffected before and after. `./mfc.sh precheck` passes.

One file, 8 insertions, 6 deletions.


sbryngelson:
Worked through the three Copilot comments.

**The two full-array update comments are right and are fixed in `29034b0e0`.** `size(patch_ib)` is `num_local_ibs_max = 2000`, so each `GPU_UPDATE` pushed about 104 KB (`recv_ft` at 6x2000x8 B plus `recv_ids` at 8 KB) no matter how many bodies actually arrived, once per neighbour per direction per step. Both sites now update what was written:

```fortran
$:GPU_UPDATE(device='[recv_ids(1:recv_count), recv_ft(:,1:recv_count)]')
```

Safe because the kernel below only reads `1:recv_count`; the untouched tail is never referenced.

I did not add the suggested `recv_count == 0` guard. The loop is already a no-op at zero, a zero-length section is legal, and the surrounding code does not guard its own slices (`GPU_UPDATE(host='[patch_ib(1:num_ibs)]')`), so the branch would only be there for a case that cannot misbehave.

**The `recv_forces_snap` / `recv_torques_snap` comment does not apply.** It is conditional on host code reading those arrays after the kernel, and nothing does: they are module-private, appear nowhere outside `m_ibm.fpp`, and their only accesses are the host zeroing plus device update at 1311-1313 and the read/write inside the kernel at 1345-1348. There is no host read between the kernel and the `@:DEALLOCATE`. Keeping them device resident is the point - the running snapshot now accumulates across receive iterations instead of round-tripping through the host on every launch. Adding a `GPU_UPDATE(host=...)` would just reintroduce the transfer the PR removes.

Verified on `AFBACA70` (`3D -> 2 MPI Ranks -> IBM Sphere`) plus `CE232828`, `A2036630` and `16574715`, amdflang OpenMP offload on MI210: 4 passed, 0 failed. `AFBACA70` is the case that matters - this path only runs when `num_procs > 1`, so the single-rank IB tests never touch it.

`b2ef85321` is unrelated: master's Lint Toolchain has been red since #1855 and this branch has that merge, so the three doc-reference fixes are here to get the lint gate green. Same change as #1861; it will reconcile when that lands.

PR #1844, opened 2026-09-11
Write vel(2) to 2D probe files and drop the per-step print

### LINKED ISSUE #1841: 2D probe output omits vel(2) and prints to stdout every step (s_write_probe_files)
## Summary

For a plain 2D case (no bubbles, no hypoelasticity) `s_write_probe_files` in `src/simulation/m_data_output.fpp` writes only `(t, rho, vel(1), pres)` to `D/probe<i>_prim.dat` — the second velocity component is dropped — and it also `print`s a line to stdout on every time step:

```fortran
                    else
                        write (i + 30, '(6X,F12.6,F24.8,F24.8,F24.8)') nondim_time, rho, vel(1), pres
                        print *, 'time =', nondim_time, 'rho =', rho, 'pres =', pres
                    end if
```

(master `33ad77a6`, the `p == 0` branch). The 3D branch writes `vel(1), vel(2), vel(3)`, and the 2D bubble and hypoelastic branches write `vel(2)` too, so this is just the plain-2D branch being incomplete. The `print` floods the job log (one line per step per probe) and looks like leftover debugging.

## Reproducer

Any 2D case with `probe_wrt = "T"`, e.g. `examples/2D_ibm_airfoil` with a probe added: the probe file has 4 columns and stdout gets `time = ... rho = ... pres = ...` every step.

## Suggested fix

```diff
                     else
-                        write (i + 30, '(6X,F12.6,F24.8,F24.8,F24.8)') nondim_time, rho, vel(1), pres
-                        print *, 'time =', nondim_time, 'rho =', rho, 'pres =', pres
+                        #:if not MFC_CASE_OPTIMIZATION or num_dims > 1
+                            write (i + 30, '(6X,F12.6,F24.8,F24.8,F24.8,F24.8)') nondim_time, rho, vel(1), vel(2), pres
+                        #:endif
                     end if
```

(the fypp guard mirrors the neighbouring bubble/hypoelastic branches, since `vel(2)` does not exist under 1D case optimization). This changes the 2D probe file layout from 4 to 5 columns, so any golden probe files for 2D probe tests would need regenerating. Happy to PR.

sbryngelson:
Fixes #1841.

For a plain 2D case (no bubbles, no hypoelasticity), `s_write_probe_files` wrote only `(time, rho, vel(1), pres)` — the second velocity component was dropped — and it also `print`ed a line to stdout on every time step for every probe:

```fortran
write (i + 30, '(6X,F12.6,F24.8,F24.8,F24.8)') nondim_time, rho, vel(1), pres
print *, 'time =', nondim_time, 'rho =', rho, 'pres =', pres
```

The 3D branch writes all three components and the 2D bubble and hypoelastic branches both write `vel(2)`, so this looks like the plain-2D branch being incomplete rather than intentional. The `print` floods the log of any long run: one line per probe per step.

The fypp guard matches the neighbouring branches, since `vel(2)` does not exist under 1D case optimization.

**Heads up for reviewers:** this changes the 2D probe file from four columns to five, so any 2D probe golden files need regenerating. I have not run a 2D probe case with this patch — I hit the missing component while post-processing and fixed it, but my later 2D runs used no probes and my 3D runs take a different branch. Worth a quick check before merge.

`./mfc.sh precheck` passes.


sbryngelson:
Now exercised at runtime, which I had flagged above as missing.

2D case (flat plate in a freestream, 3 probes, `--gpu mp`, 4 ranks on Frontier). The probe file has five columns where it previously had four, and the second velocity component reads correctly:

```
# t                rho             u               v               p
0.000833       1.00000000      1.00000000      0.00000000     17.85714286
```

`u = 1`, `v = 0` and `p = 17.857` are the prescribed freestream, so the new column is both present and right. The stdout flood is gone.

Reminder for whoever merges: 2D probe golden files need regenerating, since the file goes from four columns to five.

PR #1846, opened 2026-09-11
Record immersed-boundary force history to one shared file

sbryngelson:
Records each immersed body's force, torque and kinematics as the run proceeds, rather than only at save intervals, so a body's history can be read without post-processing whole restart dumps.

`D/ib_forces.dat`, one shared fixed-width text file:

```
ib_id time Fx Fy Fz Tx Ty Tz vx vy vz wx wy wz ax ay az xc yc zc
```

Column names and the record geometry go in the sibling `D/ib_forces.hdr`; a header line inside the file would shift every byte offset the writer computes.

## Why it is laid out this way

The obvious shape - a file per body, appended each step - does not survive contact with a real run. Each append costs an `inquire`, an `open` and a `close`, per body, per rank, per step:

| ranks | bodies/rank | opens/step | metadata ops/step |
| --- | --- | --- | --- |
| 64 | 10 | 640 | 1,920 |
| 1000 | 10 | 10,000 | 30,000 |
| 1000 | 100 | 100,000 | 300,000 |
| 4000 | 100 | 400,000 | 1,200,000 |

A Lustre MDS sustains order 1e4-1e5 operations per second, so at a thousand ranks this is seconds per step of pure metadata, and it degrades the filesystem for every other job on the machine. Buffering does not rescue it: with moving bodies the buffer has to be flushed before ownership can change hands, which is every step, so it never accumulates more than one row per body.

Instead every record goes straight to a computed offset in one file that is opened once for the whole run:

```fortran
disp = (row*num_gbl_ibs + gbl_patch_id - 1) * IB_REC_LEN
call MPI_FILE_WRITE_AT(ib_hist_file, disp, rec, IB_REC_LEN, MPI_CHARACTER, status, ierr)
```

with `row = t_step/ib_force_stride`. That is one file open for the run rather than `ranks * bodies` per step, and it drops the buffer, the sentinel bookkeeping that tracked which rows had been written, the O(n^2) scan over the buffer per distinct body, and the flush-before-handoff entirely - a record lands at its own offset, so a rank giving up a body has nothing left to reorder.

This works on a *text* file because Fortran's `ES` descriptors are fixed width. Checked before relying on it: `'(I10,19(1X,ES17.9E3))'` emits exactly 352 characters for ordinary values, for negatives with three-digit exponents, and for NaN and Inf, which right-justify in their fields rather than widening them. `IB_REC_FMT`, `IB_REC_BODY` and `IB_REC_LEN` sit together with a note that they have to be edited together, because widening the format alone would shear the file.

The one hazard is that unwritten regions of a sparsely written file read back as NUL bytes, which would corrupt the text. It does not arise because ownership is a half-open interval test, so every body is owned by exactly one rank at every step and no slot is skipped - verified below rather than assumed.

## What this gives you

- `numpy.loadtxt('D/ib_forces.dat')` with no decode step; one body's history is `d[d[:,0] == id]`.
- Output that does not depend on the decomposition: offsets key on the global body id, so one rank and a thousand ranks produce the same bytes.
- Sorted by construction, step-major then body, so a body's history is a fixed stride.
- `ib_force_stride` subsamples long runs.

## Robustness

The offsets assume `IB_REC_FMT` emits exactly `IB_REC_BODY` characters, which is checked at open time rather than left as a comment - a format one character wider would otherwise shear every record after the first with no other symptom:

```fortran
write (probe, IB_REC_FMT) 0, [(0._wp, i=1, 19)]
@:PROHIBIT(len_trim(probe) /= IB_REC_BODY, "IB force record width disagrees with IB_REC_BODY; ...")
```

Confirmed by deliberately widening `I10` to `I11`: the run aborts on that check instead of writing a sheared file.

`MPI_MODE_CREATE` does not truncate, so the file is deleted before it is opened, as the `ib_state` writer already does. Without that, a shorter run following a longer one in the same directory keeps the earlier tail past its last record.

## Testing

- 11 IB goldens regenerated and passing (`5A22B45F E085CC5A D6794F4C 4BED9896 C8AD6271 49893269 135F548B B317404C F200F862 DA8FCD2D E5B66084`), CPU, gfortran.
- Record width measured across ordinary, negative-with-three-digit-exponent, NaN and Inf values: 352 characters in every case.
- File structure: 200 records of 353 bytes, every byte printable or newline, every record newline-terminated, no NUL holes.
- The same case at 1 rank and at 2 ranks produces byte-identical files (70,600 bytes each).
- `numpy.loadtxt` returns `(200, 20)`; body 1's 50 rows have strictly increasing time.

- Serial build (`--no-mpi`, direct-access branch) produces a file of the same size and structure as the MPI build.

Not covered, and worth weighing:

- No run at the scale the layout is designed for. The argument there is arithmetic - one open for the run against `ranks * bodies` per step - rather than measurement.
- Offset arithmetic inside a text writer is unusual, and it rests on a format-width invariant. The check above makes a bad edit loud rather than silent, but it is still a construct that has to be understood before it is changed.
- The sibling `.hdr` file is a new convention here. The alternative, a header line, would shift every offset after it.
- Near-zero columns (a cancellation-level `Fy` around 1e-18 against a column scaled at 1e-3) compare on absolute error first against the 1e-12 default tolerance, so they do not carry the cross-compiler fragility that bit the prescribed-kinematics goldens. Checked rather than assumed.

## Note on the earlier revision

This replaces the per-body writer this PR originally carried. The review comments on that code about the concurrent-writer race, the `nint`/sentinel scheme and the O(n^2) flush are resolved by the code no longer existing; the `ib_force_stride` validator fix from that round is kept. `pack.py` gains `ib_forces` alongside `probe` and `integral` as a multi-column time series, which is what it is.


sbryngelson:
Two updates pushed after running this on a real case.

**A correctness fix.** My original note claimed that because each buffered row carries its own body id, a body changing owner mid-run needed no special handling. That is true for *which file* a record lands in, but not for its **order**: a rank that stops owning a body kept its buffered records and wrote them at shutdown, after the new owner's newer ones. On a 64-rank flapping case the recorded time jumped from 3.24 back to 0 partway through the file. Fixed by flushing before the ownership handoff, and the comment now says the right thing.

**The column separator.** `ES18.10E3` is exactly wide enough for a negative value with a three-digit exponent, so adjacent columns ran together and the file could not be read as whitespace-separated. Now uses an explicit separator.

Both were found by using the output rather than by review, which is the argument for landing this with a case that exercises it. The branch builds clean on Frontier with `--gpu mp`; I have now made a habit of building each of these before pushing, after learning that `precheck` lints but does not compile.

danieljvickers:
It seems to me that `ib_force_stride` should likely default to a null value, like -1, allowing users to opt out of this feature. This parameter also looks like it is going to get quite messy at exascale, where we will have multiple ranks simultaneously trying to write and append to thousands of files. A way to disable this is not only desirable, but likely required.

sbryngelson:
yep i know. i'm working on this one still

sbryngelson:
Both fair. The second point landed right after I changed the design, so let me lay out where it actually stands.

**Opt-out — you're right, and it's still missing.** `ib_force_stride` defaults to 1 and the validator rejects anything `< 1`, so today the history is on whenever `ib_state_wrt` is. Needs a way off; I'll add one.

**Writes at scale — this isn't file-per-body any more.** `6c4cc9c3` replaced that with a single shared `D/ib_forces.dat`. Every record is fixed width (353 bytes), so a rank computes a byte offset from (step, global body id) and writes straight there with `MPI_FILE_WRITE_AT`. No appends, no gather, no merge step, and the file comes out byte-identical however the domain is decomposed. It's opened once for the run, not per step.

At 1000 ranks holding 100 bodies each that's one open for the whole run plus one write per body per recorded step, against the inquire + open + close per body per rank per step the old shape cost — roughly 3e5 metadata ops per step. That cost is exactly what you're describing, and it's what prompted the rewrite, so I think this half is already answered. Shout if you see a failure mode I've missed in the offset scheme.

Separately, I just pushed `3e29729`, which is what the red CI was about. Skipping the `t_step_start` row left row 0 unwritten rather than shifting the file up. Nothing pre-fills the file — both open paths create it empty and every write lands at a computed offset — so that row stayed a sparse hole, which reads back as NUL rather than blanks, and the packer's `float()` choked on it. Hence all 11 IBM tests failing on every platform, CPU included; it was I/O arithmetic, nothing GPU about it.

Rows now count from the first step actually recorded. Restart was the worse case under the old numbering: with `t_step_start=20649` every row beneath it was a hole, and the offset formula documented in `case.md` resolved to the wrong record for anyone using it as written. Both fixed, goldens regenerated, 11/11 passing locally.

sbryngelson:
Pushed the opt-out (`1b5929c`), which was the outstanding piece from @danieljvickers' review.

**`ib_force_wrt`, default off.** A sentinel stride value would have been the odd one out here: `probe_wrt`, `prim_vars_wrt`, `fft_wrt` and `ib_state_wrt` are all logicals defaulting to false, so the history now works the same way. `ib_force_stride` keeps its meaning for runs that do want the history but not every step.

It also decouples two things that should not have been joined. Previously, asking for the SILO point mesh via `ib_state_wrt` silently enrolled the run in per-step force I/O. `ib_force_wrt` gates the history, `ib_state_wrt` keeps gating the point mesh, and neither implies the other.

The flag is on in the generated IBM test cases and in the three IBM examples that were relying on the old coupling, so the writer, the byte offsets and the row numbering stay under CI instead of only being exercised by hand.

Verified both directions on CPU: with the flag set, all eleven tests carrying `ib_forces.dat` in their goldens pass; with it absent, a case that ran 49 time steps wrote no force file at all.

---

One thing worth splitting off. While checking that default-off really suppresses the file, I broke three IBM examples on purpose by removing the flag. Only `ibm_viscous_drag_over_cylinder` failed, on a line count. The other two passed even though their goldens still list `D/ib_forces.dat` and the file was no longer produced at all.

So an expected output file can disappear entirely and the suite can still report success. That is independent of this PR and I have not filed it yet; happy to open an issue if it is not already known.

sbryngelson:
Correction to my previous comment: the test-harness concern I raised at the end of it is wrong, and I should not have reported it without testing it directly.

I went to file it as an issue and checked the mechanism first. `compare()` in `toolchain/mfc/packer/tol.py` does guard this — it compares entry counts before anything else, and separately returns `No reference to <file> in the candidate's pack` when an entry is missing. Running the controlled experiment rather than inferring from the earlier run, `ibm_flameholder` with the force history disabled and its golden still listing `D/ib_forces.dat` fails exactly as it should:

```
Error: Test tests/F200F862: 2D -> Example -> ibm_flameholder: Line count does not match.
```

What I originally saw was two of those examples passing in a run where their case directories still held output from an earlier `--generate` pass, so the previous `ib_forces.dat` was still on disk and got packed. That is a stale-artifact effect in my own local working directory, not a hole in the suite — CI starts clean, so it would not arise there.

The commit message on `1b5929c` carries the same wrong note in its last paragraph. The code in that commit is unaffected; only the claim about the harness is wrong, and there is nothing to file.

PR #1847, opened 2026-09-11
Add prescribed kinematics for immersed boundaries

sbryngelson:
### Why

Moving immersed boundaries are driven either by two-way coupling or by analytic velocity expressions in the case file. For a parameter sweep or an ensemble the analytic path has two drawbacks:

- the expressions are compiled into `case.fpp`, so every distinct motion forces a rebuild of the simulation binary — a three-point Strouhal sweep is three builds;
- the motion is obtained by integrating the supplied rates, so a restart does not reproduce the same trajectory.

### What

`patch_ib(i)%kin_model` sets the body state directly from the current time at each Runge-Kutta stage:

- **1, hinged flapping** — roll about the lab *x* axis through the hinge and pitch about the body spanwise axis through the same hinge. Inputs: hinge, body-frame hinge-to-centroid offset, roll and pitch amplitudes, mean pitch, frequency, pitch phase lead, onset time, raised-cosine onset ramp.
- **2, smoothed pitch ramp and hold** — the AIAA low-Reynolds-number canonical maneuver (Eldredge et al. 2009, Ol et al. 2010), with nominal pitch rate and the log-cosh smoothing parameter as inputs.

Nothing is integrated, so restarts are exact and one case-optimized binary serves every parameter value and every ensemble member.

### A correctness fix that comes with it

The existing path advances the three angles componentwise from `angular_vel`, and then uses that same array as a vector in `omega x r` for the ghost-point velocity. Those two uses only agree when a single angle is changing. With the rotation composed as `R = Rx(phi) Ry(theta)`, the lab-frame angular velocity is `phi' e_x + theta' Rx(phi) e_y`. The new routine sets that vector directly. On a plate rolling 30 degrees while pitching 20 degrees the difference reaches 10 percent of the surface velocity.

### Testing

Against closed-form kinematics on a 3D plate: angles, centroid, centroid velocity and lab-frame angular velocity all match to 1e-15 at every step. The velocity carried by body cells adjacent to the fluid matches the rigid-body velocity to 8e-13, versus 4e-3 to 1e-1 if the Euler-angle rates are used as the angular velocity vector — so the check distinguishes the two. Also exercised on a 3D pitch-ramp validation case against published DNS. `./mfc.sh precheck` passes.

Restricted to 3D by the validator, since the in-plane 2D rotation goes through a different path.


sbryngelson:
Pushed `467a91525` for the `reldebug` lanes. The failure is not from this PR:

```
At line 1143 of file src/simulation/m_viscous.fpp
Fortran runtime error: Index '-2' of dimension 2 of array 'fd_coeff_y' below lower bound of 0
```

`s_compute_ib_forces` samples the viscous stress `fd_number` cells out from an interior cell, and `s_compute_viscous_stress_tensor` then looks up `fd_coeff_y` at that sample index, but the coefficients are allocated `0:n`. A body within `fd_number` of a domain boundary therefore reads past the array. That path is master's, and this PR touches neither `m_viscous.fpp` nor the force integration - `3D_ibm_flapping_plate` is just the first viscous IB case in the suite with a body near a boundary. Written up as #1856, same class as #1792. Verified with `./mfc.sh test --no-gpu --debug`: the error reproduces on `86A2C59B` without the change and all three cases pass with it.

The Pitch Ramp goldens are a separate matter and are **not** fixed yet. `18f123747` cut `kin_theta0` and `kin_pitch_rate` by a hundred (holding `t_p = theta0/rate = 0.025`, so the Eldredge ramp keeps its shape and its `a*t_p = 5` smoothing) because at the original amplitude the tip sweeps 1.5 cells over the 50 steps and perturbing `kin_smooth` by 5e-13 moved the step-50 field by 5e-3; at the reduced amplitude the same perturbation moves it by 2e-15. That was not enough. `ubuntu no-debug GNU` still reports

```
Max abs among failing: Candidate 2.5   Golden 1.30556685791813   (91% relative)
```

A candidate of exactly 2.5 against a golden of 1.31 is a cell still at its initial condition in one run and evolved in the other, i.e. solid in one and fluid in the other - a classification flip, not drift. The plate's y faces sit at exactly 0.30 and 0.70, which are cell centres on this 25-cell grid, and `f_is_inside_cuboid` decides with `<=`; the rotated coordinate reaching that comparison differs in its last bit between compilers. Perturbing the geometry does not test this, because it moves the face off the tie rather than changing how the tie is broken - which is why my earlier probe came back clean and I wrongly concluded the amplitude was the whole story.

The fix I would make next is to move the body off the grid so no face lands on a cell centre or boundary, then regenerate. Holding off until you say whether you want that or would rather drop the Pitch Ramp case, since `86A2C59B` already covers `kin_model = 1` and passes everywhere.

sbryngelson:
Worked through the three review comments. One is fixed, two do not hold - details so the reasoning is checkable rather than just asserted.

**Docstrings (`m_ibm.fpp:993`) - fixed in `32e85546`.** The brief did say `kin_model = 1` while the routine serves both models; it now names both and says the ramp holds phi at zero. Inserting the two new routines had also orphaned the `!> Compute pressure and viscous forces ...` brief, which was left attached to `f_log_cosh` and had taken `s_compute_ib_forces`'s away; that is restored.

**`kin_hinge`/`kin_offset` registered as `REAL` rather than `A_REAL` - not a mismatch.** `A_REAL` is `ParamType.ANALYTIC_REAL` (`definitions.py:583`), which is "may be given as an analytic expression", not "array real": `schema.py:38` maps it to `{"type": ["number", "string"]}`, the string being the expression. `vel(j)`/`angular_vel(j)` use it because a moving IB's velocity can be written as an expression of time. A hinge point and a body-frame offset are static geometry, so plain `REAL` is the right type, and `x_centroid`/`length_x` are registered the same way for the same reason.

There is no scalar-versus-array mismatch either: both sit inside the same `for j in range(1, 4)` loop, so `kin_hinge(1..3)` are three registered components exactly as `vel(1..3)` are. The Fortran side is `dimension(1:3)` and `m_mpi_proxy` broadcasts it as length 3, which matches.

**`t_stage` hard-coded for RK3 stage 2 - checked against the coefficients, and it is right.** Reading `rk_coef` (`m_time_steppers.fpp:426-433`): stage 1 produces `u1` at `t^n + dt`; stage 2 produces `(3/4)u^n + (1/4)(u1 + dt L)` at `t^n + dt/2`; stage 3 lands at `t^n + dt`. RK2's two stages both land at `t^n + dt`, and RK1's single stage at `t^n + dt`. So the `s == 2` special case and the `mytime + dt` fall-through are each correct for every `time_stepper` in the tree.

`mytime` is also the right base to add to: it holds `t^n` throughout `s_tvd_rk` and is advanced only afterwards, at `m_start_up.fpp:639`. And the first stage of the first step is covered separately by `s_ibm_setup`'s `t_init`, which reproduces `p_main.fpp:46-56` for both the `cfl_dt` and fixed-`dt` paths.

Deriving the stage time from a stored abscissa table would be more robust against a future integrator, and I would take that change if the stage times were ever wrong today - but they are not, and a table would have to be added and kept in step with `rk_coef`, so it trades one coupling for another. Leaving it, with the reasoning recorded here.

danieljvickers:
This PR adds 15 reals and an integer to the increasingly bloated patch_ib data type. Since this type is constantly packed and unpacked into arrays for MPI communicaiton, we should seek to not inflate the size more. In fact, I think we should be delting it. I agree this refactor is nice, but continually added elements to the patch_ib array that are often unused will cause problems in the future.

I sugest making this PR go the way of the STL models, where you create a separate data type that you point to for an IB. That concenses all of these values into a single integer and reduces the bloat.

PR #1850, opened 2026-09-11
Allow a GRCBC inflow to start up smoothly

sbryngelson:
*Written with Claude Code.*

A jet or a tunnel that starts from rest cannot be represented by a constant inflow when the start-up **is** the event of interest. This scales a GRCBC inflow velocity by

$$f(t) = f_0 + (1 - f_0)\left[1 + \tanh\left(6 (t - t_0)/\tau - 3\right)\right]/2$$

so it holds $f_0$ of the final velocity at $t_0$ and is within half a percent of it at $t_0 + \tau$, with $\tau$ = `bc_[x,y,z]%vel_in_ramp`, $t_0$ = `vel_in_t0`, $f_0$ = `vel_in_frac0`. A boundary with `vel_in_ramp = 0` is held constant exactly as before, so an existing case is bit-identical.

The ramp is evaluated **inside the CBC kernel** from `mytime`, which the time stepper already keeps on the device. Nothing is computed on the host and copied per Runge-Kutta stage.

## Rewritten after review — what this PR no longer contains

The branch had grown three problems, all now gone. It is one commit on `master` and builds all three targets plus `--gpu mp`.

1. **It did not compile.** The Dirichlet ramp lived in `src/common/m_boundary_primitives.fpp` and read `mytime`, which only `src/simulation` declares — but `src/common` is compiled into all three executables. `./mfc.sh build -t pre_process` failed with `ftn-297 ftn: ERROR S_DIRICHLET ... must be specified for data object "MYTIME"`. The Dirichlet path is dropped from this PR; it needs the time threaded in as an argument (per the contributing guide's rule that `src/common` takes stage-varying behavior as an explicit argument), which is a separate change and will be its own PR, complete, not a promise attached to this one.
2. **It deleted the GRCBC device update.** `$:GPU_UPDATE(device='[vel_in, vel_out, pres_in, pres_out, Del_in, Del_out, alpha_rho_in, alpha_in]')` had been removed from `s_initialize_cbc_module` — a silent wrong-answer regression for every existing GRCBC GPU case, not just ramped ones. Restored, along with the matching `@:DEALLOCATE` that had gone with it.
3. **The three new members were read on device but never put there.** `bc_[x,y,z]%vel_in_ramp/_t0/_frac0` are read in the CBC kernel. Under OpenACC only named `bc_x%` members are `GPU_DECLARE`d, and nothing updated them, so the kernel read uninitialized device memory. Added to both the declare list and the `s_start_up` update block, mirroring how `grcbc_in` is handled.

## How the validation is obtained

`vel_in_ramp = 0` is the default and short-circuits to unity in `f_vel_ramp`, so every existing case is untouched — the pure function returns `1._wp` without reading the other two members.

For a ramped case, the check that matters is that the boundary delivers the prescribed function rather than something the CBC relaxation has reshaped. Put a probe in the first cell off the inflow plane and compare its velocity against $U_j f(t)$ over the ramp: on a Mach 0.1 round jet at $\tau = 0.5$, sampled through $2.5\tau$, the worst-case error is under 8 % of $U_j$, which is the tolerance in that case's `check.py`. A constant-inflow run of the same case reproduces its previous history bit-for-bit.

## Files

`m_derived_types` (three members on `int_bounds_info`), `m_boundary_primitives` (the pure `f_vel_ramp`), `m_cbc` (apply it), `m_global_parameters`/`m_start_up` (defaults, device residency), `m_mpi_proxy` (broadcast), `definitions.py`/`case_validator.py` (registration, and prohibitions on a negative duration, `f_0` outside [0, 1], and a ramp with no `grcbc_in` to act on), `case.md`.

### DIFF
--- a/docs/documentation/case.md
+++ b/docs/documentation/case.md
@@ -1328,6 +1328,9 @@ The entries labeled "Characteristic." are characteristic boundary conditions bas
 | `bc_[x,y,z]%%grcbc_out`        | Logical | Enable grcbc for subsonic outflow (pressure)|
 | `bc_[x,y,z]%%grcbc_vel_out`    | Logical | Enable grcbc for subsonic outflow (pressure + normal velocity) |
 | `bc_[x,y,z]%%vel_in`           | Real Array | Inflow velocities in x, y and z directions |
+| `bc_[x,y,z]%%vel_in_ramp`      | Real | Duration of a smooth start-up of the inflow velocity (0 = none) |
+| `bc_[x,y,z]%%vel_in_t0`        | Real | Time at which that ramp begins |
+| `bc_[x,y,z]%%vel_in_frac0`     | Real | Fraction of the final inflow velocity held before the ramp |
 | `bc_[x,y,z]%%vel_out`          | Real Array | Outflow velocities in x, y and z directions |
 | `bc_[x,y,z]%%pres_in`          | Real    | Inflow pressure |
 | `bc_[x,y,z]%%pres_out`         | Real    | Outflow pressure |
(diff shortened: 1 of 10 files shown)


sbryngelson:
Pushed a fix: the time stepper called `s_update_inflow_ramp` without importing `m_cbc`, so the simulation target failed to link. `precheck` lints but does not compile, so it passed and I only caught this on the first real build. The branch now builds clean on Frontier with `--gpu mp`. Apologies for the noise — worth reviewing after the rebuild.

sbryngelson:
Good catch on the per-stage transfer — you're right, and it's fixed in a0fd312b.

**What it was doing:** `s_update_inflow_ramp` ran once per Runge–Kutta stage (3× per step), computed the ramp factor on the host, and issued a `GPU_UPDATE` of `vel_in` and `bc_vel_ramp`. The payload is 9 + 3 doubles = **96 bytes**, so this was never bandwidth — it's three transfer *latencies* per step, order 10–20 µs each on MI250X. Against a 300 ms step that's 0.02%; against a 1 ms test case it's several percent. Not a good trade to ship upstream on the grounds that it's cheap on one problem.

**Why it was avoidable:** `mytime` is already placed on the device by the time stepper one line earlier, and `bc_x/y/z` are already `GPU_DECLARE`d. The ramp factor is a closed-form `tanh` of those, so it can be evaluated where it's consumed.

**What replaced it:** one `pure` device-callable `f_vel_ramp(tau, t0, frac0, t)`. Gone: the host routine, the `GPU_UPDATE`, the module-level `bc_vel_ramp` mirror, and the shadow `vel_in_final` array that existed only to keep `vel_in` in sync. Net simpler.

**The trade, stated plainly:** it now evaluates a `tanh` per boundary cell per stage instead of once per stage. That's cheap on a GPU and confined to boundary faces, but it is a different trade. Two things worth a reviewer's eye:

- the `GPU_ROUTINE(parallelism='[seq]')` decoration should make it inline on Cray, but the OpenACC path deserves a check that it doesn't become a real call per cell;
- an unramped case is untouched either way — `f_vel_ramp` returns unity when `vel_in_ramp <= 0`, where before the whole block was skipped by a branch.

Builds clean on CCE.

PR #1858, opened 2026-09-12
Abort when post_process is given a different grid than the restart files

sbryngelson:
`post_process` checks that the restart files **exist** but never that they hold the grid the case file asks for. Give it a case whose resolution no longer matches the run and it reads past the end of every file, post-processes the overrun, and **exits 0 with NaN-filled output**. Nothing reports a problem until someone plots it — or until a diagnostic built on that output starts returning NaN and gets blamed instead, which is how I found it.

### Reproduction

A 349 M cell run (`m = 1644`) post-processed with a case file that had since moved to `m = 2056`:

```
simulation.inp:    m = 1644     <- wrote the restart files
post_process.inp:  m = 2056     <- read them
```

681 M cells read from files holding 349 M. Exit code 0, silo written, every field NaN.

### The check

`x_cb.dat` holds exactly `m_glb + 2` cell boundaries, so its size states which grid wrote the restart. One `inquire` on a file already being opened, before any bulk read:

```
Restart grid mismatch: this case has m = 2056 but
.../restart_data/x_cb.dat was written with m = 1644. Post-processing must use the
same grid as the run that wrote the restart files, or it reads past the end of
every file and writes NaN.
```

Verified against the failing data: 2058 boundaries would need 16464 bytes, the file holds 13168, so it aborts where it previously ran to completion.

Builds clean (`./mfc.sh build --no-gpu -t post_process`).


sbryngelson:
Rebased onto `master` as one commit. The branch had been cut from a working branch and was dragging along stale, divergent copies of #1850, #1846, #1847 and the probe changes — including a `bc_vel_ramp` module array with `GPU_DECLARE(create=...)` and no device update, which under OpenACC would multiply ghost-cell velocities by uninitialized device memory on any Dirichlet boundary. That code does not exist in #1850, which evaluates the ramp on device from `mytime` through a pure function; it was an older draft that should never have been on this branch. Gone now — the PR is the 19-line check and nothing else.

Two real bugs in the check itself, both fixed:

**Hardcoded 8 bytes.** `x_cb.dat` is written with `mpi_p`, and `mpi_p` follows `wp`, so a `--single` build writes four-byte reals and every single-precision post_process run would have aborted, reporting half the true `m`. The `--single` CI lane runs `--test-all`, which includes post_process, so this would have failed CI. Now taken from `storage_size(0._wp)/8`.

**`down_sample` broke.** `m_start_up.fpp:83` reduces `m` to `int((m+1)/3) - 1` before `m_glb = m`, while `x_cb.dat` still holds full-resolution boundaries, so the strict equality fired spuriously. `examples/3D_IGR_33jet/case.py` sets it and no test covers it. The read is strided by 3 and touches `stride*(m_glb + 1) + 1` boundaries, so down-sampling needs *more* of the file, not less; the check now compares against that, and keeps the exact equality only for the un-strided read — three source grids of different size can down-sample to the same `m_glb`, so exact is not available there.

`./mfc.sh build -t post_process` is clean. Dropped the `Claude-Session` trailer.

PR #1861, opened 2026-09-12
Fix three doc-reference lint hits blocking every PR

sbryngelson:
Master's Lint Toolchain has been red since #1855 (595368445), and because PR checks run against the merge with base, it fails every open PR rather than just master. #1847, #1843 and #1839 are all red on it right now.

```
docs/documentation/contributing.md:220 single-backtick span `$:GPU_UPDATE(device='[name]')` contains a
  single quote, which Doxygen treats as ending the span
docs/documentation/gpuParallelization.md:877 single-backtick span `GPU_ROUTINE(parallelism='[seq]')`
  contains a single quote, which Doxygen treats as ending the span
docs/documentation/testing.md:16 links to #selection-and-execution-pitfalls but no
  {#selection-and-execution-pitfalls} anchor exists
```

Three lines: double backticks around the two spans that contain a quote, and the anchor added to the `### Selection and Execution Pitfalls` header the link already points at.

`python3 toolchain/mfc/lint_docs.py` and `./mfc.sh precheck` both pass on this branch.

Split out of #1859 so it can go in on its own - that PR carries an unrelated `m_viscous.fpp` fix and will sit in review, and this is blocking everyone's CI meanwhile.


sbryngelson:
Superseded by #1857, which covers these three plus the rest of the doc-lint cleanup.

PR #1862, opened 2026-09-12
Give the finite-difference coefficients the buffer range the IB force asks for

### LINKED ISSUE #1860: Immersed-boundary force reads finite-difference coefficients out of bounds, giving a force that depends on the MPI decomposition
**This issue originally attributed the symptom to grid stretching in the force stencil. That diagnosis was wrong and is corrected below; the stretching correlation was a coincidence of which runs happened to put a subdomain edge through the body.**

## What happens

`s_compute_ib_forces` (`src/simulation/m_ibm.fpp`) evaluates the viscous stress tensor at every cell its own stencil reaches:

```fortran
do l = -fd_number, fd_number
    call s_compute_viscous_stress_tensor(viscous_stress, q_prim_vf, dynamic_viscosity, i + l, j, k)
```

`s_compute_viscous_stress_tensor` (`src/simulation/m_viscous.fpp:1141`) then indexes the coefficient array at the cell it was handed:

```fortran
velocity_gradient_tensor(l, 1) = velocity_gradient_tensor(l, 1) + fd_coeff_x(r, i)*q_prim_vf(...)%sf(i + r, j, k)
```

so the coefficients are read at `i - fd_number` through `i + fd_number`. They are allocated `0:m` (`src/simulation/m_derived_variables.fpp:43`). A body cell within `fd_number` of a subdomain edge therefore reads `fd_coeff` off the end of the array, and whatever is adjacent in memory becomes a force.

Nothing reports this. The read is in bounds for a body sitting in the interior of one rank's subdomain, which is why single-rank runs and the golden files never show it.

## Reproducer

`examples/2D_ibm_force_decomposition` — a static cylinder at Re 40, 200×200, 200 steps, `fd_order = 4`, body at the domain center so an even rank count in a direction puts a subdomain edge through it. Drag from `restart_data/ib_state_200.dat`:

| ranks | Fx |
|---|---|
| 1 | 1.02352019 |
| 4 | 1.02933438 |

0.57 % apart, on a uniform grid, same case, same binary. A force that changes when you re-cut the domain.

The size of the discrepancy is set by whatever happens to be adjacent in memory, so it is not bounded by anything. On a 3D sphere at Re 100 on 64 GPU ranks, with the body at the center of the domain and so at the corner of eight subdomains, it appears as a transverse force of 1.08 × the drag on a body that has no transverse force at all: Fx +0.4040, Fy +0.4364, Fz +0.4364. That number is pinned at 0.4364 from the first time step to the fifty-thousandth while Fx relaxes from 1.638 to 0.404 — it is not a flow quantity. Reproducing MFC's own volume sum offline from the restart field gives Fx +0.40459 (0.15 % from MFC's 0.40399) and Fy +0.00002, confirming the local sum itself is right and the spurious part enters through the out-of-bounds read.

## Fix

`s_compute_finite_difference_coefficients` already accepts an `offset_s`. Passing one and widening the allocation to `-fd_number:m + fd_number` covers the range that is actually read. `x_cc` carries at least `buff_size = 10` ghost cells whenever `ib` is set (`s_configure_coordinate_bounds`), so the extra coefficients are built from real coordinates rather than clamped.

With that change the reproducer gives 1.02352019 at 1 rank and 1.02351629 at 4 — 0.0004 % apart, and the 1-rank answer is bit-identical to before, as it must be.

## Scope

Affects `ib` + `viscous` on more than one rank whenever a body straddles a subdomain edge. Forces are diagnostic for a static body, so nothing else changes there; for `moving_ibm` the force drives the motion, so trajectories are affected too.

## Separate, smaller observation

The force integral's outer stencil is the centered one, whose coefficients on a non-uniform grid are the uniform-grid weights over a non-uniform denominator. It is first-order there and, more to the point for a force, non-telescoping, so the volume sum of −∇p does not reduce exactly to the surface integral it stands for. Measured on the stretched sphere grid this moves the pressure drag about 2 % (0.20975 vs 0.20541 for a conservative flux-form stencil) and contributes no transverse force. Worth a look, but it is an accuracy question and not the defect above; filing separately if it is wanted.

sbryngelson:
Fixes #1860.

## The bug

`s_compute_ib_forces` evaluates the viscous stress tensor at every cell its own stencil reaches:

```fortran
do l = -fd_number, fd_number
    call s_compute_viscous_stress_tensor(viscous_stress, q_prim_vf, dynamic_viscosity, i + l, j, k)
```

`s_compute_viscous_stress_tensor` then indexes the coefficient array at the cell it was handed (`m_viscous.fpp:1141`):

```fortran
velocity_gradient_tensor(l, 1) = velocity_gradient_tensor(l, 1) + fd_coeff_x(r, i)*q_prim_vf(...)%sf(i + r, j, k)
```

so `fd_coeff_x` is read at `i - fd_number` through `i + fd_number`, while it is allocated `0:m`. A body cell within `fd_number` of a subdomain edge reads off the end of the array and whatever is adjacent in memory becomes a force.

Nothing reports it. The read stays in bounds for a body in the interior of one rank's subdomain, which is why single-rank runs and the goldens never showed it.

## The fix

`s_compute_finite_difference_coefficients` already takes an `offset_s`. Pass one, widen the allocation to match. `x_cc` carries at least `buff_size = 10` ghost cells whenever `ib` is set (`s_configure_coordinate_bounds`), so the extra coefficients come from real coordinates rather than a clamp.

Three lines of allocation, three call sites, one local. Nothing else reads outside `0:m`, so no other consumer changes.

## How the validation is obtained

`examples/2D_ibm_force_decomposition/` — static cylinder, Re 40, Ma 0.1, uniform 200 x 200, `fd_order = 4`, 200 steps. The body is at the domain center, so an even rank count in a direction puts a subdomain edge through it. The force on a rigid body is a property of the flow, so the same case on a different rank count must give the same answer.

Read `restart_data/ib_state_200.dat` (20 doubles, `[time, Fx, Fy, Fz, Tx, Ty, Tz, ...]`):

| ranks | Fx before | Fx after |
| --- | --- | --- |
| 1 | 1.02352019 | 1.02352019 |
| 4 | 1.02933438 | 1.02351629 |

0.57 % of drag out of adjacent memory, against 0.0004 % after. The single-rank number is bit-identical before and after, as it must be — one rank never reaches outside its own interior. Lift stays at ~8e-6 in all four configurations (the cylinder is symmetric about y = 0), so the drag agreement is not bought with a symmetry error.

The README in that directory carries the same table and the command to reproduce it.

## Scope

`ib` + `viscous` on more than one rank with a body straddling a subdomain edge. Forces are diagnostic for a static body; for `moving_ibm` the force drives the motion, so those trajectories change too. Existing IB goldens are single-rank and are unaffected — no regeneration needed.

The discrepancy is not bounded by anything, since it is set by adjacent memory. On a 3D sphere at Re 100 on 64 GPU ranks, body at the center of the domain and so at the corner of eight subdomains, the same read shows up as a transverse force of 1.08 x the drag on a body that has none (Fx +0.4040, Fy +0.4364, Fz +0.4364), pinned at that value from the first time step to the fifty-thousandth while Fx relaxes from 1.638 to 0.404.

### DIFF
--- a/examples/2D_ibm_force_decomposition/README.md
+++ b/examples/2D_ibm_force_decomposition/README.md
@@ -1,0 +1,43 @@
+# Immersed-boundary force under a changing MPI decomposition
+
+A static cylinder in uniform flow at Re 40, Ma 0.1, on a uniform 200 x 200 grid. The body sits at the
+origin, which is also the center of the domain, so any decomposition with an even number of ranks in a
+direction puts a subdomain edge straight through it.
+
+The force on a rigid body is a property of the flow. Running the same case on a different number of ranks
+must not change it. This case measures whether it does.
+
+## Running it
+
+```
+./mfc.sh run examples/2D_ibm_force_decomposition/case.py -n 1
+./mfc.sh run examples/2D_ibm_force_decomposition/case.py -n 4
+```
+
+Each run writes `restart_data/ib_state_200.dat`: 20 doubles, `[time, Fx, Fy, Fz, Tx, Ty, Tz, ...]`.
+
+```
+python3 -c "import numpy as np; a = np.fromfile('restart_data/ib_state_200.dat'); print(a[1], a[2])"
+```
+
+## What it shows
+
+`fd_order = 4` here, so the force integral's stencil reaches two cells, and a body cell two cells from a
+subdomain edge asks for finite-difference coefficients outside the interior. Before the coefficients were
+defined over that range they were read off the end of the array:
+
+| ranks | Fx before | Fx after |
+| --- | --- | --- |
+| 1 | 1.02352019 | 1.02352019 |
+| 4 | 1.02933438 | 1.02351629 |
+
+0.57 percent of drag appearing out of adjacent memory, against 0.0004 percent after. The single-rank
+answer is unchanged, because a single rank never reaches outside its own interior.
+
+The size of the discrepancy is set by whatever is adjacent in memory and is not bounded by anything: on a
+3D sphere at Re 100 on 64 ranks the same read produced a transverse force of 1.08 times the drag on a body
+that has none.
+
+Lift is a second, independent check: the cylinder is symmetric about y = 0, so Fy must be zero. It is
+about 8e-6 here in every configuration, which is the level at which the staircase representation of a
+circle on this grid is symmetric, and it does not move.
--- a/examples/2D_ibm_force_decomposition/case.py
+++ b/examples/2D_ibm_force_decomposition/case.py
@@ -1,0 +1,93 @@
+#!/usr/bin/env python3
+"""The smallest case that shows MFC's immersed-boundary force depending on the MPI decomposition.
+
+A static cylinder in uniform flow, Re 40, on a grid coarse enough to run on a laptop. The body sits at the
+origin, which is also the domain center, so a decomposition with an even number of ranks in a direction puts
+a subdomain edge straight through the body. Run it at one rank and at four and compare `restart_data/ib_state_200.dat`:
+the cylinder is symmetric about y = 0 and the lift must be zero, and the drag must not care how the domain
+was cut up.
+"""
+
+import json
+import os
+
+Re, Ma, d, gamma, U, rho = 40.0, 0.1, 1.0, 1.4, 1.0, 1.0
+P = rho * U**2 / (gamma * Ma**2)
+L = float(os.environ.get("L", 5.0))
+dx = float(os.environ.get("DX", 0.1))
+N = int(2 * L / dx)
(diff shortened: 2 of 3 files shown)


sbryngelson:
Duplicate of #1859, which was opened first and fixes the same out-of-bounds read by clamping to the nearest interior cell rather than widening the allocation. Closing in its favour.

The example case and the measured before/after table have been moved to that branch, and the clamp-versus-widen tradeoff is written up in [my comment there](https://github.com/MFlowCode/MFC/pull/1859#issuecomment-5643479001) — the diff here stays available if the wider allocation is preferred, since the two differ at first order on a stretched grid where the clamp substitutes a neighbouring cell's stencil.

PR #1877, opened 2026-09-13
Require the relaxation targets that grcbc_out and grcbc_vel_out read

sbryngelson:
## Contribution Policy

- [x] I confirm this PR meets the above expectations and reflects my own understanding and real-world context.

Made with Claude Code.

---

## What

`check_grcbc` now requires the targets the CBC kernel actually reads:

* `bc_<dir>%pres_out` when `bc_<dir>%grcbc_out` is set
* `bc_<dir>%vel_out(1..num_dims)` when `bc_<dir>%grcbc_vel_out` is set

Neither was required before. Setting `grcbc_out` without `pres_out` validated cleanly.

## Why

`m_cbc.fpp` relaxes a GRCBC outflow toward a prescribed pressure:

```fortran
L(eqn_idx%adv%end) = c*(1._wp - Ma)*(pres - pres_out(${CBC_DIR}$))/Del_out(${CBC_DIR}$)
```

With `pres_out` never given, that relaxes the boundary toward an undefined value.

**The failure is silent and lands nowhere near its cause.** I hit it on a 2D case and lost hours to it. On a uniform quiescent field -- no gust, no immersed boundary, nothing to disturb it -- density in the last cell of the outflow direction walked away linearly at 0.032 per step from step 1, crossed zero near step 31, and the run aborted on ICFL at step 57:

| step | max abs(rho - 1) | cell |
| --- | --- | --- |
| 1 | 0.034 | (1912, ·) |
| 8 | 0.258 | (1912, ·) |
| 16 | 0.482 | (1912, ·) |

`m = 1912`, so every deviation sat on the outflow boundary. Nothing in the output named the BC.

It did not read as a boundary problem either. `run_time.inf` shows ICFL at **exactly 0.4000 for 56 steps** with no drift, then **1.3e15** in one step -- so it looks like a single catastrophic event rather than a boundary slowly poisoning the solution, and the obvious suspects (grid, initial condition, immersed boundary, resolution) all get eliminated first. It was identified only by dumping the field every step and asking where it first departed from uniform.

The same configuration is fine the moment `pres_out` is supplied, which is what `examples/2D_shockdroplet` and friends already do.

## The `vel_out` half

`grcbc_vel_out` reads `vel_out(${CBC_DIR}$, dir_idx(1))` -- the *normal* component, selected through `dir_idx`, which is `(1,2,3)` / `(2,1,3)` / `(3,1,2)` for x / y / z. So `vel_out(1)` is not the normal component for a y or z boundary. Rather than guess which component to require, this requires all `num_dims` of them, following `check_synthetic_turbulence`. That is the same direction-blindness that #1854 corrects on the inflow side; the outflow side was never covered.

## Verification

* **Rejects nothing in tree.** All 181 example cases generate and validate; 9 of them use `grcbc_out` or `grcbc_vel_out`. `./mfc.sh format` clean, and the full precheck passes including "All example cases are valid".
* **Full validator suite passes** -- 72 tests.
* **Three regression tests added** (`TestGrcbcOutflowTargets`): the rejection, the acceptance once targets are given, and that the third velocity component is required only in 3D.

No Fortran changes, no GPU code, no effect on any case that was already valid.


sbryngelson:
Thanks - taking the second point, which is correct and was the real bug in this PR.

**vel_out components (accepted).** `m_cbc.fpp` adds only the normal velocity:

```fortran
! Add GRCBC for Subsonic Outflow (Normal Velocity)
L(adv%end) = L(adv%end) + rho*c**2*(1 - Ma)*(vel(dir_idx(1)) + vel_out(CBC_DIR, dir_idx(1))*sign(1, cbc_loc))/Del_out(CBC_DIR)
```

and `dir_idx = (/1,2,3/) | (/2,1,3/) | (/3,1,2/)` for x|y|z (m_cbc.fpp:942-948), so `dir_idx(1)` is just 1, 2, 3. The transverse components are read by the *inflow* branch a few lines up (`dir_idx(2)`, `dir_idx(3)`), which is why `grcbc_in` needs them all and this one does not. Requiring them here would have rejected a y-boundary case that sets only `vel_out(2)` and runs correctly. Now requires the normal component only, and the message names it. My "matching check_synthetic_turbulence" rationale was precedent standing in for reading the kernel; the kernel is explicit.

Added a regression test for the direction-awareness (`bc_y` must demand `vel_out(2)`, not `vel_out(1)`) and one asserting a transverse component is *not* required - both fail on the previous commit.

**Message wording (accepted).** It now names the specific component instead of the phrase "for all num_dims".

**L(adv%end) comment (checked, no change needed to the logic).** Both sides really do write `L(adv%end)`: one branch handles them,

```fortran
else if ((cbc_loc == -1 .and. bc%beg == BC_CHAR_NR_SUB_OUTFLOW) .or. (cbc_loc == 1 .and. bc%end == BC_CHAR_NR_SUB_OUTFLOW)) then
```

with `sign(1, cbc_loc)` selecting the side rather than a different characteristic index. I reworded the comment to state that explicitly so the next reader does not have to re-derive it.

PR #1887, opened 2026-09-17
Refresh the device copy of num_gps when a moving immersed boundary is updated

### LINKED ISSUE #1886: Moving IBs on GPU: device copy of num_gps is never refreshed after s_update_mib, so ghost-point kernels use the setup-time count (nondeterministic, wrong near-surface states)
## Summary

For a moving immersed boundary on GPU, the device copy of `num_gps` (the number of ghost points) is refreshed once, in `s_ibm_setup`, and never again. `s_update_mib` recounts it on the host every RK stage, but every device loop over the ghost points is bounded by `num_gps` read inside the target region, i.e. the stale device copy. The kernels therefore process the setup-time number of ghost points for the whole run:

- when the true count has dropped, entries beyond it are stale ghost points from earlier stages -- cells that may now be fluid at the surface -- and `s_ibm_correct_state` writes reconstructed wall states into them;
- when the count has grown, the newest ghost cells are never corrected.

Which stale entries survive depends on the order in which `s_find_ghost_points` fills the list (atomic capture), so the result differs from run to run.

## Where (master @ 5c158b3)

`src/simulation/m_ibm.fpp`:

- L38-42: `integer :: num_gps` with `$:GPU_DECLARE(create='[num_gps]')` -- a persistent device copy.
- L118, L127 (`s_ibm_setup`): `call s_find_num_ghost_points(num_gps)` ... `$:GPU_UPDATE(device='[num_gps]')` -- the only refresh.
- L963 (`s_update_mib`): `call s_find_num_ghost_points(num_gps)` -- host recount, no `GPU_UPDATE`.
- L228 (`s_ibm_correct_state`), L482 (`s_compute_image_points`), L713 (`s_compute_interpolation_coeffs`): `do i = 1, num_gps` / `do q = 1, num_gps` inside `GPU_PARALLEL_LOOP` (`default(present)`), reading the device copy. (`s_apply_levelset` receives `num_gps` as an argument and is fine.)

## How it shows

Frontier, CCE 20 / OpenMP offload, `--gpu mp`, MFC branch `sbryngelson/MFC:nsf-bird` (master plus moving-STL fixes; the lines above are unchanged from master).

Two flapping STL wings (kin_model 1, prescribed roll+pitch), 8 M cells, one node. Two identical runs restarted from the same checkpoint mid-beat differ after 5 steps in ~15,000 cells with O(1) differences in momentum and energy, ~120,000 cells after 20 steps. A restart during the glide (wing at rest) is bit-exact after 210 steps, so the nondeterminism is specific to a body in motion. Symptoms in production: near-surface density collapses (rho down to ~0.1 in a few cells at the trailing edge) at the phase of peak pitch rate, left/right lift asymmetry of up to 25 % of peak on a mirror-symmetric two-wing case, and ~30 % of first-beat lift.

The count really does change for a rotating thin body: every cell of a body thinner than ~6 cells is a ghost point, so `num_gps` is the whole voxelised body, which changes by tens of percent as a plate pitches through the grid. For the translating-only cases in the test suite it barely changes, which is presumably why this has not shown.

## Fix

Refresh the device copy after the recount in `s_update_mib`:

```fortran
        call s_find_num_ghost_points(num_gps)
        $:GPU_UPDATE(device='[num_gps]')
```

With this single line, the two identical restarts above are bit-identical at every cell (0 differing values in rho, momentum, E at +5 and +35 steps), and a mid-beat restart matches the unrestarted run to 1e-13 at +105 steps. Cost is one 4-byte copy per stage. A cleaner variant would leave the counter on the device (have the counting kernel write the device variable and read it back only for the host-side allocation check).

Notes: the mechanism (loop bound taken from the device copy) is inferred from the code and from the fix making the runs bit-exact; a direct host-vs-device probe is running and I will add its output. The OpenACC/NVIDIA path may evaluate the trip count on the host and hide this -- untested here.

Commit with the fix on the fork: sbryngelson/MFC@207d3246 (bundled with two unrelated changes; the `num_gps` line is the one that matters). Found and reproduced with Claude Code.

sbryngelson:
## Description

`num_gps` is a declare-target module variable that bounds every device loop over the ghost points. It is copied to the device once, in `s_ibm_setup`; `s_update_mib` recounts it on the host every RK stage for a moving body but never refreshed the device copy, so the kernels kept the setup-time count for the whole run. Entries beyond the current count were stale ghost points from earlier stages, written as wall states into cells that are now fluid; when the count had grown, the newest ghost cells were never corrected. Which stale entries survived depended on the atomic fill order, so results differed run to run. Fixes #1886 (which has the probe data and the reproducer).

One line: `$:GPU_UPDATE(device='[num_gps]')` after the recount in `s_update_mib`, with a comment.

Static bodies never enter `s_update_mib` and are unaffected; the change is one 4-byte host-to-device copy per RK stage.

## Type of change

## Scope

## How Has This Been Tested?

Frontier, CCE 20, `--gpu mp`, two flapping STL wings with prescribed kinematics (8 M cells, one node):

- Two identical restarts from the same checkpoint: 15,000 cells differed after 5 steps before; **bit-identical** (0 differing values in rho, momentum, E at +5 and +35 steps) after.
- Mid-run restart vs the unrestarted run: O(1) differences before; 1e-13 at +105 steps after.
- A probe kernel reading `num_gps` on the device disagrees with the host value in 11 of 19 samples once the body moves; the count on one rank swings 68..1800 within a wingbeat against a setup value of 510.
- First-wingbeat mean lift changes by 30 %; a mirror-symmetric pair of wings goes from 30 % left/right asymmetry to 0.3 %.

Not run: the OpenACC/NVIDIA path. Existing GPU tests use translating bodies whose ghost-point count barely changes, which is presumably why this was invisible to CI; a determinism test (run twice, compare) would catch it.

## Checklist

- [x] I added Doxygen docstrings to the new code (n/a: no new routines)
- [x] I have made corresponding changes to the documentation (n/a)
- [x] I have added regression tests to the test suite so that people can verify in the future that the feature is behaving as expected (n/a: the effect is nondeterminism on a moving body; the reproducer is in #1886)
- [x] I have added example cases in `examples/` that demonstrate my new feature performing as expected (n/a)

This PR was made with Claude Code.


sbryngelson:
@danieljvickers you'll want to be aware of this fix

PR #1890, opened 2026-09-17
Shard the Frontier benchmark across two SLURM jobs so it fits the 2 h QOS cap

sbryngelson:
## Description

The Frontier bench lane has failed on every PR since 2026-09-16 with

```
error: *** JOB 5499026 ON frontier10198 CANCELLED AT 2026-09-17T14:28:53 DUE TO TIME LIMIT ***
```

It is `continue-on-error`, so nothing has been blocked — but no PR has had a Frontier speedup table since.

### What actually changed (this is not a workload-growth problem)

`bench.yaml` has had 7 cases since July and the 01:59:00 walltime since June. The lane **passed in 55–66 min through 09-15** and has failed at 126+ min on **every** run since 09-16. Comparing a passing and a failing run's per-target `exec` times from the archived YAMLs:

| case | sim, 09-15 | sim, 09-17 |
|---|---|---|
| ibm | 129.1 s | 128.9 s |
| 5eq_rk3_weno3_hllc | 113.7 s | 115.1 s |
| viscous_weno5_sgb_acoustic | 122.3 s | 121.2 s |

Identical. Every case still finishes in ~3 min inside `mfc.sh run`. The entire difference is the gap between one case's `End-time` and the next case's start:

| | 09-15 | 09-17 |
|---|---|---|
| inside each case | ~3 min | ~3 min |
| **between cases** | **4 s** | **14 min 15 s** — the same to the second, every case |

A fixed stall in process exit, silent in every log. No change in the window to the toolchain run path, the loaded modules, or the nodes (`frontier10198/9` in both runs). Darshan — which Frontier preloads into every MPI job and flushes at `MPI_Finalize` — asserted in exactly that phase the same day (#1866, `heatmap_mpi_redux: Assertion 'heatmap_runtime' failed` in `syscheck`). That is the prime suspect but is **unproven**.

### What this PR does

1. **Diagnose.** `bench.py` now prints, per case, launch-to-return wall time and how much of it fell *after* the run's own `End-time`. On `frontier`/`frontier_amd`, shard 1 runs with `DARSHAN_DISABLE=1` and shard 2 with Darshan on, in the same allocation. **This PR's own Frontier bench lane is the A/B.** If shard 1's post-exit time drops to seconds, the fix is one environment variable and the sharding below is just headroom.

2. **Shard** each tree's bench across two concurrent SLURM jobs, using the `i/N` round-robin the case-optimization lanes already use: `mfc.sh bench --shard i/N`, `bench.sh` passes `$job_shard` through, `run_parallel_benchmarks.sh` submits `1/2` + `2/2` per tree on Frontier and folds them back with a new `mfc.sh bench_merge` (strips `--shard`/`-o` from the recorded invocation so both trees' metadata compare equal; refuses a duplicated case or a lock mismatch). Every other cluster gets an empty shard list and is unchanged. Whether sharding stays depends on what the A/B says — I'll update this PR with the result.

## Type of change

## Scope

## How Has This Been Tested?

- `bench_merge` + `bench_diff` on synthetic shard files with the real `output_summary`/`exec`/`grind` schema: normal speedup table, all cases present, no metadata warning. `bench_merge` rejects a duplicated case and a lock mismatch.
- Shard partition checked against `bench.yaml` (every case in exactly one shard).
- The timing report checked against the real failing-run numbers: reports 856 s post-exit for `ibm`.
- `./mfc.sh precheck`: passes.
- The Frontier A/B runs in this PR's CI.

## Checklist

- [x] New and existing tests pass locally — n/a for GPU, CI/toolchain-only change

Diagnosed and written with Claude Code.


sbryngelson:
Closing in favour of #1894, which takes a different approach.

Two things were wrong here.

**It treated the symptom.** Sharding the case list makes the work fit the walltime, but the reason it stopped fitting is a 14 min 15 s stall between cases that appeared on 2026-09-16 — the same to the second on every case, with solver times unchanged. That is a defect, not a capacity limit, and halving the cases per job hides it.

**It did not work.** This PR's own Frontier lane proved it. Both shards of a tree build into the same staging directory — same case, same config, so the same build slug — and raced on the generated sources:

```
error: Failed to open file '.../fypp/pre_process/m_perturbation.fpp.f90' for write [FyppFatalError]
error: [Errno 2] No such file or directory: ... [FileNotFoundError]
```

All 60 staging references in that job were to one directory, `gpu-mp-0e981924c0`, 71 of them under `pr/`. The original two-job design was safe because `pr/` and `master/` are separate trees; adding a second concurrent job *within* a tree introduced the collision. The build died before any benchmark ran, which is also why the Darshan A/B this PR was carrying never produced a result.

That is the same defect I flagged as finding 7 in the #1821 review — concurrent runs sharing one staging directory — and then wrote into a fix for an unrelated problem.

#1894 keeps only the diagnostic half, 59 lines instead of ~150: per-case wall-time accounting that shows how much of each case is spent after the solver has finished, plus disabling Darshan on the Frontier benches, which is the leading suspect for the stall. No extra jobs, so no race.

Sharding may still be worth having as headroom later, but only once the stall is understood and only with per-shard build isolation.

PR #1894, opened 2026-09-18
Measure where the Frontier benchmark's time actually goes

sbryngelson:
Replaces #1890, which took the wrong approach. See below.

## What is wrong

The Frontier bench has failed on every PR since 2026-09-16, killed by the 2 h limit partway through its seven cases:

```
error: *** JOB 5499026 ON frontier10198 CANCELLED AT 2026-09-17T14:28:53 DUE TO TIME LIMIT ***
```

It is `continue-on-error`, so nothing has been blocked — but no PR has had a Frontier speedup table since.

**This is not workload growth.** `bench.yaml` has had 7 cases since July and the walltime has been 01:59:00 since June. Comparing a run that passed on 09-15 against one that failed on 09-17, from the archived result YAMLs:

| case | sim, 09-15 | sim, 09-17 |
|---|---|---|
| ibm | 129.1 s | 128.9 s |
| 5eq_rk3_weno3_hllc | 113.7 s | 115.1 s |
| viscous_weno5_sgb_acoustic | 122.3 s | 121.2 s |

Identical, and each case still finishes in ~3 minutes. The entire difference is between cases:

| | 09-15 | 09-17 |
|---|---|---|
| inside each case | ~3 min | ~3 min |
| **between cases** | **4 s** | **14 min 15 s** — the same to the second, every case |

A fixed stall in process exit, silent in every log. Nothing else changed in the window: not the toolchain run path, the modules, or the nodes (`frontier10198/9` in both runs).

## What this does

**Turns Darshan off for the Frontier benches.** Frontier preloads it into every MPI job and it flushes its log in `MPI_Finalize` — and on the same day, its heatmap module asserted in exactly that phase and killed a `syscheck` outright (`heatmap_mpi_redux: Assertion 'heatmap_runtime' failed`, seen on #1866). A benchmark has no use for I/O profiling.

**Makes the stall measurable.** `bench.py` now prints, per case, the wall time from launch to return and how much of it fell *after* the run printed its own `End-time`:

```
> Wall:    1035s launch to return, 856s of it after the run printed End-time
```

That interval is the one the logs never showed. It separates a slow solver from a process that has finished and is stuck exiting, so this regression becomes measurable rather than inferred — and the next one is legible whatever its cause.

If the bench now completes in ~60 min, Darshan was it. If it still stalls, the new timing says where the time went instead, and that is worth having either way.

## Why #1890 is closed

That PR sharded the case list across concurrent SLURM jobs so the work would fit the walltime. Two problems:

1. **It treats the symptom.** A 14-minute per-case stall is a defect, not a capacity limit. Halving the cases per job hides it.
2. **It did not work.** Both shards of a tree build into the same staging directory — same case, same config, same build slug — so they raced on the generated sources and the build died before any benchmark ran:
   ```
   error: Failed to open file '.../fypp/pre_process/m_perturbation.fpp.f90' for write [FyppFatalError]
   error: [Errno 2] No such file or directory: ... [FileNotFoundError]
   ```
   The original two-job design was safe because `pr/` and `master/` are different trees; adding a second concurrent job *within* a tree introduced the race. (Which is the same defect flagged as finding 7 in the #1821 review, then written into the fix for a different problem.)

Sharding may still be worth having as headroom, but only after the stall is understood, and only with per-shard build isolation.

## Type of change

## Scope

## How Has This Been Tested?

- The wall-time report checked against the real failing-run numbers: given a log whose `End-time` is 856 s before the return, it prints exactly that; with no `End-time` in the log it falls back to the launch-to-return figure rather than raising.
- `./mfc.sh precheck` (full lint gate incl. toolchain test suite): passes.
- The Darshan hypothesis is tested by this PR's own Frontier bench lane. It is a hypothesis until that runs — the evidence above is consistent with it and rules out the alternatives I could check, but it is not yet proof.

## Checklist

- [x] New and existing tests pass locally — n/a for GPU, CI/toolchain-only change

Diagnosed with Claude Code.


sbryngelson:
## Result: the hypothesis this PR started from is refuted

The Frontier lane ran, and disabling Darshan changed nothing — the job still died on the time limit at case 6 of 7. The timing shows why it never could have worked:

```
> Wall:    1037s launch to return, 0s of it after the run printed End-time
> Wall:    1041s launch to return, 1s of it after the run printed End-time
> Wall:    1047s launch to return, 1s of it after the run printed End-time
```

Darshan flushes in `MPI_Finalize`, which is exit. **Exit is 0 s.** The missing time is not there.

It is *before* the run starts. Cross-checking against the 09-17 logs: `ibm.out` brackets itself `Start-time 14:11:37` → `End-time 14:14:43`, 186 s, while the bench log showed 1044 s between case starts. I attributed that 858 s to process exit — on the strength of a number that came from a synthetic test of my own reporting code, not from Frontier. Measured, it falls before `Start-time`.

I've removed the Darshan change accordingly.

## What this PR is now

Only the instrumentation, which is the part that produced the correction. It splits each case three ways instead of two:

```
> Wall:    1037s = 851s startup + 186s run + 0s exit
```

That is the whole change: ~28 lines in `bench.py`, nothing in the workflows.

## What it points at next

`run()` in `toolchain/mfc/run/run.py` calls `build(targets)` on every invocation, and `bench.py` invokes `mfc.sh run` once per case — so seven build checks per benchmark, over Lustre. That is the first thing to rule in or out, along with venv and toolchain imports on that filesystem, and `srun` step launch latency.

I have not fixed the stall, and I would rather land the measurement than guess a third time. With this in, the next Frontier bench run says which of those three it is.

## Why the summary above differs from the original description

The original framing — that this was a teardown stall and Darshan was the likely cause — was wrong in both halves. The evidence for "14 minutes after the solver finished" was an inference from timestamps, not a measurement, and the measurement disagrees. The one thing it got right is that the solver is not slower: ~3 min then, ~3 min now.

## How Has This Been Tested?

- The three-way split reproduces the real shape: given a log with the 09-17 `Start-time`/`End-time` pair and a 1037 s launch-to-return, it reports `851s startup + 186s run + 0s exit`. With neither stamp present it falls back to the total rather than raising.
- `./mfc.sh precheck`: passes.

sbryngelson:
**Root cause found — it's #1904, and it isn't what this PR guessed.**

The startup time this PR instruments is a **spurious whole-image device LTO relink**. CMake ≥3.27 has the linker report its dependencies; amdflang's `clang-linker-wrapper` stages the device image through a randomly-named `/tmp/*.image.wrapper-*.o`, records it as a prerequisite, then deletes it. `simulation` is therefore permanently out of date and every build redoes a ~19-minute link with every object already current. `bench.py` calls `mfc.sh run` per case, so a 7-case job paid it seven times — about two hours, which is what crossed the 1:59 walltime.

Fixed in #1904: no-op build goes 19m37s → 0.4s.

**What this PR should be now.** The instrumentation is what produced the correction, and it earns its place — it's what showed the time was in startup rather than the exit I had assumed, which is what sent the search in the right direction. But it is diagnostic scaffolding for a bug that now has a fix, so I'd rather it be judged on that basis than merged on the strength of a story that turned out wrong.

Options, and I don't have a strong preference:

1. **Keep it.** ~28 lines in `bench.py`, no workflow changes. The next time a case's wall time and its solver time diverge, the split says which phase without a day of inference.
2. **Close it.** #1904 removes the reason anyone would look, and the instrumentation can come back if something like this recurs.

For the record, what this PR originally claimed and how it was wrong:

- *"~14 minutes after the solver finished"* — measured, exit is **0–1 s**. The number I had came from a synthetic test of my own reporting code, not from Frontier.
- *"Darshan flushing at MPI_Finalize"* — refuted by that measurement; disabling it changed nothing, and the change was removed.
- The earlier #1890 sharded the case list to fit the walltime, which both treated the symptom and raced two builds into one staging directory.

Three wrong hypotheses, and the thing that settled it was measuring rather than reasoning about what *should* be slow.

Credit where it's due: the Frontier-side diagnosis — catching `ld.lld` in the act, `make --debug=b` naming the vanished `/tmp` prerequisite, the module-file mtimes, and the four-build verification — was all done by @sbryngelson on a login shell. I had the phase and none of the mechanism.

PR #1904, opened 2026-09-18
Stop amdflang GPU builds relinking the whole device image every time

sbryngelson:
A no-op build of `simulation` on Frontier took **19m37s**. It now takes **0.4s**.

## The bug

CMake ≥ 3.27 has the linker report its own dependencies, passing `--dependency-file=link.d`. amdflang's `clang-linker-wrapper` stages the device image through `/tmp/simulation.openmp.image.wrapper-<random>.o` and that path lands in `compiler_depend.make` as a prerequisite of the target — but the wrapper deletes the file on exit and randomises the name on the next link.

`make --debug=b` says it plainly:

```
File '/tmp/simulation.openmp.image.wrapper-b3b0a6.o' does not exist.
Must remake target '/tmp/simulation.openmp.image.wrapper-b3b0a6.o'.
Prerequisite '/tmp/simulation.openmp.image.wrapper-b3b0a6.o' of target 'simulation' does not exist.
Must remake target 'simulation'.
```

So `simulation` is **permanently out of date**, and every build redoes the whole-image device LTO link with every object already current — `ld.lld` single-threaded at 99.6% CPU, 6.3 GB RSS, ~19 minutes. Confirmed by two full runs on a warm tree at 20m11.9s and 19m37.0s with zero compiler invocations.

## Why this form of the fix

CMake already disables this feature for the GNU binutils linkers, for the same stated reason — *"they generate spurious dependencies on temporary files when LTO is enabled"*, GNU bug 30568. Its exclusion list names only `ld`, `ld.bfd` and `ld.gold`, so `ld.lld` reached through the offload linker-wrapper isn't covered. This is CMake's own remedy applied where its detection doesn't reach, not a workaround.

It has to be **the variable, at directory scope**. Two other forms were tried and both fail *silently*:

- `set_target_properties(... LINK_DEPENDS_USE_LINKER OFF)` — there is no such property; CMake creates a meaningless custom one without warning, and `--dependency-file` stays. Verified in an isolated two-target project, for C as well as Fortran.
- `set(...)` inside the amdflang branch of `cmake/MFCTargets.cmake` — that block is inside `function(MFC_SETUP_TARGET)`, so it would be function-scoped and evaporate.

Gated on `LLVMFlang` because `clang-linker-wrapper` is the mechanism, so that's the analogue of CMake's linker-based exclusion; the other three compilers keep their dependency tracking unchanged.

## Verification

Measured on Frontier, amdflang `--gpu mp`:

| build | | wall | binary |
|---|---|---|---|
| 1 | after reconfigure | 24m40s | one legitimate relink |
| 2 | no changes | **0.395s** | identical |
| 3 | `touch src/simulation/m_weno.fpp` | 25m16s | fypp + compile + link all re-ran |
| 4 | no changes | **0.581s** | identical |

Build 3 is the control that matters: `LINK_DEPENDS_USE_LINKER OFF` is the pre-3.27 behaviour, so the theoretical risk is *under*-rebuilding. A source touch still drives fypp, the object compile and the relink. The fix removes the spurious rebuild without removing the real one.

The binary is byte-identical across all four including after the touch — `touch` didn't change file *content*, so fypp emitted identical Fortran and the chain reproduced exactly. Worth recording separately: **the amdflang offload build is bit-reproducible.**

Builds 1 and 3 ran while login10's load went 2.07 → ~16, so don't read their wall times as the link cost; they aren't comparable to the quiet-machine 19–20 min figures. It doesn't affect the verdict — 0.4 s versus 25 min isn't a subtle margin.

## Impact

`bench.py` calls `mfc.sh run` once per case and `run()` calls `build()`, so a 7-case benchmark paid **seven** whole-image device links — roughly two hours — which is what pushed it past the 1:59:00 walltime. Every Frontier bench on every branch has failed on that limit since 09-16.

The test suite dedupes by build slug (`test.py:361-365`) and passes `--no-build`, so it paid once per configuration rather than once per case — still ~20 minutes each, and case-optimization runs generate many distinct slugs.

## Not fixed here

Two separate things surfaced alongside this and are worth their own issues:

1. **The device link got ~2× more expensive around 2026-09-15 21:40**, when every `cpe/*.lua` and `amd/*.lua` modulefile on Frontier was rewritten with version strings unchanged — including the `cpe/25.09` and `amd/7.2.0` that `toolchain/modules` pins by name. Bench mean wall time stepped 31.6 → 37.5 → 66.8 min across that boundary. Correlation, established by mtimes; not proven causal.
2. **`-attributor-max-pi-accesses=16384` is ~40% of the link cost** — 11m47s without it versus ~19–20 min with. It can't simply be reverted: #1759 added it because without it the Attributor goes pessimistic and untouched kernels regenerate with 2.4–4.5× worse ISA. Tuning it below 16384 is a second-order win that needs an ISA check at each step. With this PR the cap is paid once per job instead of seven times, which makes that tuning optional rather than urgent.

## Type of change

## How Has This Been Tested?

The four builds above, on Frontier. `./mfc.sh precheck` passes. The change is inert for GNU, NVHPC and Cray.

## Checklist

- [x] New and existing tests pass locally with my changes — n/a, build-system change; verified by build/no-op/touch/no-op on the affected toolchain


sbryngelson:
Related: #1905 records the other half of the Frontier slowdown — the device link itself got ~2x more expensive across the 2026-09-15 PE reinstall, independent of the always-relink bug fixed here.

The two compounded: this bug made you pay the link 7 times per benchmark, the PE change made each one twice as expensive. Fixing this alone brings a bench job back under the walltime; #1905 is worth reporting to OLCF regardless, and carries the open question of whether the PE refresh also changed generated kernel quality.

sbryngelson:
Measured follow-up: this PR is the whole problem, and the margin arithmetic makes that concrete.

```
7 bench cases x 17.1 min per device link  = 119.5 min
SLURM time limit                          = 119.0 min
```

The benchmark was sitting *exactly* on its walltime. It didn't fail on 09-16 because something broke — it had never had more than a minute of margin, and ordinary code growth tipped it. A job pinned to its limit fails like a cliff even when the cost underneath is drifting up gently, which is what made this look environmental.

With this PR it's one link instead of seven: **~154 min → ~22 min** against that same 119 min limit.

Two things ruled out along the way, both with controlled measurements (see #1905, now closed):

- **No environmental regression.** Build-only jobs, which do the same work every run, read 35.8 / 35.5 / 35.1 / 44.4 / 46.8 / 34.8 / 36.4 min across 12–18 Sep. The bump recovered on its own before anything was fixed. The AFAR toolchain was never reinstalled — it dates from 2026-04-21.
- **Code growth isn't the driver either.** Two concurrent links, identical flags: 15.1 MB bitcode → 17m04s, 18.1 MB → 21m58s. Link cost goes as size^1.37, so +20% size buys +29% time. Real, mild, not a doubling.

One suggestion for afterwards, not for this PR: with ~97 minutes of headroom restored, `bench_time` in `.github/scripts/submit-slurm-job.sh` is worth setting deliberately rather than left at the value that happened to be one minute above the old cost. The failure mode here was a budget with no slack in it.

